PR #1011, opened 2015-01-06
FEAT: added support for stable sorting.

qtxie:
FIX: wrong result when compare two objects.
TESTS: added test for stable sorting.


dockimbel:
Nice work! :-)

Issue #1031, opened 2015-02-12
Red/System spaces don't seem to be printing from within Red


kealist:
Red[]
    #system [print "Why are all my spaces disappearing"]

will print as:

    Whyareallmyspacesdisappearing

Additionally with routines it will as well:

    Red[]
    f: routine [] [print "Why are all my spaces disappearing"]
    f

with the same output as above.

PR #1032, opened 2015-02-12
FIX: issue #1031 (Red/System spaces don't seem to be printing from withi...

qtxie:
...n Red)


dockimbel:
Thanks for fixing this nasty bug!

PR #1036, opened 2015-02-16
FEAT: modifying win32 print functions to use buffer which improves outpu...

Oldes:
...t to console significantly (as originaly it's printing chars one by one)

### DIFF
--- a/runtime/platform/win32.reds
+++ b/runtime/platform/win32.reds
@@ -38,6 +38,9 @@ platform: context [
 	page-size: 4096
 	confd: -2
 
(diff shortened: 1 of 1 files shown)


dockimbel:
Sounds promising...will give it a try asap. :-)

dockimbel:
Tested, impressive speed improvement, definitely worth merging it! Thank you for it! :+1: 

As we are on the verge of making the new release, and as we might need to make a few changes to that patch you are proposing, I will let @qtxie take care of it and do the merge in the next days.

PR #1037, opened 2015-02-17
FEAT - Add Red #system directive tests

kealist:
This is a preliminary addition of the Red/System tests in Red within the `#system` directive.  I added a file per group of Red/System tests with the exception of the "Compiler Test" group which has Rebol files (which still needs to be converted over).  Currently there are some errors in three groups of the tests which may come from naming conventions.    

I will fix and amend this PR as these are dealt with, but wanted feedback.   

* Is this the ideal way to add these tests (`%run-all-extra.r`)
* Is the `%system/` folder a good place for the tests?
* Should they be broken into files in this way?
* Should the compiler test group be changed to match the form of the other tests or is it in Rebol for a specific reason?


dockimbel:
I thought that these tests files should be generated by a script?

I will let @PeterWAWood have the final word on your PR.

kealist:
I will refactor it that way.  Thank you

PeterWAWood:
@kealist 

We need to have a mechanism that will generate the tests so that when we add more Red/System test files later they will be automatically included in the "embedded test".

The way I would do this is to follow the approach we use to generate the "run-all" Red/tests.

The first step is to extract the list of -test.reds files from Red/system/tests/run-all.r (in the same way that Red/tests/source/units/all-tests.txt was extracted from Red/tests/run-all.r.

The second step is to amend Red/system/tests/run-all.r to run the tests listed in all-tests.r plus the existing Rebol tests.

The third step is to amend Red/system/tests/run-all.r to generate a test files or files that include the Red/System tests and then run those tests.

If you take a look at Red/tests/run-all.r you should get a good idea of what is needed.

kealist:
Step one completed.  Added `%system/tests/source/units/make-reds-test-list.r` to generate the list.  I added the list to the commit until it's auto-generated as part of the process

kealist:
@PeterWAWood So these Red tests of the `#system` directive will live in the `%system/tests` folder?

PR #1039, opened 2015-02-18
Switch multi test

iArnold:
Added a test that for now fails multiple value selection for switch and while editing that file corrected some errors in the commented test switch-all.

### DIFF
--- a/testforswitch
+++ b/testforswitch
@@ -1,1 +1,0 @@
-remove this file
--- a/testforswitch
+++ b/testforswitch
@@ -1,0 +1,1 @@
+remove this file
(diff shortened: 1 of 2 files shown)


PeterWAWood:
Thanks for submitting this extra test Arnold and special thanks for spotting and correcting the error in the commented out test.

The test fails under the compiler. This is not surprising as this feature is not yet implemented (and may even not be implemented).

The automatic builds rely on the tests passing so we cannot add a failing test to the test suite. I will manually add your tests to the test suite with the assert commented out so that the test does not fail. It can act as a placeholder for us until the decision has been made whether to implement this feature or not.

iArnold:
Well the test was supposed to succeed, i.e. the compiled version giving the wrong value, but the test was not executed as compiled code :)

PR #1040, opened 2015-02-22
TESTS: Create a switch for running CALL with /SHOW or /WAIT

kealist:
Made this because I get tired of manually having to manually modify quick-test.r in order to get tests to run on my Windows machines.  Most of uses of `call` were equivalent, and I modified the only unique one to `/output` to an anonymous buffer.


PeterWAWood:
I have tested that these changes do not affect run-all.r under Windows, Linux or OS X.

Issue #1064, opened 2015-03-15
re-pre-compiles console


WiseGenius:
Red has regressed to re-pre-compiling the console every time it opens the interpreter.

	Pre-compiling Red console...

This occurs both in Windows (XP) and Linux (Puppy). I haven't tested any others.

Consequently, if an instance is already running, it can't open another one.

Windows:

	*** Linker Error: locked or unreachable file: /H/Documents and Settings/All Users/Application Data/Red/console.exe

Linux:

	*** Linker Error: locked or unreachable file: /tmp/red/console

pekr:
Confirmed on Windows 8.1

dockimbel:
Cannot reproduce it here on both Windows 7 and Linux. 

Is the problem still present now for both of you?

pekr:
Sadly, not here anymore. Dunno what happened. I run it one hour ago, and it recompiled, but now it seems being stable. Deleted console.exe and tried again, now it recompiles only once. Strange ...

PeterWAWood:
Could the issue possibly be related to a difference in the timezone between the timestamp of the binary and the console generated on the first run of it?

WiseGenius:
I ran them on both Windows and Linux again today, and they each pre-compiled again one more time, but then I can open them as many times as I like and they're fine.
On both, I'm using the same binaries I downloaded yesterday which were pre-compiling every time yesterday.
@PeterWAWood I was thinking something similar.

dockimbel:
It's still the timezone issue and the fact that you get an executable with a timestamp in your future for a few hours, so that the precompilation process is triggered on each run. I have tried to avoid that issue by setting manually the Linux and Mac binaries timestamp 24 hours in the past, but those timestamps were overwritten later by a new building batch process... I have automated that for Linux, so for next release, it should avoid those issues (at least I hope so).

For Windows, I'm very surprised that it occurs there too, I never had any such problem on Windows 7, but don't have a 8.1 to test it. @pekr Can you tell me what is the last modification date of your red-051.exe file?

WiseGenius:
red-051.exe:
Created: Sunday, 15 March 2015, 8:29:25 PM
Modified: Sunday, 15 March 2015, 7:31:37 PM
Accessed: Today, 17 March 2015, 12:09:16 AM

I'm GMT+10:00.

pekr:
I am GMT + 1:

created: 12 ‎March ‎2015, ‏‎19:23:45
[...]

PR #1068, opened 2015-03-20
FIX: Timezone's issue when release a new Red binary.

qtxie:
Making this fix according to @earl's prososal. Thanks @earl .


dockimbel:
Fixes issue #1064.

earl:
:+1: Thanks for implementing, @qtxie.

PR #1073, opened 2015-03-27
FEAT: enhanced Tab completion for console

qtxie:


dockimbel:
What is the improvement?

qtxie:
Tab completion in the middle of a line. For example,
```
red>>round cosi<TAB> 55
```

Issue #905, opened 2014-08-23
Change radians trigonometric functions prefix to "a".


dockimbel:
Replace `arcsin, arcos, arctan` by `asin, acos, atan`.

PR #1078, opened 2015-03-30
change arc* to a*

rheber:
#905 Simple text replace, no?


dockimbel:
I had a quick check on dependencies and there is none. I also checked about the naming (we usually stick to plain words only) and we will make an exception for those radian-oriented functions (default one use full names and are degree-oriented).

So, this PR is fine. :-)

PR #1095, opened 2015-04-17
[FIX] Automatic recompilation of crush library if the source is more rec...

Oldes:
...ent.


dockimbel:
Some parts of your code were manually merged and the rest was written differently to account for the encapped case. So I am closing this PR without merging it. Thanks anyway for your help improving that feature. :+1:

Issue #1098, opened 2015-04-20
OSX: red console crash upon window resize


x8x:
```
% red
--== Red 0.5.2 ==-- 
Type HELP for starting information. 

red>> 
      *** Runtime Error 1: access violation
                                           *** at: 93CE044Bh
```

x8x:
Fixed thx 8-)

PR #1100, opened 2015-04-20
FIX: issue #1098 (OSX: red console crash upon window resize)

qtxie:
```
struct sigaction {
       void     (*sa_handler)(int);
       void     (*sa_sigaction)(int, siginfo_t *, void *);
       sigset_t   sa_mask;
       int        sa_flags;
       void     (*sa_restorer)(void);
};
```
If SA_SIGINFO is specified in sa_flags, then sa_sigaction (instead of sa_handler) specifies the signal-handling function for signum.
As we use `sa_handler` here, we don't need to set the ` SA_SIGINFO ` flag.

### DIFF
--- a/environment/console/POSIX.reds
+++ b/environment/console/POSIX.reds
@@ -441,6 +441,7 @@ init: func [
 		term [termios!]
 		cc	 [byte-ptr!]
 		so	 [sigaction!]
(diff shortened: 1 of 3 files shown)


dockimbel:
Good job!

PR #1105, opened 2015-04-22
FEAT: first+ native implementation

Oldes:
`first+` is native! [introduced in Rebol3](http://www.rebol.com/r3/docs/functions/first-plu.html), which return the FIRST of a series and then increment the series index.

```
red>> s: "abcd"
== "abcd"
red>> while [not tail? s][probe first+ s]
#"a"
#"b"
#"c"
#"d"
red>> b: [1 "foo"]
== [1 "foo"]
red>> first+ b
== 1
red>> first+ b
== "foo"
red>> first+ b
== none
red>> tail? b
== true
red>> head b
== [1 "foo"]
red>> f: %foo
== %foo
red>> first+ f
== #"f"
red>> p: 'foo/bar
== foo/bar
red>> first+ p
== foo
```


Oldes:
With today's update it's possible also:
```
>> b: #{000102FF} while [v: first+ b][print v]
0
1
2
255
```

Oldes:
Btw... I'm aware, that above loop may be done even more efficiently using:
```
forall b [print b]
```
But I believe purpose of the `first+` was for returning value from function with increment series position. Which may be done currently using `also`, but is not so clear:
```
>> b: #{000102FF}
== #{000102FF}
>> also b/1 b: next b
== 0
>> also b/1 b: next b
== 1
```

greggirwin:
This PR is pretty old now, and has conflicts. @Oldes, do you want to update it to see if Nenad will accept it? I don't know if it will make it in as a native, as it's kind of a strange func/name, though handy at times. What do you think of doing it as a mezz, so it's simpler? e.g.
```
first+: func [
	"Return first value in a series, and increments the series index."
	'word [word!] "Must refer to a series (series index modified)."
][
	also  pick get/any word 1  set/any word next get word
]
```
We should start looking at things like this we can put into the support repo for extended function sets, which may be a good place for it.

Oldes:
I'm not interested in mezzanine function for something like this. It was introduced in Rebol to speed up tight loops. Using a mezzanine would just slow them. If you don't like it.. and Nenad probably don't like it.. just close it.

greggirwin:
Thanks for clarifying the goal. That helps a lot. Do you have examples of use cases it was intended for?

Oldes:
No I don't have any real use case... I was not using Rebol3 for real work... and as the PR is from year 2015 and now is year 2018, I believe I did it mainly as an exercise how to make a new native function.

Oldes:
Btw.. to fix the conflict is very easy.

greggirwin:
I think Nenad won't want it, because there's no known use case. I remember thinking it was a good exercise as well. 

I'm going to close this PR, but I'm sure one of us will remember it if `first+` comes up again in the future. I'd also love for you to put it in a gist or something, as an example for others writing natives.

Issue #1115, opened 2015-04-25
red/console stuck in a block


x8x:
```rebol
red>> ["^"
[    ]
[    ]
[    ]
[    
```
neither esc nor `]` break out of multiline

PeterWAWood:
This is due to you not closing the string. ^ is used as an "escape" character in Red. The ^ character needs to be escaped in a string.

```
red>> ["^""
[    ]
== [{"}]
red>> ["^^"
[    ]
== ["^^"]
```

That said strings contained within "" are single line strings and there could be a case for the console automatically closing an unclosed string automatically at the end of a line (if it is possible).

x8x:
Still, esc or `]` should work or I have no choice but to restart the console.
This is properly reported as error:
```rebol
red>>  "^"
*** Syntax Error: invalid Red value at: "^"
```

qtxie:
On Windows, it works fine.

x8x:
osx here

dockimbel:
New fix for this problem, now ESC key will let you go back to prompt.

x8x:
## Console
Thought, `ESC` key will let me go back to prompt, in the future this should have a proper fix.

Here the string! value is properly reported as invalid:
```rebol
red>> "^"
*** Syntax error: invalid value at "^"

*** Where: do
```
Here the interpreter should report same above error and NOT enter multi-line mode, as unterminated string! value can not be set in multi-line mode when using quotation marks.
```rebol
red>> ["^"]
[    ]
[    
(escape)
```
Same example in rebol2 returns proper error:
```rebol
rebol>>  ["^"]
** Syntax Error: Missing " at ["^"]
```
Not sure about this case:
```rebol
red>> print [{^
[    /}]

/
red>> 
```

## Compiler
Compiling below code:
```rebol
Red []
print ["^"]
```
will report:
```
Compiling /red/test.red ...
*** Syntax Error: Missing matching ]
*** line: 2
*** at: {["^^"]
}
```
Compiling:
```rebol
Red []
[{^{]
```
will report:
```
Compiling /Users/alpha/.red/test.red ...
*** Syntax Error: Missing matching ]
*** line: 3
*** at: "[{^^{]^/"
```
Both above compiled examples should return error about bad string! value instead of missing matching square brace.


**NB:** These are corner cases, so very low priority should be assigned, noted for correctness.

PR #1118, opened 2015-04-26
FIX: issue #1115 (red/console stuck in a block)

qtxie:


dockimbel:
Not the right fix. :-)

PR #1121, opened 2015-04-27
FEAT: using hashtable to search symbols

qtxie:


qtxie:
@dockimbel Fixed it. :-)

Issue #1123, opened 2015-04-27
RC to accept just file!


Oldes:
In the example here: http://www.red-lang.org/2015/04/053-faster-compilation-and-extended.html
is:

    >> do %red.r
    >> rc "-c tests\demo.red"
    ...

which is not what is current result. That looks like:

    >> do %red.r
    No source files specified.

Also `rc` could support source file directly, not just string and unnecessary `-c`
I'm creating a pull request with changes.

Oldes:
Now I can do:

   >> do %red.r
   == none
   >> rc %environment/console/console.red

dockimbel:
The "No source files specified." is not a problem at all and even actually informs you that the toolchain has loaded properly and is able to run fine.

PR #1124, opened 2015-04-27
FIX: issue #1123 - rc accepting file! and no 'fail' message if running %...

Oldes:
...red.r without args.


Oldes:
There is one more little issue, if providing `-v` or `-h` as arguments, the compiler calls `quit/return 0`, which is not much friendly behavior.

PR #1139, opened 2015-04-30
FEAT: added support for percent! datatype

qtxie:


qtxie:
Forgot to commit `percent.reds".

PR #1140, opened 2015-04-30
FEAT: added support for percent! datatype

qtxie:


dockimbel:
Think about removing `get*` and `get-any*` functions from percent! and float!.

PR #1155, opened 2015-05-06
FIX: Not encoding tuple correctly in redbin

qtxie:
FEAT: expanded tuple max size to 12 (interpreter only)


dockimbel:
The use-case for colors is to invoke them by name directly for GUI or TUI-oriented code, so they need to reside in global context.

PR #1156, opened 2015-05-06
FEAT: With function

Oldes:
This simple function simplifies binding. Here is example:

```
    red>> name: "Bob"
    == "Bob"
    red>> ctx: context [name: "Jane"]
    == make object! [
        name: "Jane"
    ]
    red>> with ctx [print ["Hello" name]]
    Hello Jane
    red>> name
    == "Bob"
    red>> colors: context [red: 0.0.255]
    == make object! [
        red: 0.0.255
    ]
    red>> with [ctx colors][ print [name "likes red to be" red] ]
    Jane likes red to be 0.0.255
```

### DIFF
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -308,6 +308,22 @@ pad: func [
 
 eval-set-path: func [value1][]
 
+with: func [
+	"Binds context or multiple contexts to block and evaluate it"
+	context [object! block!]  "Context or block with contexts"
+	body    [block!]          "Block to evaluate"
+][
+	either object? context [
+		do bind body context
+	][
+		context: reduce context
+		forall context [
+			body: bind body context/1
+		]
+		do body
+	]
+]
+
 ;------------------------------------------
 ;-				Aliases					  -
 ;------------------------------------------
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -310,17 +310,15 @@ eval-set-path: func [value1][]
 
 with: func [
 	"Binds context or multiple contexts to block and evaluate it"
-	context [object! block!]  "Context or block with contexts"
+	contexts [object! block!]  "Context or block with multiple contexts"
 	body    [block!]          "Block to evaluate"
 ][
-	either object? context [
-		do bind body context
+	do either block? contexts [
+		contexts: reduce contexts
+		forall contexts [ body: bind body contexts/1 ]
+		body
 	][
-		context: reduce context
-		forall context [
-			body: bind body context/1
-		]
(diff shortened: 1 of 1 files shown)


x8x:
Very nice! What is your proposed behaviour if red is defined in both contexts in your last example?

Oldes:
@x8x by default the red has value 255.0.0 (or true, if you are using current version where colors are not defined yet), but there is no real proposed behavior in this example.

Oldes:
But @dockimbel will not include it anyway, I think he likes global variables and is fine with simple binding.. but why not to have it there for others.

Here is a discussion which started me to write the `with` function: http://chat.stackoverflow.com/transcript/message/23122077#23122077

Oldes:
Also in a real life, I would use `with` not for data in the context like in the examples above, but more for something like `modules`. Like here https://github.com/Oldes/rs/blob/master/projects/texture-packer/latest/texture-packer.r#L35

But maybe @dockimbel will come with something better :-)

greggirwin:
The multi context approach is pretty cool. My `with` func only takes a single object, but it's still very handy and make some code much cleaner. The multiple context idea would make a great example, I think, of unification (e.g. prolog).

9214:
Great minds think [alike](https://github.com/9214/daruma/blob/master/src/utils.red#L21) 😉

hiiamboris:
I also independently came to the usage of *with* function, although only needed it for a single context so far. I'd like to support the @9214's version and specifically note how it's different: it does not force a `do` action.
Normally I'm using `with` in two cases:
```
func [a b c] with some-context [
    do a lot of things
]
```
Using plain `bind` in the above is aesthetically awkward because it often makes `some-context` appear on another page, and it's easy to forget about it when reading the function declaration. And note that `do` does not fit this scenario at all.

Another case is this:
`do with some-context [many bad things here]`
instead of
`some-context/many some-context/bad some-context/things here`

Or sometimes more generally:
```
some-damage: with some-context [many bad things here]
either cond1 [... do some-damage... ][... if cond2 [do some-damage ...]]
```

9214:
Personally, I used this construct as a crude substitute for module system, and my use-cases correlate with ones that @hiiamboris have described above.

PeterWAWood:
There seem to be no tests included in this PR.

greggirwin:
Closing this ticket due to age and conflicts. Design ideas at https://github.com/red/red/wiki/WITH-function-design refer to it. We can reopen or create a new ticket as needed.

PR #1285, opened 2015-07-24
FEAT: reduce flickering when drawing image

qtxie:


Oldes:
I'm not sure what flickering you were trying to fix, but with this version I get very bad flickering, when try to resize the window, which was not in previous version.

dockimbel:
Flickering is gone with this patch on my Win7 box.

Oldes:
Flickering is gone with this path ONLY when I revert this line: https://github.com/qtxie/red/commit/39c150d4a90739d8cb844cf1a1e6f2f83971e98d#diff-9a00d7a4561dfb36921acc42a2095d48L1164

PR #1303, opened 2015-08-07
FEAT: multiline area widget

Oldes:
I wonder if you want it this way or you will just style the field. We will want to change fields flag on runtime anyway, won't we?


dockimbel:
I haven't looked into it yet.

Oldes:
There is more widgets based on field like `password`, where just flags are different.

dockimbel:
We'll add FLAGS facet soon, so we'll be able to support more styles.

PR #1325, opened 2015-08-21
FEAT: minimal support for draw TEXT

qtxie:
FEAT: temporary enable anti-alias update.


qtxie:
Need font support later.

PR #1351, opened 2015-09-15
Partial fixes for call

x8x:


x8x:
PR updated with fix for UTF only

PR #1367, opened 2015-09-26
FEAT: added delete-key function in hashtable

qtxie:


dockimbel:
Thank you very much! :-)

PR #1389, opened 2015-10-15
FEAT: added draw function

qtxie:


qtxie:
Forgot to remove `View` module in console. Fix it now. ;-)

PR #1423, opened 2015-11-08
FIX: issue 1284 (Cannot open crush.dll as library)

qtxie:


dockimbel:
Wow, so many issues you fixed in the relocation code! I owe you a dinner, haha! ;-)

PR #1439, opened 2015-11-14
FEAT: added file! to string! support

qtxie:
FIX: fix a resource leak.


dockimbel:
Could you separate the changes in two different PRs?

PR #1466, opened 2015-11-20
FIX: lexer rule for pair! may corrupt stack

qtxie:


qtxie:
This fix will cause another problem, so close it.

PR #1473, opened 2015-11-26
FIX: improving make pair! asserts

Oldes:
Before this fix:
```
red>> make pair! [2.1 3.456]
*** MAKE Error: pair expects a block with two integers
== -858993459x1408749273
red>> 2x2 + 'c
*** Math Error: unsupported right operand for pair operation
== 4408243x2
red>> make pair! 1.2

*** at: 0041A6ECh
```

With this fix:
```
red>> make pair! [2.1 3.456]
*** Syntax error: invalid construction spec: 2.1 3.456
*** Where: make
red>> 2x2 + 'c
*** Script error: word type is not allowed here
*** Where: +
red>> make pair! 1.2
*** Script error: 1.2 type is not allowed here
*** Where: make
```

### DIFF
--- a/runtime/datatypes/pair.reds
+++ b/runtime/datatypes/pair.reds
@@ -108,7 +108,7 @@ pair: context [
 				int: as red-integer! spec
 				push int/value int/value
 			]
-			default [
+			TYPE_BLOCK [
 				int: as red-integer! block/rs-head as red-block! spec
 				int2: int + 1
 				if any [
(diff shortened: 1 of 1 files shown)


dockimbel:
I would have been glad to accept this PR, but that's not the right behavior in such case, as it should just truncate float values as does `as-pair`:
```
red>> as-pair 2.1 3.456
== 2x3
```

Oldes:
Cannot you accept the request and solve possible crash and wait for another with decimal pair support? I'm busy now.. maybe during weekend I could give it a try (with just simple rounding)

dockimbel:
Ok, I want the fix for the error handler, so I'll take the PR. Thanks. ;-)

PR #1474, opened 2015-11-30
FEAT: String from block

Oldes:
With this change one can do:
```
red>> make string! ["a" "b"]
== "ab"
red>> make string! [a "b"]
== "ab"
red>> a: 1 make string! reduce [a "b"]
== "1b"
```

### DIFF
--- a/runtime/datatypes/string.reds
+++ b/runtime/datatypes/string.reds
@@ -909,6 +909,9 @@ string: context [
 				int: as red-integer! spec
 				size: int/value
 			]
+			TYPE_BLOCK [
+				size: 32 								;@@ should we do some analysis and predict required size?
+			]
 			default [--NOT_IMPLEMENTED--]
 		]
 		str: as red-string! stack/push*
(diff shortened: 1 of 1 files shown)


Oldes:
@dockimbel It would be easy to add ```ajoin``` mezzanine or native, based on this feature, but I somehow feel you would not accept it without deep review of all possible string joining functions.

dockimbel:
You are guessing right. ;-)

Oldes:
Why is the type (re)setting needed?

dockimbel:
Because you are allocating a new slot using `stack/push*` but not setting the value type and resetting the `/header` bits.

Oldes:
@dockimbel isn't this enough? https://github.com/Oldes/Red/blob/5ef3c6b7b97ba9cc753f6a96340cedf69842dd85/runtime/datatypes/string.reds#L918

dockimbel:
Ah sorry, yes it's right, github UI was hidding it from me, so I miss it...

image

Oldes:
@dockimbel Ok.. must go to work now.. let me know, if you want me to do the size prediction later or if it's fine for now.

Issue #1559, opened 2016-01-28
changing tab-panel's facets doesn't affect contents


WiseGenius:
Changing a `tab-panel`'s facets while the window is open doesn't affect its contents:
```
view [
	button [f/visible?: false]
	f: tab-panel [
		"English" [text "Hello, World!"]
		"Greek" [text "Χαῖρε, κόσμε!"]
	]
]
```
When clicking the button from the above code, the contents remain visible:
![visible contents](https://cloud.githubusercontent.com/assets/7657453/12643130/f2943284-c605-11e5-8749-1bb1bea5391a.png)

Another example:
```
view [
	button [f/offset: 20x20 f/size: 200x200]
	f: tab-panel [
		"English" [text "Hello, World!"]
		"Greek" [text "Χαῖρε, κόσμε!"]
	]
]
```
![left behind contents](https://cloud.githubusercontent.com/assets/7657453/12643134/01af71ca-c606-11e5-9e2c-613b5bc8da83.png)

PR #1577, opened 2016-02-02
FIX: issue #1559 (changing tab-panel's facets doesn't affect contents)

qtxie:


dockimbel:
We need the same fix for moving group-box as for tab-panel. It would be better to share the same code for that.

Issue #1619, opened 2016-02-12
wrap inverted for text face


WiseGenius:
When clicking the "Wrap" button from the following code, the text is unwrapped, and wrapped when clicking "No Wrap":

```
longer: {The quick brown fox jumps over the lazy dog.}
view [
    f: text "Hello, World!" wrap
    return
    button "Wrap" [f/text: longer f/para/wrap?: yes]
    button "No Wrap" [f/text: longer f/para/wrap?: no]
]
```

Both in Windows XP and Windows 7.
All other faces seem to ignore `wrap?`.

meijeru:
Same for Windows 10. The effect is even clearer if you adjust the size of the window to show the full text (default window size on W10 is too small).

dockimbel:
It works fine now.

Issue #1576, opened 2016-02-02
`align` overrides `wrap?`


WiseGenius:
If an `area`'s `align` is `none` or `'left`, its text is not wrapped, regardless or it's `wrap?` value.
If an `area`'s `align` is `'center` or `'right`, its text is wrapped, regardless or it's `wrap?` value.

So the following code:

```
l: layout [
    text    t: text "wrap?: false" 160x30   tw: text "wrap?: true"
    return
    text "align: none"  a: area aw: area
    return
    text "align: left"  al: area    alw: area
    return
    text "align: center"    ac: area    acw: area
    return
    text "align: right" ar: area    arw: area
]

p: make para! []
pl: make para! [align: 'left]
pc: make para! [align: 'center]
pr: make para! [align: 'right]

pw: make para! [wrap?: true]
plw: make para! [align: 'left wrap?: true]
pcw: make para! [align: 'center wrap?: true]
prw: make para! [align: 'right wrap?: true]

a/para: p
al/para: pl
ac/para: pc
ar/para: pr

aw/para: pw
alw/para: plw
acw/para: pcw
arw/para: prw

view l
```

Produces:
![align overrides wrap](https://cloud.githubusercontent.com/assets/7657453/12753084/3545ca02-ca11-11e5-8a59-c90c12e6c26e.png)

Same behaviour in both Windows XP and Windows 7.

WiseGenius:
An `align:` of `'center` or `'right` only causes the text to be wrapped if that is its initial value. Changing it from an initial `none` or `'left` doesn't give it wrap, and `wrap?` still has no effect.
This can be seen by observing the difference between areas `a1` and `a2` in [tests/align-test.red](https://github.com/WiseGenius/red/blob/801ec032793c5673e9cc6cc0fa5858d0609eba68/tests/align-test.red) when setting them both to `center` or `right` with the radio buttons.

dockimbel:
*If an area's align is none or 'left, its text is not wrapped, regardless or it's wrap? value.*

This is fixable.

*If an area's align is 'center or 'right, its text is wrapped, regardless or it's wrap? value.*

This cannot be fixed using the current `area` implementation, as the underlying Windows class doesn't support the feature. We'll need to use a different class if possible, or implement a custom one to workaround such limitation (though, this is low-priority, as the need for that is probably rare).

WiseGenius:
I agree with the second part having low priority.

The first part is still not quite there, though:
[...]

Issue #1488, opened 2015-12-04
area face needs CRLF (^M^/) to go to newline, ^/ alone is not enough


meijeru:
Apparently there is no "translation" of the Red newline to what Windows needs.

dockimbel:
Another good catch.

dockimbel:
Fix for this class of issues has been pushed to branch `issue-1488`. It needs a few additional changes in the IO code before being mergable in master.

dockimbel:
Fixed by commit [a39c3edb](https://github.com/red/red/commit/dd4009baaab2ed4933dcc23e7813a4f1a39c3edb).

PR #1651, opened 2016-02-21
Expand align-test.red to test #1576, #1619 & #1488

WiseGenius:
@dockimbel Is this close to what you had in mind?


dockimbel:
Yes, thanks! :+1:

PR #1667, opened 2016-03-01
FIX: improvement of control of the Windows version at compile time to...

StephaneVeneri:
... avoid the problem with the function GetGestureInfo


dockimbel:
I accept this PR, though, next time you need to set your code editor in such way that it does not mess up the whitespaces in the file (click on the diffed file above).

Issue #1671, opened 2016-03-02
Drag&drop problems


rebolek:
See this code:
```
Red [
    Needs: 'View
]

system/view/auto-sync?: no

extend styles [
    cbase: [
        template: [
            type: 'base size: 100x50 color: red
            options: [drag-on: 'down]
            actors: object [
            	on-drag-start: func [f e] [f/color: yellow show f]
            	on-drag: func [f e] [print "drag"]
            	on-drop: func [f e] [f/color: red show f]
            ]
        ]
    ]
]

view [cbase "base"]
```

This code shows two problems (on some systems, confirmed by me (W7 VM) and @Pekr (W10), @dockimbel can’t replicate it (W7)):

1) `on-drag` doesn’t work - "drag" is not printed
2) the face disappears on `on-drop`.

dockimbel:
@rebolek Can you confirm that this issue is fixed now?

pekr:
Windows 10, prints "drag" here, when dragged. The face also does not disappear, so hopefully fixed.

dockimbel:
Ok, thanks.

PR #1672, opened 2016-03-02
FIX: issue #1671 (Drag&drop problems)

qtxie:


rebolek:
I’m afraid this does not fix #1671 and needs more testing.

qtxie:
It works on my Win7. I will test it on Win10 later.

dockimbel:
@qtxie Works on Win10 now?

qtxie:
Yes. It works on both Win 7 and Win 10 in my test.

Issue #1534, opened 2016-01-12
read from a file with messy code, lost data


bitbegin:
This is a test file with messy code:

[test.txt](https://github.com/red/red/files/87442/test.txt)

It shows below in notepad++.

```
hello ? world!
```

However, the `read` funciton return a cut off string.

 ```
red>> read %test.txt
== "hello �"
red>> to binary! read %test.txt
== #{68656C6C6F20EFBFBD}
red>>
```

PeterWAWood:
The current ```read``` function expects text to be encoded in UTF-8 format.

If you look at your input as if it was UTF-8 encoded it would be like this:

68 - h
65 - e
6C - l
6C - l
6F - o
20 - (space)
EF - Start of a 3 character UTF-8 sequence
EF BF BD - Is the UTF-8 replacement character

This indicates that your file included an invalid UTF-8 sequence, hence the input was truncated.

bitbegin:
The test file was created on purpose. 

I just think if there are messy codes, `to string!` function can hold these codes and transfer the rest data instead of truncation

PeterWAWood:
The IO in Red 0.6.0 is temporary. Such features are likely to be included when IO support is developed for Red 0.7.0.

qtxie:
Should report an error now with the latest version.

PR #1688, opened 2016-03-09
FIX: Red stopped reading file when it encountered the invalid characters as I described in issue #1534.

bitbegin:
This PR can temporary fix issue #1534.


dockimbel:
Silently ignoring invalid UTF-8 characters while decoding an input is *not* an option.

qtxie:
Now we will report an error in this case. see commit 6f8a15c.

PR #1691, opened 2016-03-10
FEAT: added CHECKSUM native

qtxie:


dockimbel:
Great!

@qtxie We need to move the imports part of the crypto code to %platform/ files.

Issue #1618, opened 2016-02-11
Cursor remains two characters left in Gui console under Wine. Typing becomes difficult


RedJosh:
red-11feb16-9ff1262.exe started from X terminal in Ubuntu 15.10

In the snapshot below, cursor can be seen between b and e of number.

![redgui](https://cloud.githubusercontent.com/assets/17100026/12977175/3ae95700-d0ef-11e5-8d77-042c79fe2ea7.png)

RedJosh:
Cursor became al right after installing Consolas font in Wine.

x8x:
Instructions to install Wine with Consolas font on OSX:
```
brew install wine --without-x11 --devel
brew install winetricks
# set Wine to default to Windows 7
winetricks win7
# install Consolas font
winetricks consolas
```
almost ready to go, download latest Windows Red build and lunch it with wine:
```
curl -o red.exe http://static.red-lang.org/dl/auto/win/red-19feb16-f834e41.exe
wine red.exe
```
[should be similar with apt-get (Debian) and pacman (Arch Linux)]

RedJosh:
I recommend including these instructions in the Red Installation instructions and thereafter close this issue.

PR #1737, opened 2016-03-26
Add Some Notes to the Readme

nc-x:
Add notes regarding
-  `--cli` on Windows. (@dockimbel the default mode on other OSes is `cli`, right? )
-  Issue related to Font on Wine https://github.com/red/red/issues/1618

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,12 @@ Running the Red REPL
         red>> inc 123
         == 124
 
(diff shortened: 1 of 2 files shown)


dockimbel:
@nc-x Yes, on non-Windows OSes, it is `--cli` by default, but that might change in the future.

PR #1738, opened 2016-03-26
console: better support for Ctrl-K and Ctrl-D.

mrjbq7:
Ctrl-K will erase to end of line.
Ctrl-D will remove character or exit like Ctrl-C if empty line.

### DIFF
--- a/environment/console/input.red
+++ b/environment/console/input.red
@@ -388,8 +388,25 @@ unless system/console [
 							refresh
 						]
 					]
-					KEY_CTRL_C
+					KEY_CTRL_K [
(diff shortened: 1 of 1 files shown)


mrjbq7:
Should I re-commit with ``FEAT:`` prefix in commit message?  I didn't notice that is your convention.

dockimbel:
Thank you! Yes, it would be better to use `FEAT:` prefix if you can re-submit.

mrjbq7:
Okay, I re-did the commit message and re-pushed.

dockimbel:
Thanks again! :+1:

PR #1756, opened 2016-03-30
Some variable scope policy fixes in the curses binding

muspellsson:
As I can see variables defined within functions are no more global ones in Red > 0.6.0. This behaviour has slightly broken the curses library binding (and helped to find two typos).

Also I'm not sure about the initial value of stdscr, so for now it's just zero.


dockimbel:
Fixes very much appreciated, thanks!

PR #1805, opened 2016-04-10
FEAT: Transcode/next

SteeveGit:
USAGE as R3:
red> system/lexer/transcode/next "1 2 3" [ ]
== [1 " 2 3"]

You should pass your own returning block since the default allocated one is of length 200.
Like in:
red> system/lexer/transcode/next "1 2 3" none

### DIFF
--- a/environment/lexer.red
+++ b/environment/lexer.red
@@ -315,6 +315,7 @@ system/lexer: context [
 		dst	[block! none!]
 		/part	
 			length [integer! string!]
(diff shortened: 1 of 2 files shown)


dockimbel:
`load/next` has been implemented. It follows the R3 model for `/next` rather than R2's model (the one you implemented in your PR). The Red/R3 version does not allocate a new block on each call.

Sorry for not merging your PR, it was well written.

PR #1806, opened 2016-04-10
FIX: sub-block in draw makes next item skipped

SteeveGit:


modules/view/draw.red
```diff
@@ -361,7 +361,6 @@ Red/System [
 					]
 					TYPE_BLOCK [
 						parse-draw as red-block! cmd DC catch?
-						cmd: cmd + 1
 					]
```
SteeveGit: cmd: cmd + 1, is done later in code

qtxie:
Good Catch! Thanks. :+1:

PR #1811, opened 2016-04-13
FIX: Boolean expression simplification in set-flag

SteeveGit:


dockimbel:
Perfect, thanks!

qtxie:
I think this change is not correct. `1 and 0 or 0` is not equal to `1 or 0`.

PR #1849, opened 2016-04-25
FEAT: New-line marker

SteeveGit:


runtime/datatypes/block.reds
```diff
@@ -1278,7 +1299,6 @@ block: context [
 				][
 					copy-cell value head
 					if hash? [_hashtable/put table head]
-					head: head + 1
```
dockimbel: Can you tell me what is this change for? If it's a fix, which use-case does it fix?

SteeveGit: I don't recall it, I think it's unwanted.

PR #1855, opened 2016-04-26
Added /WITH refinement to PAD function

endo64:
Sample usage:

    red>> pad/with "test" 10 #"." ;== "test......"


dockimbel:
Thanks! Could you please `squash` all the changes into a single commit please? Also put a `FEAT:` prefix in the new commit message.

Issue #1869, opened 2016-05-01
Base doesn't get focus when field have the focus


qtxie:
```
view [
    base "base"
         on-key  [probe event/key]
         on-down [event/window/selected: face]
    field "ok"
]
```

Click the field first, then click the base, the base does not get the focus.

hiiamboris:
There's a regression here. The provide code will no longer give focus to the base (except on double-click).
Moreover, `field` also steals focus from other faces as well:
try `view [field area "X" on-down [set-focus face] on-key-down [probe event/key]]`
or `view [field button "X" on-down [set-focus face] on-key-down [probe event/key]]`.

greggirwin:
@qtxie is this a regression, or newly intended behavior?

qtxie:
@greggirwin Looks like a regression.

Oldes:
@qtxie if I run your last code snippet, I think that it should be possible to click anywhere outside the `field` or `area` to lost focus to them.

I think it would be fine to just add `SetFocus hWnd` at this line: https://github.com/red/red/blob/53b5b37b3fd37a2a4b88452b2cc90a7e89c5f428/modules/view/backends/windows/events.reds#L1333

Btw... you are catching just the left mouse button?

greggirwin:
@Oldes, on Windows, that's not the standard behavior. Not saying we can't do it, but maybe we should check other platforms to see what they do.

hiiamboris:
I think it magically fixed itself one day ☻
Can't reproduce on W7.
Anyone still having the issue?

greggirwin:
Seems fine here on Win10.

PR #1870, opened 2016-05-01
FIX: issue #1869 (Base doesn't get focus when field have the focus)

qtxie:


SteeveGit:
Thanks, I was waiting for it.

PR #1915, opened 2016-05-18
FEAT: convert key to word! when foreach a map

qtxie:


dockimbel:
Is this PR trying to fix a github ticket?

qtxie:
No. As `keys-of` returns all the keys in a map as `word!`, for the same reason, I made this change.

dockimbel:
But the question then is: why should we allow FOREACH on map? Map has a name/value structure, which is not a good match for FOREACH which is for series...

qtxie:
One use case is that I want to iterate all the key/value pairs in a map! when trying to convert a map! to a JSON object. My first thought is to use `foreach`.

rebolek:
@dockimbel *why should we allow FOREACH on map?* Because it’s convenient:

R3:

```
>> m: map [a: 1 b 2]
== make map! [
    a: 1
    b 2
]

>> foreach k m [print m/:k]
1
2
```

Red:

```
red>> m: #(a: 1 b: 2)
== #(
    a: 1
    b: 2
)
red>> foreach k m [m/:k]
*** Script error: foreach does not allow map! for its 'word argument
*** Where: foreach
```

In R3, `foreach value map []` is a shortcut of `foreach value keys-of map []` and it’s useful.

PR #1927, opened 2016-05-22
Added SHA 256/384/512 support (Windows only).

greggirwin:
There is some refactoring that could be done in %natives.reds and condensing the %crypto.reds CRC/MD5/SHA* funcs to take an alg type arg.

### DIFF
--- a/environment/natives.red
+++ b/environment/natives.red
@@ -752,13 +752,13 @@ wait: make native! [[
 ]
 
 checksum: make native! [[
-		"Computes a checksum, CRC, or hash"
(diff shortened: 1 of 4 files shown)


dockimbel:
@greggirwin Could you please make a separate PR for the routine descriptions?

greggirwin:
Would you rather I just delete this commit? I thought it would go in as a separate PR, but now searching shows git doesn't work that way. What is the best way to do this, going forward, without having to create a branch?

greggirwin:
Commit reverted.

Issue #1907, opened 2016-05-16
Redefining `set` crashes the console


nc-x:
```
red>> set: 1

*** Runtime Error 1: access violation
*** at: 00000000h
```

meijeru:
Isn't there a general danger that redefining built-in words will lead to unpredictable results?

nc-x:
@meijeru Yes. Modules support is gonna help a bit. But besides modules, there should be a way to create immutable variables, else I wouldn't really like if someone breaks a library I create by accidentally redefining an internal private api...

dockimbel:
@meijeru Difficult to say. The console does rely on some interpreted code, so changing core actions/natives will have unpredictable effects. I'm not even sure in that context if it's possible or not to circumvent all potentially crashing outcomes, this needs a deeper study.

dockimbel:
Modules with isolated namespaces should provide a more robust solution to this class of issues.

PR #1928, opened 2016-05-23
FIX: regression on fix for issue #1907

qtxie:


dockimbel:
Thank you for fixing my mistake! ;-)

PR #1971, opened 2016-06-04
FIX: #1970 (Too much 'over events on draggable face)

### LINKED ISSUE #1970: 'over events continue to be fired after a face has been dragged  (loose)
In the console, run the following test:
```
system/view/debug?: yes
view/options [b: base red loose][size: 400x400]
system/view/debug?: no
```
Move the mouse over the red box. 
No 'over events are fired at that point (except when entering and leaving the box). 
Drag the red box a little, then release it.
You'll notice in the console that a lot of 'over events continue to be fired when the mouse is inside the red box.
It should not be.

The reason is, the facet face/flags is not cleared:
`red>> b/flags`
`== all-over

...FIX incoming...

SteeveGit:
Fix #1970

### DIFF
--- a/modules/view/view.red
+++ b/modules/view/view.red
@@ -764,8 +764,10 @@ insert-event-func [
 		face: event/face
 		type: event/type
 		either type = drag-evt [
-			unless find-flag? flags: face/flags 'all-over [
-				either block? flags [append flags 'all-over][face/flags: 'all-over]
+			face/flags: any [
+				all [not block? flags: face/flags :flags reduce [:flags 'all-over]] 
+				all [flags append flags 'all-over]
+				'all-over
 			]
 			do-actor face event 'drag-start
 			face/state/4: event/offset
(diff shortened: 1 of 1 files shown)


modules/view/view.red
```diff
@@ -764,8 +764,10 @@ insert-event-func [
 		face: event/face
 		type: event/type
 		either type = drag-evt [
-			unless find-flag? flags: face/flags 'all-over [
```
SteeveGit: Note that the flag 'all-over is always added, event if it already exists  (not anymore checked).
This way, the flag suppression is safe. 
The developer may have already set the 'all-over previously. And it's kept as-is even after the second  'all-over flag has been suppressed.

SteeveGit: Based on this code sample, I think append-flag and remove-flag natives function would be good additions. I'm pretty sure it's used elsewhere (VID)

dockimbel: I'm ok with your fix, though, as you say, it would be better to implement `set-flag` and `clear-flag` for such usages (as optimized natives, on the model of `find-flag?`). I have implemented a [set-flag](https://github.com/red/red/blob/master/modules/view/view.red#L30) function at Red level, but that's temporary, it really needs to be coded in R/S as a native (as those natives are generic enough, they could be used in other contexts).

You can also implement the `set-flag` and `clear-flag` (as new native! functions) in a new PR if you wish, I'll accept it gladly. (`find-flags?` should also be moved to a native).

PR #1976, opened 2016-06-07
New checksum spec and test suite

greggirwin:
Sorry it's 3 commits instead of 1. Even branched and tried to squash it.

### DIFF
--- a/environment/natives.red
+++ b/environment/natives.red
@@ -752,15 +752,11 @@ wait: make native! [[
 ]
 
 checksum: make native! [[
-		"Computes a checksum, CRC, or hash. Default is CRC32."
+		"Computes a checksum, CRC, hash, or HMAC."
 		data 	[binary! string! file!]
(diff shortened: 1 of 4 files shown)


dockimbel:
Number of commit does not matter as long as they are related to the same feature. Merge commits resulting from just upgrading the branch from `master` should be avoided though (it's possible to do merges without generating commits). I squash sometimes the proposed PR when I see too much "noisy" commits, or badly written commit logs.

greggirwin:
Thanks. I think I screwed myself by pulling the TCP/HMAC fixes out to submit first. I ended up in a death spiral of always getting a conflict on that, when the file was identical. 

Will try to avoid commits on merges and use prefixes. In this case, would you have preferred the test suite being a separate commit with its own prefix?

dockimbel:
For tests, yes, it's better to put them in a separate commit with a TESTS: prefix.

PR #1978, opened 2016-06-07
Correction on LOAD/PART

iArnold:
The code that effectively should work for load/part was never performed because in the first part it was always caught and returned make block! 1 too early.
Also zero? performed on length when length is a string crashed load/part.

Added tests too.

### DIFF
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -300,7 +300,7 @@ suffix?: function [
 
 load: function [
 	"Returns a value or block of values by reading and evaluating a source"
-	source [file! url! string! binary!]
+	insource [file! url! string! binary!]
 	/header "TBD: Include Red header as a loaded value"
 	/all    "TBD: Don't evaluate Red header"
 	/part
@@ -310,6 +310,7 @@ load: function [
 	/as   "Specify the type of data; use NONE to load as code"
 		type [word! none!] "E.g. json, html, jpeg, png, etc"
 ][
+    source: copy insource
 	if as [
 		if word? type [
 			either codec: select system/codecs type [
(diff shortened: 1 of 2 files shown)


dockimbel:
The first step when you identify a problem is to open an issue, describing the problem. This would have   spared you the time on fixing the wrong problem and implementing the wrong feature... That is not how `/part` refinement is supposed to work with a series argument (re-reading the Rebol documentation as I told you several times would have taught you that).

The right fix would have been simply to move the `zero? length [return make block! 1]` case on second position after the `string? length [...` case.

You can open a ticket for the issue, with a proper error case and then submit a new PR with the right fix for the issue (including some tests like you did here).

iArnold:
Afaict, there is NO refinement of the name 'PART in REBOL. Not in R2 nor in R3.
Red help documentation does have a /part refinement for load. The mentioned length can be both an integer! or a string! and for string! it did not do much.

dockimbel:
*Afaict, there is NO refinement of the name 'PART in REBOL. Not in R2 nor in R3.*

Both in Rebol2 and 3: `copy/part`, `find/part`, `insert/part`, `change/part`, `remove/part` and so on... Next time try to write "NO" even bigger and in bold, maybe it will turn something utterly false into true, who knows, magic seems to be the only thing you can rely on, reading a documentation, even after I tell you __three__ times now to do so, is way beyond your abilities.

iArnold:
You are right. I understand the basics of how bits and bytes work. I can read almost any PL out there. I can use some of them. Some I can really use well. Some professionally. Some hobby wise. I have been programming for a long time. 
Only using Rebol I was able to program a real chess/checkers game. I have streched the abitlities of the language (I talk about Red now) by adapting code to what was possible. Sometimes, more often than not, the code was broken for regressions or change of insight. I reported these. 
I am trying to progress this project by any means of being an INDEPENDENT outsider. A person who is wharmhearted to this particular project, trying to fix some minor things now and then, not expecting to be put down like you have done here.

PR #1979, opened 2016-06-07
FEAT: new checksum spec for Mac and Linux

qtxie:
FIX: using allocate memory in get-digest as static array is not thread safe

### DIFF
--- a/runtime/crypto.reds
+++ b/runtime/crypto.reds
@@ -236,7 +236,7 @@ crypto: context [
 		free ipad
 		free opad
 		free idata
-		free odata			
+		free odata
 		free ihash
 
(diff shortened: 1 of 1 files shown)


greggirwin:
%crypto.reds/CALC-HMAC returns the result of GET-DIGEST. There's a "who frees this?" comment there. It seems that %natives.reds/CHECKSUM* should do that (its local 'b word). I don't know how native dispatching works. If `stack/set-last` has to be the last thing evaluated, where can we free `b`?

dockimbel:
`stack/set-last` just sets the return value on the Red stack, you can safely put R/S code after it.

PR #1982, opened 2016-06-08
Celebrating Tau Day 6 28 2016

iArnold:


dockimbel:
Pi is right: http://www.thepimanifesto.com

iArnold:
You are not just doing this because I did it, right?

hostilefork:
From "The Pi Manifesto":
> Last updated July 4th, 2011

From "The Tau Manifesto"
> updated Tau Day, 2015

http://tauday.com/tau-manifesto#sec-the_pi_manifesto_a_rebuttal

Guess some people would prefer telling those who have better ideas they are "cranks", and cling to the half-thought through contents of long dead websites, *cough* vs. [actually using one's brain](https://trello.com/c/rmsTJueg).

#goodluckwiththat
#notanendorsementofiarnold

rebolek:
Hm, Brian, you must be joking, right? If there was an article "Earth is flat" from 2015 and "Earth is round" from 600 B.C., would you choose the newer one just because it’s newer?

hostilefork:
@rebolek I'm joking in the sense I don't care if this PR is accepted or not, as it is pointless, and iArnold actually probably *is* a troll, though I guess I don't actually know what his deal is.

I'm not joking in the sense that if Red does not--at some point--review the large body of thought and work I've done, and adopt the designs, going rather with lots and lots of mistakes on top of mistakes...that I'm just going to have to assume Doc is an idiot.

PR #1984, opened 2016-06-09
FEAT: Add functions ENTAB and DETAB

iArnold:
Complete with testfile and success in testing:

```
>> do/args %quick-test/run-test.r "tests/source/units/entab-detab-test.red"
Quick-Test v0.12.0
Running under REBOL 2.7.8.3.1
~~~started test~~~ entab-detab
~~~finished test~~~  entab-detab
  Number of Tests Performed:      7
  Number of Assertions Performed: 7
  Number of Assertions Passed:    7
```

Maybe you want to have a look at the use of new macro TAB_CHAR? I used. It is just like SPACE_CHAR? but perhaps use is not wanted or not consistent enough.


SteeveGit:
I'm wondering... (Not to distrust your good will, a genuine question)
Since it probably will never be used by other natives and only on relatively small sets of small scripts (so, maximizing speed is not and issue). Is it necessary to code it with Red/system? Isn't parse not  enough and well fit for such functions ?

iArnold:
Hi Steeve,

Thank you for your concern.
Both functions are native in R2 (and R3). The task is put under header 'NATIVE'. It is relatively easy to transform c code to Red/System code. 
Your suggestion on 'PARSE I cannot share. To me PARSE is a two-headed beast. It looks simple, and in a sense it is, as it is much friendlier than Regex. Some things are simple in PARSE, but other tasks are not. As an example of this is the implementation of SPLIT. SPLIT has a bug that makes it skip empty entries in the input. So split "a;;b" ";" will result in ["a" "b"] instead of ["a" "" "b"]. Steeve, please be my guest to fix this SPLIT bug using a simple fix on the current PARSE expression and prove 

> "maintenance would be easier and doable by more developers."

I foresee 1. little maintenance on these functions 2. Lots of programmers grasp the program flow of these Red/System functions quicker than they do complex PARSE statements. But that is my insignificant opinion.

Regards,

Arnold

xqlab:
`split: func [series dlm] [ collect [ parse series  [any [[copy s to dlm skip (keep s)] | [copy s to end  (keep s) skip ] ]  ]  ]]`

iArnold:
Hi xqlab,

Thank you for your contribution here. Perhaps it is usable as a separate PR for this problem (I think it needs its own issue to be reported yet).
Two things:
a) I do not consider that parse expression as a trivial, something a lot of ppl could come up with. And
b) I asked Steeve to come up with an answer. Not an expert like you ;-P

Regards,

Arnold

SteeveGit:
Ha, sorry @Arnold, I missed your request.

greggirwin:
This is a tough one. I love the elegance, and tiny size of a `parse` solution (I have a gist up), but it isn't super fast. It takes maybe a second here to process %compiler.r, which is ~100K. I'm guessing it could be a lot faster by counting, allocating a full buffer once, and collecting into that, but Arnold's/System code will probably still run rings around it. Each implementation also has value as example code.

If the /System version is as fast as Rebol's C code, or close, it's probably worth the added size and complexity. If there's another good func, or someone writes examples, for high performance string processing, it's an even tougher call.

Arnold, how hard would it be to implement the `/leading` refinement like my parse version? I think there's a lot of value in that.

iArnold:
Hi Gregg,

entab is just like Rebol also leading spaces only.

I think it is doable. 
After this one is pulled ;-)

Arnold

We can chat about this endlessly on AltME.

dockimbel:
Unilateral implementation of entab/detab without prior discussion with Red team for the specs.
Sub-optimal port of Rebol3 C code to R/S with poor test coverage.
Sneaky changes of unrelated unit tests variables.
Sneaky changes of unrelated R/S documentation.
Sneaky addition of a global variable in Red environment.

PR #1987, opened 2016-06-10
Update functions.red

xqlab:
More in the line how Rebols simple parse splits strings.
red>> split "1,2..,4" charset ",."
== ["1" "2" "" "" "4"]
red>> split "1,2,,,4"  ","
== ["1" "2" "" "" "4"]
red>> split "1,2,.4"  ",."
== ["1,2" "4"]

### DIFF
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -685,16 +685,19 @@ flip-exe-flag: function [
 	write/binary file buffer
 ]
 
-split: function [
-	"Break a string series into pieces using the provided delimiters"
-	series	[any-string!]	"String series to split"
-	dlm		[string! char!]	"Delimiter as a char or string pattern"
-	return: [block!]		"Returns a block of split pieces without the delimiters"
-	/local value
-][
-	rule: complement charset dlm
-	parse series [collect [any [keep copy value some rule | skip]]]
+split: func [
+    {Break a string series into pieces using the provided delimiters} 
+    series [any-string!] dlm [string! char!  bitset!]  /local s  num
+] [
+    either string? dlm [num: length? dlm] [num: 1]
+    collect [
+        parse series [
+            any [[copy s to dlm num skip (keep s)] | [copy s to end (keep s) skip] ]
+        ]
+    ]
 ]
+
+
 
 dirize: func [
 	"Returns a copy of the path turned into a directory"


meijeru:
I don't understand the example `red>> split "1,2,.4" ",."` shouldn't the last argument be `"."` for this to work?

xqlab:
Am 10.06.2016 um 14:52 schrieb Rudolf Meijer:
>
> I don't understand the example |red>> split "1,2,.4" ",."| shouldn't 
> the last argument be |"."| for this to work?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub 
> <https://github.com/red/red/pull/1987#issuecomment-225173722>, or mute 
> the thread 
> <https://github.com/notifications/unsubscribe/ANK3aXIKeAbGM0vfkzZXzQlZoTA-BN8oks5qKV39gaJpZM4Iy6aB>.
>
That's a new and additional split feature. It splits at strings as one 
continuous delimiter.

iArnold:
Hi xqlab,

There is a problem in your implementation of split
split "a;b::c;d" ";:" 
does not split the string using the different possible delimiters in this case both the colon and the semi-colon.

Regards
Arnold

xqlab:
Hi Arnold,

have a look at the examples.

split "a;b::c;d"  charset ";:"

should do what you expect / desire.

regards


Am 10.06.2016 um 16:39 schrieb iArnold:

iArnold:
Hi xqlab,

No it does not. Your uses the complete string to delimit, not each character as a delimiter character. The result should be like the result on try.rebol.nl for R3:


== ["a" "b" "" "c" "d"]
Your example
red>> split "1,2,.4" ",."
== ["1,2" "4"]
clearly does something different.

The matter is really pretty complex, as I mentioned earlier.

Nevertheless I think I could live with this split until a better one is accepted/provided by the Red team if the working is correct for a single delimiter character. Which is the usecase I needed the change originally for.

Regards,

Arnold

xqlab:
Hi Arnold,

yes, it does something different in this case. But this is by design 
offering to split with a whole contiguous string additionally.

If you want to get the exact same behaviour, you should use this variation


split: func [
     {Break a string series into pieces using the provided delimiters}
     series [any-string!] dlm [string! char! ]  /local s  num
] [
     if string? dlm [dlm: charset dlm ]
     collect [
         parse series [
             any [[copy s to dlm  skip (keep s)] | [copy s to end (keep 
s) skip] ]
         ]
     ]
  ]

red>> split "1,.2,3"  ",."
== ["1" "" "2" "3"]
red>> split "a;b::c;d"  ";:"
== ["a" "b" "" "c" "d"]
red>> split "1,2,.4" ",."
== ["1" "2" "" "4"]


regards


Am 10.06.2016 um 19:12 schrieb iArnold:

iArnold:
xqlab,

Delimiting by a longer string is also functionality I can see put to use in some reallife use cases. In that case if the Red team also agrees, a combination of both will be perfect and a suiting refinement-name chosen. (perhaps /all?)

regards,

Arnold

dockimbel:
I take this PR, though, `split` is still a temporary implementation for now, it should be replaced by a much more capable one.

dockimbel:
@xqlab One thing though, Red sources are using tabs for indentation not spaces, please convert your code to proper format next time you submit a PR.

iArnold:
@xqlab congrats 
@dockimbel I am looking into 'deline native implementation. Deline has an refinement /lines that splits the string into a block of lines. Is this functionality wanted or should the programmer use split and not deline/lines? I think deline with /lines must make use of its own split? (Which would suggest creating split as a native split function.)

PR #1998, opened 2016-06-12
FEAT: added SET-ENV, GET-ENV and LIST-ENV natives

qtxie:


runtime/platform/POSIX.reds
```diff
@@ -253,4 +274,38 @@ set-current-dir: func [
...
 	zero? chdir path
+]
+
+set-env: func [
+	name	[c-string!]
+	value	[c-string!]
+	return: [logic!]			;-- true for success
+][
+	either value <> null [
+		either -1 = setenv name value 1 [false][true]
```
dockimbel: Could be just: `-1 <> setenv name value 1`

runtime/platform/POSIX.reds
```diff
@@ -253,4 +274,38 @@ set-current-dir: func [
...
+	][
+		either -1 = unsetenv name [false][true]
```
dockimbel: Could be just: `-1 <> unsetenv name value 1`

SteeveGit:
I'm trying to check out (from a red script) if my gui-console.exe is running under wine on an linux system, because I have problems with bad formatted file paths (something with missing slashs, I'll probably report an issue later).
I found: 
> Red>> get-env "SELINUX_INIT"
> == "YES"
It would be interesting to know what OSX users get and compare.
Unfortunatly, I didn't found any var with "wine" inside but a lot with "ubuntu" which is too much system specific.

**### BTW: list-env is crashing here with wine. Would be interesting to know if I'm the only one before posting an issue**

dockimbel:
@SteeveGit Can you show us the crash logs from Wine? It should show which API call failed.

SteeveGit:
Youhouuu!
```
--== Red 0.6.0 ==-- 
Type HELP for starting information. 


*** Runtime Error 9: assertion failed
*** in file: /Z/home/a/Desktop/Git/red/runtime/allocator.reds
*** at line: 
***
***   stack: red/alloc-series-buffer   
***   stack: red/alloc-series-buffer   
***   stack: red/alloc-series   
***   stack: red/unicode/load-utf16 h  h false
***   stack: red/string/load-in h  h 
***   stack: red/list-env
***   stack: red/natives/list-env* false
***   stack: red/interpreter/eval-arguments h h h h h
***   stack: red/interpreter/eval-code h h h false h h h
***   stack: red/interpreter/eval-expression h h false false
***   stack: red/interpreter/eval h true
***   stack: red/natives/catch* true 
***   stack: ctx275~try-do h
***   stack: ctx275~do-command h
***   stack: ctx275~eval-command h
***   stack: ctx275~run h
***   stack: ctx275~launch h
***   stack: ctx294~launch h
854798427422620725249440100100022167400000000000221674000683D70-10071206C0079A9980079A99800000000000000000071206C0079A9980079A99800000000000000000071206C0079A9980079A99800683D5010033C2F40033C2F40033C2F40033C2F40033C2F40033BDF8
```

SteeveGit:
Maybe it's related, but I got from:
> get-env "LANG"
> == "en_US.UTF-8"
So any problems, when trying to get strings as UTF-16 ?, maybe.

Issue #1954, opened 2016-06-01
Error in view-test.red


TimeSlip:
Hello Doc,
This is a simple one. 
image: load %./bridges/android/samples/eval/res/drawable-xxhdpi/ic_launcher.png
(line 533)
requires an extra period?
image: load %../bridges/android/samples/eval/res/drawable-xxhdpi/ic_launcher.png

BTW, this implementation of VID is incredible. Love it already. Regards, James

dockimbel:
James, thanks! Could you submit a Pull Request with the change. You can use only Github for that if your are not used to Git:

1. Fork red/red repo using the "Fork" button.
2. Navigate to the right file, open it and then click on the pen icon ("Edit this file").
3. Make the change, save it, then click the "New Pull Request", then "Create Pull Request".
4. Wait for us to review and accept it. That's all.

qtxie:
Actually it depends on where you put your `view-test.exe` or what the current dir is in console.
In gui-console:
```
cd %/d/red-repo
do %tests/view-test.red
```
It works fine without error.

nc-x:
@qtxie notice the `.` after `%`

TimeSlip:
The thing is that yes it works depending on where you are. @qtxie, in your case, you're in the %/d/red-repo directory and from that level and position it's fine. I'm just worried that it's not obvious that one has to be in a particular level, calling the app which is located in a different folder if you know what I mean. @dockimbel Rather than me messing it up some more, could I make the suggestion that you either leave it as is or add an "either" call to a load http://where-the-image-can-be-found and/or perhaps even an dummy image (generated on the fly maybe), so although it is dependent on having connectivity, it does not stop the application completely.
Anyway, you're the best my friend. I still remember the little show-n-tell we had with Henrik and Gabriele in the early days.

dockimbel:
The %view-test.red is really a "test" script used for development, so, like all other test scripts, it should be run from the Red root folder.

PR #2040, opened 2016-06-17
Fixed path to android png image.

greggirwin:


qtxie:
Is it a fix for this issue #1954 ?

greggirwin:
Shoot. Yes. Didn't see that. Dismiss my change then, but note that I got bit like James did.

PR #2064, opened 2016-06-23
Added native SIGN? func.

greggirwin:


runtime/natives.reds
```diff
@@ -1474,6 +1474,40 @@ natives: context [
...
+	sign?*: func [
+		check?  [logic!]
+		return: [red-integer!]
+		/local
+			i   [red-integer!]
+			f	[red-float!]
+			res [red-value!]
+			ret [integer!]
+	][
+		#typecheck -sign?-								;-- `sign?` would be replaced by lexer
```
dockimbel: Decorating `sign?` there is not necessary, as it's only used when the word could conflict with an existing R/S definition, in such case, a decorated/undecorated pair needs to be added to this function: https://github.com/red/red/blob/master/compiler.r#L3652

You can remove the decoration in another PR, I will accept this one as-is.

greggirwin: Thanks! Learning all the time.

iArnold: @greggirwin Another Guru meditation coming up? And how does one quickly check if a conflict with an existing R/S definition occurs?

greggirwin: I added the primary note, but don't have a solution for quickly checking names yet.

PR #2066, opened 2016-06-23
Removed period from end of checksum doc string.

greggirwin:
Edited directly in github


dockimbel:
For small and safe changes, Github web interface is indeed everything you need. Very handy sometimes. ;-)

PR #2131, opened 2016-07-24
FEAT: use DirectWrite to draw Text in BASE face when possible

qtxie:


pekr:
errors out, as rebol.com is down ...

PR #2140, opened 2016-07-30
Added SCALAR? and IMMEDIATE?

greggirwin:


greggirwin:
So you don't want tests that would catch typeset changes? i.e. those are not considered regressions?

dockimbel:
If we change the typesets definitions, it will be for good reasons, having to maintain such tests is double work.

PeterWAWood:
I think it would be a good idea to have just two tests, one positive, one negative for each of the functions just to confirm the function is working. The other tests are ideally redundant as they should  be duplicate of typeset tests.

greggirwin:
I understand, don't agree entirely, but OK. Not worth your time right now. Go ahead and close this. Or do I need to since it's not being merged?

dockimbel:
@greggirwin You can just add another commit to change the tests part. The functions definitions are fine.

greggirwin:
Done.

dockimbel:
@greggirwin Thanks. I can't merge it as there's a conflict now (not sure from where it comes from).

dockimbel:
Conflict is gone, merging now.

PR #2168, opened 2016-08-11
Remove period from SIGN? doc string

greggirwin:
I thought this had been done already.


greggirwin:
Garrrh. Should have put FIX on the front of the commit message.

PR #2197, opened 2016-09-02
FEAT: Now returns process exit code and delines 'out and 'err.

greggirwin:
This makes it R2 compatible, as a drop-in replacement for CALL.


greggirwin:
Just realized that `deline` isn't in 2.7.6. It's a mezzanine in 2.7.8, so could be patched if wanted.

PR #2222, opened 2016-09-10
FIX: change write paths to relative for libred

geekyi:
Changed absolute paths to relative, so it works regardless of where the red source folder is.
If this isn't what you were intending to do, feel free to close it.

I did this partly to understand how github pull requests work (so a simple change)


dockimbel:
The access paths to those files are temporary as it is a development branch. Though, I see no harm in accepting your PR, as that code is temporary anyway. ;-)

PR #2226, opened 2016-09-12
FIX: adjust rotate, scale, translate, skew, transform to affect also commands after sub-block

Zamlox:
Within a 'draw' block,commands 'rotate', 'scale', 'translate', 'skew' and 'transform' will not affect commands after a sub-block:
```
view [base 100x100 draw [ translate 10x10 [ pen blue box 0x0 30x30] pen red box 0x0 50x50 ] ]
```
the red rectangle will be drawn at 0x0 instead of 10x10
This happens because state is saved on stack using 'push' before executing one of those commands. When 'pop' is executed, it will discard the effect of that command.
The fix will save state on stack after transformation command is executed and before parsing sub-block.

### DIFF
--- a/modules/view/draw.red
+++ b/modules/view/draw.red
@@ -426,8 +426,8 @@ Red/System [
 								DRAW_FETCH_OPT_VALUE(TYPE_PAIR)
 								DRAW_FETCH_OPT_VALUE(TYPE_BLOCK)
 								either pos = cmd [
-									OS-matrix-push :state
 									OS-matrix-rotate as red-integer! start as red-pair! cmd - 1
+									OS-matrix-push :state
 									parse-draw as red-block! cmd DC catch?
 									OS-matrix-pop state
 								][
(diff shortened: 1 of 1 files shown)


qtxie:
@Zamlox It's a feature added recently 82f418d .

Zamlox:
indeed, it is.
But there is an issue when a transformation is done before a sub-block, when that sub-block ends, the transformation will not affect following commands. If you run the example from my comment you will see that red rectangle is drawn at 0x0 instead of 10x10.
Specification for sub-blocks says:
"Inside Draw code, commands can be arbitrarily grouped using blocks. Semantics remain unchanged, this is currently just a syntactic sugar allowing easier group manipulations of commands (notably group extraction/insertion/removal). Empty blocks are accepted." Does it mean that transformation still applies after sub-blocks ? If not, then my fix does not apply.

dockimbel:
@Zamlox The sub-block following a matrix-modifying command has special behavior, the modification is applied only to that sub-block. This is a convenient syntactic sugar. If you want the transformation applied to the rest of the commands, you can write it like that:
```
view [base 100x100 draw [ translate 10x10 pen blue box 0x0 30x30] pen red box 0x0 50x50  ]
```

Zamlox:
@dockimbel I thought transformations should behave in same way on sub-block and non sub-blocks.

dockimbel:
Transformation commands are standalone commands, they change the state of the hidden transformation matrix. If immediatly followed by a block, then the transformation applies only to that block (similar to pushing the transformation on stack before entering the block). It is a new feature we added recently, not sure if the documentation mentions it yet.

PR #2236, opened 2016-09-21
FEAT: Keep rebol console on screen when syntax error occur while compiling red from sources

Zamlox:
Rebol console does not remain on screen if there is a syntax error in red sources while compiling it.
This change will keep the console on screen giving a chance to developer to see the error message.
Usage: add ```-show-error``` at the end of command line when compiling:
```
rebol -s red.r -t Windows environment/console/gui-console.red -show-error
```


Zamlox:
Nothing happened for a long time, I will close this.

PR #2262, opened 2016-10-08
Shape dialect implementation

Zamlox:
I have implemented shape dialect for Red with same commands as for R2. Backends: GDI+ and GDI.
I have updated draw arc command for GDI.


modules/view/backends/windows/draw.reds
```diff
@@ -331,6 +353,625 @@ to-gdiplus-color: func [
...
 
+
+radian-to-degrees: func [
+    radians     [float!]
+    return:     [float!]
+][
+    (radians * 180.0) / PI
+]
+
+abs: func [
```
qtxie: This function is defined in float context, we can use it directly. e.g. `float/abs -1.0`.

PR #2270, opened 2016-10-12
Initial font support and new widgets

7hi4g0:
- Font support - **NO UPDATE YET**
 - `font-size`
 - `font-color`
 - `style` - (bold, italic, undeline, strike)
- New widgets - **NO UPDATE YET**
 - `progress`
 - `text`
 - `field`
 - `area`
- Update view
 - `color` for `base`

[Color selection example](http://www.red-lang.org/2016/06/061-reactive-programming.html) working now.

### DIFF
--- a/modules/view/backends/gtk3/gui.reds
+++ b/modules/view/backends/gtk3/gui.reds
@@ -418,7 +418,7 @@ OS-update-view: func [
 		int2	[red-integer!]
 		bool	[red-logic!]
 		s		[series!]
-		hWnd	[integer!]
+		widget	[integer!]
 		flags	[integer!]
 		type	[integer!]
 ][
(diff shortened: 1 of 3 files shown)


7hi4g0:
According to the GTK docs there are only two ways to set the styles on the widgets:
- Using a CSS style through GtkCssProvider
- Implementing the GtkStyleProvider interface

Reading the docs and the source code there's no explicit way to set a widget style directly in an easy manner. Both ways to style listed above have their drawbacks.

The easiest/best way to implement is to use css due to the following reasons:

## GtkCssProvider

### Pros
- CSS is easy
- Supports almost everything **red** needs

### Cons
- Not really made for dynamic styles like **red** needs
- Need to manipulate strings to build the style
- Changing one property needs to rebuild, reload and reparse the whole style OR build one style for each property

## GtkStyleProvider

### Pros
- We can build a dynamic style system
- No need to rebuild the whole style when only one property changed

### Cons
- More complex
- Implementing the interface uses C macros from GObject creating the need for a C implementation or a GObject port to **red**

qtxie:
Great work! :+1: 

Thanks for the comments. I'll have a deep look at it this weekend. 

I use those two libraries as references, maybe you'll find it useful:

https://github.com/andlabs/libui
https://www.eclipse.org/swt/

qtxie:
@7hi4g0 After checking the docs and SWT source code,  I think we can mix the use of CSS style and native GTK functions. 

* We can use the `gtk_widget_override*` functions to modify the style even it has been deprecated. (e.g. `gtk_widget_override_background_color` to change to background color)
* We should create the css provider the first time only when we need it.

modules/view/backends/gtk3/style.reds
```diff
@@ -0,0 +1,196 @@
...
+	face   [red-object!]
+	/local
+		context  [handle!]
+		provider [handle!]
+][
+	provider:	gtk_css_provider_new
+	context:	gtk_widget_get_style_context widget
+
+	gtk_style_context_add_provider context provider GTK_STYLE_PROVIDER_PRIORITY_APPLICATION
+
```
qtxie: We should call `g_object_unref provider` here, so it can be freed when we destroy the widget.

modules/view/backends/gtk3/gui.reds
```diff
@@ -382,12 +387,26 @@ OS-make-view: func [
...
+		sym = area [
+			widget: gtk_text_view_new
+		]
 		true [
 			;-- search in user-defined classes
 			fire [TO_ERROR(script face-type) type]
 		]
 	]
 
+	create-widget-style widget face
```
qtxie: We should create it only when we need it.

modules/view/backends/gtk3/style.reds
```diff
@@ -0,0 +1,196 @@
...
+add-to-string: func [
+	string  [c-string!]
+	format  [c-string!]
+	value   [handle!]
+	return: [c-string!]
+	/local
+		temp [c-string!]
+][
+	temp: g_strdup_printf [format string value]
```
qtxie: we should optimize this function as we use it a lot to construct the CSS style string.
`g_strdup_printf` will create a new string every time, maybe not very efficient.

modules/view/backends/gtk3/style.reds
```diff
@@ -0,0 +1,196 @@
...
+	a: 1.0
+
+	if size = 4 [
+		a: (as-float 255 - color/array1 >>> 24) / 255.0
+	]
+
+	alpha: as c-string! allocate G_ASCII_DTOSTR_BUF_SIZE
+	g_ascii_dtostr alpha G_ASCII_DTOSTR_BUF_SIZE a
+
+	rgba: g_strdup_printf ["rgba(%d, %d, %d, %s)" r g b alpha]
```
qtxie: Is `gdk_rgba_to_string` a better solution?

modules/view/backends/gtk3/style.reds
7hi4g0: I couldn't find the correct place to place face destruction. But it wasn't my priority either.
Will fix that in a next PR.

modules/view/backends/gtk3/style.reds
7hi4g0: I was also worried about `g_strdup_printf` not being efficient, but I used to simplify my work for the time being. I was planning on replacing it for a fixed buffer when I had a clear idea of how many characters would be good as an upper bound.

modules/view/backends/gtk3/style.reds
7hi4g0: Didn't know about it. Will take a look. Thanks :+1:

7hi4g0:
I forgot to.mention the deprecated functions. Sorry!

I was trying to avoid them if I can. Although I believe it's gonna be complex, I prefer to implement the `GtkStyleProvider` interface than use the deprecated functions. That will give us the same functionality and a similar API but it won't be running the risk to stop working in a future release.

If you consider that using the deprecated functions is really necessary, then I can change the code to use them.

modules/view/backends/gtk3/style.reds
7hi4g0: Forgot to mention, I had to use the `g_ascii_dtostr` due to locale issues.

I also noticed that compiling with view support in Linux gives locale problems on the console as well, like I mentioned in Gitter.
I'll open an issue for that.

modules/view/backends/gtk3/style.reds
qtxie: Yes. we can optimise it later.

modules/view/backends/gtk3/style.reds
qtxie: We use this function `OS-destroy-view` in `gui.reds` to free the face resources.

qtxie:
I thought it's easier to do with the deprecated functions, and for those simple functions (e.g. change background color), I don't think they'll really be deprecated in the future. ;-)

But if you already did some research or implementation of the `GtkStyleProvider` interface, then keep working on that. Anyway it's a better solution. :-)

PR #2294, opened 2016-10-28
Improvements on shape and draw

Zamlox:
Following improvements have been done:

- FEAT: Shape - preserve last point among different shape blocks
Last point of a shape block will be saved and used as start point for next shape block within same draw block if no 'move' command used before using one of drawing shape commands. This might be useful when drawing one shape requires changing different graphic attributes (pen, fill, width). In this case shape block can be split and next shape block can continue from where it stopped previous block.

- FEAT: Draw - Improve clip command
Clip command have been improved by allowing to clip a region (given by a path), not only a rectangle. The region can be specified as a block of shape commands, like:
```
view [base 200x200 draw [clip [move 80x120 'arc 1x1 30 30 0 large sweep] fill-pen red box 0x0 199x199]]
```

- FIX: Shape - joint mismatch for shape arc when scaled
Multiple figures does not join exactly when shape is scaled. With this fix, they will join exactly.

- FIX: Shape - make GDI start end points as float!
For GDI, start and end points were ```integer!``` type which might lead to incorrect join for multiple figures. With this fix, they are of type ```float!```

- FIX: Shape - GDI shape arc start point
Removed explicit setting of current position with ```MoveToEx``` and improving ```OS-draw-shape-endpath``` by using ```CloseFigure``` API

- FIX: Shape - GDI+ connect subpaths
```arc``` shape must be part of current path if not preceded by  ```move``` command

### DIFF
--- a/modules/view/backends/windows/draw.reds
+++ b/modules/view/backends/windows/draw.reds
@@ -246,6 +246,11 @@ draw-begin: func [
 	modes/font-color?:		no
 	dc:						null
 
+    last-point?: no
+    prev-shape/type: SHAPE_OTHER
+    path-last-point/x: 0
+    path-last-point/y: 0
+
 	rect: declare RECT_STRUCT
 	either null? hWnd [
 		modes/on-image?: yes
@@ -586,10 +591,6 @@ OS-draw-shape-beginpath: func [
     /local
         path    [integer!]
 ][
-    last-point?: no
-    prev-shape/type: SHAPE_OTHER
-    path-last-point/x: 0
-    path-last-point/y: 0
     either GDI+? [
         path: 0
         GdipCreatePath 0 :path	; alternate fill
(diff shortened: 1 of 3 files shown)


dockimbel:
Nice improvements. I just have one question: shouldn't we rather support `pen, fill, width` commands in Shape dialect, instead of allowing it to be broken into several pieces for the same shape?

Zamlox:
I was thinking the same. Looks like shape dialect tries to mimic SVG path behavior, yet does not allow to change graphic attributes like SVG does (```stroke``` ```stroke-width``` ```fill```) https://www.w3.org/TR/SVG/paths.html#PathElement

dockimbel:
I think the Shape dialect was just an afterthought in Rebol (not present in Draw dialect for years), and SVG compatibility not planned initially also. 

We can do a better job and make our Shape dialect more coherent, by adding those missing features (reusing existing code as much as possible).

Zamlox:
I agree with that and I'm willing to improve it.

pekr:
Remember there is also a Curve dialect in Rebol. We don't imo need to follow Rebol precisely, but the functionality should be supported and I agree, that mostly to being able to handle SVG.

http://www.rebol.com/r3/docs/view/draw-curves.html

dockimbel:
@pekr I think you're confused, those commands are part of the Draw dialect, and Red/Draw supports them already.

PR #2296, opened 2016-11-03
FEAT: STYLIZE function and STYLES keyword for VID

rebolek:


DideC:
I did not test but I wonder: in line 318 and 358 you use just `#()` to create an empty map!.
Don't you need to copy it like with block! ? Or use `make map! 2` like for block! ?

rebolek:
Hm, you’re right. I thought that map! will be initialized every time. Anyway, there is some other problem with this pull request, it can produce mangled layouts in some cases. I have to investigate where the problem is.

PR #2300, opened 2016-11-03
FEAT: STYLIZE function and STYLES keyword for VID

rebolek:


modules/view/VID.red
```diff
@@ -71,7 +71,15 @@ system/view/VID: context [
 				]
 				size-text/with face copy/part mark len
 			]
-			'else [either face/text [size-text face][size-text/with face "X"]]
+			'else [
+				either face/text [
+					size: size-text face
+					if find [button radio check] face/type [size/x: size/x + size/y]
```
rebolek: IIRC this is leftover from some older View version on which I based this PR.

modules/view/VID.red
```diff
@@ -318,12 +364,7 @@ system/view/VID: context [
 		global?: 	  yes								;-- TRUE: panel options expected
 		
 		cursor:	origin: spacing: pick [0x0 10x10] tight
-		
-		opts: object [
-			type: offset: size: text: color: enable?: visible?: selected: image: 
-			rate: font: flags: options: para: data: extra: actors: draw: now?: none
-		]
-		
+
```
rebolek: You're right, they aren't. I have a fix in my branch (2109e7 from 2016-11-03), but it is missing from this PR. I am not very familiar with PRs, can I fix it here or should I make a new PR?

modules/view/VID.red
```diff
@@ -289,7 +297,46 @@ system/view/VID: context [
...
+		/styles 				"Use an existing styles list"
+			css 	  [map!] 	"Styles list"
+		/local tmp face args parent style
+		/extern face-options
+	] [
+		local-styles: case [
+			master 	(system/view/VID/styles)
+			styles 	(copy css)
+			true 	(make map! 2)
+		]
```
rebolek: I am not sure why I am using parens in `case`, I've been doing this for many years. I will change it (and try to rewire myself to blocks here :) )

modules/view/VID.red
dockimbel: You can fix it with an additional commit to this PR.

rebolek: Thanks, I'll try.

rebolek:
I've tried to fix all things mentioned in review and also base it on latest VID version. Hopefully it is OK now.

modules/view/VID.red
```diff
@@ -184,7 +176,7 @@ system/view/VID: context [
 				| 'font		  (opts/font: make any [opts/font font!] fetch-argument obj-spec! spec)
 				| 'para		  (opts/para: make any [opts/para para!] fetch-argument obj-spec! spec)
 				| 'wrap		  (opt?: add-flag opts 'para 'wrap? yes)
-				| 'no-wrap	  (opt?: add-flag opts 'para 'wrap? no)
+				| 'no-wrap	  (add-flag opts 'para 'wrap? no opt?: yes)
```
geekyi: Nit pick: I don't know why you did this but it doesn't this make sense?
After looking at the code for a while, it still didn't.. so I may be wrong about it..

rebolek:
@greggirwin I need to check it against newest Red, I believe that there are some merge conflicts now. I'll take a look at it and let you know.

PR #2315, opened 2016-11-11
Draw and shape improvements

Zamlox:
Following improvements have been done:

- 	**_FEAT: Draw - added clip mode_**: an optional parameter have been added to draw ```clip``` command. It can have following values: ```replace```, ```intersect```, ```union```, ```xor``` and ```exclude```. By default the value is ```replace```. Once set a clip mode, it will persist even if ```clip``` command is called without this optional parameter. Clip mode changes only when explicitly set.

- 	**_FEAT: Draw - added matrix-order command_**: an additional command have been added to ```draw``` dialect: ```matrix-order```. It can have following values: ```append```, ```prepend```. It tells the system how to order matrix multiplications for transformation commands. ```append``` will add the transformation matrix on the right side of multiplication operation, ```prepend``` will add the transformation matrix on the left side of multiplication 

- **_FEAT: Shape - added pen, fill-pen, line-width, line-join, line-cap_**: added following commands to ```shape``` dialect: ```pen```, ```fill-pen```, ```line-width```, ```line-join``` and ```line-cap```

- **_FEAT: Shape - added pen, fill-pen, line-width, line-join, line-cap_**: added proper initialization for clip mode for GDI

- **_FIX: Draw - clip coordinates_**: lower coordinate was bigger by 1 pixel

- **_FIX: Draw - create pen with width in world units_**: creating pen with pixel coordinates does not allow line width to scale when using scale transformation

- 	**_FIX: Draw - allow float for line-width_**: allow ```float!``` type parameter for ```line-width``` command

- **_FIX: Draw - rectangle coordinates_**: minor rectangle coordinates fix

-  **_FIX: Draw - radial gradient_**: currently radial gradient is not reproducing same output as in C++ or SVG. List of points must be subtracted from 1.0 and reversed when passed to ```GdipSetPathGradientPresetBlend``` GDI+ API function. Samples given at: https://gist.github.com/Zamlox/5d78d5a7b6a47acf0b8d30c2ee44bc82 will give same output when Red fixed by this commit.

- 	**_FIX: Shape - take out limitation of 1000 points at end of path_**: Taken out limitation of 1000 points. Not needed.

### DIFF
--- a/modules/view/backends/windows/draw.reds
+++ b/modules/view/backends/windows/draw.reds
@@ -1148,12 +1148,12 @@ OS-draw-box: func [
 			gdiplus-draw-roundbox
 				upper/x
 				upper/y
-				lower/x - upper/x
-				lower/y - upper/y
+				lower/x - upper/x + 1
+				lower/y - upper/y + 1
 				rad
 				modes/brush?
 		][
-			RoundRect dc upper/x upper/y lower/x + 1 lower/y + 1 rad rad
+			RoundRect dc upper/x upper/y lower/x lower/y rad rad
 		]
 	][
 		either GDI+? [
@@ -1165,18 +1165,18 @@ OS-draw-box: func [
 					modes/gp-brush
 					upper/x
 					upper/y
-					lower/x - upper/x
-					lower/y - upper/y
+					lower/x - upper/x + 1
+					lower/y - upper/y + 1
 			]
 			GdipDrawRectangleI
 				modes/graphics
 				modes/gp-pen
 				upper/x
 				upper/y
-				lower/x - upper/x
-				lower/y - upper/y
+				lower/x - upper/x + 1
+				lower/y - upper/y + 1
 		][
-			Rectangle dc upper/x upper/y lower/x + 1 lower/y + 1
+			Rectangle dc upper/x upper/y lower/x lower/y
 		]
 	]
 ]
--- a/modules/view/backends/windows/draw.reds
+++ b/modules/view/backends/windows/draw.reds
@@ -1777,7 +1777,7 @@ OS-draw-grad-pen: func [
 		color/value: to-gdiplus-color clr/array1
 		next: head + 1 
 		if TYPE_OF(next) = TYPE_FLOAT [head: next f: as red-float! head p: f/value]
-		pos/value: as float32! p
+		pos/value: as float32! ( 1.0 - p )
 		if next <> head [p: p + delta]
 		head: head + 1
 		color: color + 1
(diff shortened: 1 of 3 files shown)


modules/view/backends/windows/draw.reds
```diff
@@ -1066,12 +1067,23 @@ OS-draw-fill-pen: func [
...
-	width [red-integer!]
+	width [red-value!]
+    /local 
+        width-i     [red-integer!]
+        width-f     [red-float!]
+        width-v     [float32!]
 ][
-	if modes/pen-width <> width/value [
-		modes/pen-width: width/value
+    either TYPE_OF(width) = TYPE_INTEGER [
```
qtxie: Here we can call function `get-float32` in `draw.red`.

modules/view/backends/windows/draw.reds
```diff
@@ -1137,12 +1149,12 @@ OS-draw-box: func [
 			gdiplus-draw-roundbox
 				upper/x
 				upper/y
-				lower/x - upper/x
-				lower/y - upper/y
+				lower/x - upper/x + 1
```
qtxie: Why plus 1 here?

modules/view/backends/windows/draw.reds
Zamlox: thanks, I will update.

modules/view/backends/windows/draw.reds
Zamlox: third and fourth parameters of ```gdiplus-draw-roundbox``` are ```width``` and ```height``` and ```lower``` and ```upper``` are absolute coordinates.
Let's say we have a rectangle of 2 pixels defined by absolute coordinates (1x1 2x2). This means upper: 1x1, lower: 2x2. The width and height are 2 pixels: 2 - 1 + 1 (lower/x - upper/x + 1, lower/y - upper/y + 1) and the rectangle starts at 1x1.

modules/view/backends/windows/draw.reds
```diff
@@ -1075,7 +1075,7 @@ OS-draw-line-width: func [
 ][
     either TYPE_OF(width) = TYPE_INTEGER [
```
qtxie: Can we just use `get-float32 width` instead of the whole `either` expression?

Zamlox: ```width [red-value!]``` can be of type ```red-integer!``` or ```red-float!```. ```get-float32``` works only on ```red-integer!``` types. How can I convert safely a ```red-value!``` to ```red-integer``` if the actual value is a ```red-float!``` ?

qtxie: This in Red:
```
red-integer!: alias struct! [
	header 	[integer!]
	padding	[integer!]
	value	[integer!]
	_pad	[integer!]	
]
```
is equivalent to this in C:
```
typedef struct {
	int header;
	int padding;
	int value;
	_pad
} *red_integer
```
So `width` is just a pointer, casts it to another type does not change it underlying value.
And we do handle `red-float!` in function `get-float32`:
https://github.com/red/red/blob/master/modules/view/draw.red#L127

Zamlox: all right. I will replace that code with ```get-float32```
Thanks.

Issue #2380, opened 2016-12-26
Font settings are incorrect after VID merges


endo64:
Size and Color facets get appended when VID merges the settings:

```
f1: make font! [name: "Arial" size: 24 color: red]
view [f2: text "Test" font f1 font-color blue font-size 22]
? f2/font
size  block!    [24 22]  ;<< should be 22
color  block!    [255.0.0 0.0.255]  ;<< should be 0.0.255

view [f3: text "Test" font f2/font font-color green font-size 20]
? f3/font
size             block!    [24 22 20]  ; << should be 20
color            block!    [255.0.0 0.0.255 0.255.0]  ; << should be 0.255.0
```

meijeru:
As far as I can see, this is because `add-flag` is used to handle the setting, which always forces the facet to be a block. For `font/color` and `/size` another function should be used.

DideC:
Already posted as #2269

dockimbel:
@DideC Well done!

Issue #2269, opened 2016-10-12
VID does not handle "blockable" font! (or para!) field reliably


DideC:
In `font!` object, only `/style` field can be `none!`, a `word!` or a `block!`.
But the `system/view/VID/add-flag` function that handle the `word!` to `block!` transformation, does it for any field of the object it manipulate.
So unexpected results can occurs:

```
view [
    below
    text "Font default"
    text "Font 18" font-size 18
    text "Font 36" font-size 36
    f-t: text "Font 8, 18 and 36" font-size 8 font-size 18 font-size 24
   do [? f-t/font]
]
```

Even if it's not a "correct VID code", it is valid and does not produce errors.
`para!` object could also be affected.

dockimbel:
@DideC Well done!

DideC:
Thanks. Not too hard aniway ;-)

PR #2401, opened 2017-01-05
'add-flag only handle font/style as block!

DideC:
To my understanding, only `font/style` facet needs to be handled as a `block!` if several styles words are provided.
No other facet in `font!`, nor `para!` object has to handle more than one value.
With the exception of `/state` and `/parent` maybe, but there are managed by low level code AFAIK, not by VID functions.

It must fixe #2269 and fixe #2380


dockimbel:
Good job! (both for identifying and fixing it!)

PR #2404, opened 2017-01-06
Complex pen

Zamlox:
Implementation of complex pen as described [here](https://github.com/qtxie/red/wiki/Create-Complex-Pen-in-Draw). Only Windows platform GDI+ backend


Zamlox:
Yes, indeed. They are powerful. I like most the 'pattern' because it allows to define patterns recursively using  draw commands.

Oldes:
cool!

DideC:
Need a good demo to demonstrate it now !

Zamlox:
For now, only the test is available: tests/complexpen-test.red
Any other ideas for a good demo ?

PR #2407, opened 2017-01-09
Small  libRed improvement

Oldes:


dockimbel:
`bool` is not part of C language, it's C++.

PR #2446, opened 2017-02-20
FEAT: added support for handle! datatype

Oldes:


runtime/datatypes/handle.reds
```diff
@@ -0,0 +1,227 @@
+Red/System [
+	Title:   "Handle! datatype runtime functions"
+	Author:  "Nenad Rakocevic, Oldes"
+	File: 	 %integer.reds
```
x8x: should be:
`File: %handle.reds`

PR #2449, opened 2017-02-22
FIX: Replace function infinite loop problem

endo64:
This PR fixes infinite loop problem for `replace/all "abab" #"a" #"a"`


dockimbel:
Great work, thanks!

PR #2450, opened 2017-02-23
FIX: Draw - radial gradient colors position

Zamlox:
There is an issue with colors position for radial gradient when they are not specified in draw dialect. This commit will fix that issue.


qtxie:
Thanks for the fix. 👍

PR #2454, opened 2017-02-27
Remove unnecessary /local line in fetch-help.

greggirwin:


greggirwin:
Curses! Sorry I didn't check that.

greggirwin:
Using `function` *and* `.local` makes for unclear code IMO. We could certainly add a comment, but what should the style guide say about this?

dockimbel:
`function` cannot accurately determine which words are local or not. It uses a simple heuristic based on set-words collection. Though, there are cases where words are set differently, like in Parse rules with `copy`, `set` or `collect`, which cannot be captured using the set-word method. Moreover, some words are only created at runtime, but still can be locally bound (if the user desires so), e.g.:
```
function [/local a][do bind "a: 123" 'a]
```

So, `function` can do a good basic job, but abilitity to manually add local words anyway using `/local` is necessary in Redbol languages.

greggirwin:
Of course. Not saying we should remove that feature, but mixing `function` and `/local` seems like a bad idea.

dockimbel:
I don't see any issue with that. I used it in several places in the Red runtime library, when `function` alone is not enough, but I don't want to manually have to declare all locals.

dockimbel:
So, do you want to submit a change to this PR, adding back only local words not used as set-words?

greggirwin:
I can just close this PR. My thought is about best practices in general. Combining `function` and `/local` in a single func has two mechanisms, that work differently, doing the same job. I'd like to hear other input though. Admittedly, I had never used it myself when it confused me here. My instinct says that it will add mental overhead, but that's only a feeling (and my botched patch as evidence).

dockimbel:
*Combining function and /local in a single func has two mechanisms, that work differently, doing the same job*

Not exactly the same job:
* `function` will gather set-words only and mark them as local.
* `/local` allows the user to declare any word as local.

dockimbel:
I have reduced the list of local words in the `fetch-help` function to only ones which are never used as set-words. I'll close this PR now, but we can move that discussion somewhere else.

greggirwin:
By "same job" I meant capturing locals.

Issue #2435, opened 2017-02-13
help object! needs to be formatted for readable output.


gltewalt:
`help` needs to specifically format the output for object! to show object-name => description, or object-name and attributes => description.
Currently `help object!` appears to emit the source for all objects in the global context.

gltewalt:
Line 87 can be changed to:
`print-out out [tab pad form :w 15 ": " mold words-of get/any w]`

`help system` will then emit:

```
red>> help system
`system` is an object! of value:
     version          string!   0.6.1 
     build            object!   [date config] 
     words            object!   [datatype! unset! none! logic! block! paren!... 
     platform         function! Return a word identifying the operating system 
     catalog          object!   [datatypes actions natives errors] 
     state            object!   [interpreted? last-error trace?] 
     modules          block!    [] 
     codecs           map!      #(png: make object! [title: "" name: 'PNG mi... 
     schemes          object!   [] 
     ports            object!   [] 
     locale           object!   [language language* locale locale* months da... 
     options          object!   [boot home path script args do-arg debug sec... 
     script           object!   [title header parent path args] 
     standard         object!   [header error] 
     lexer            object!   [throw-error make-hm make-msf make-hms make-... 
     console          object!   [prompt history limit catch? count ws gui? r... 
     view             none!     none 
     reactivity       object!   [relations stack queue debug? do-safe eval-r... 
```

Although, the results are then truncated with "..." by Lines 209 through 211

dockimbel:
@gltewalt Feel free to submit a PR.

greggirwin:
The above change won't work, as it applies to all other types, which don't all support `reflect`. We could add a case at line ~85 to handle objects
```
object? get/any w [
    print-out out [tab pad form :w 15 mold words-of get/any w]
]
```
But maps also support words-of, which might be nice. And if we want it to optimize for 80-char width, as later code in help does (`emit` + ...), that should be a clear decision.

greggirwin:
```
any [object? get/any w  map? get/any w] [
    print-out out [tab pad form :w 15 mold words-of get/any w]
]
```

greggirwin:
Hmmm, the `tab*` vars are not /local in `fetch-help`, though others are. `Fucntion` is used, so they're still captured, but that means the `/local` line isn't needed at all. Worth cleaning up?

greggirwin:
Quick PR done for that /local bit. Seems worth it, since it confused me at a glance.

gltewalt:
[...]

PR #2455, opened 2017-02-28
Added object? case statement to help.red

gltewalt:
Line 86


geekyi:
Any example output where this is different from the normal `fetch-help` ?

gltewalt:
Issue #2435

I didn't change it to account for maps

gltewalt:
Should I redo it to include Gregg Irwins code sample that includes map?
Let Irwin add it in his PR?

PR #2459, opened 2017-03-03
FIX: Convert Source File Path to Local Path in Error

nc-x:


compiler.r
```diff
@@ -138,7 +138,7 @@ red: context [
 			either word? err [
 				join uppercase/part mold err 1 " error"
 			][reform err]
-			"^/*** in file:" mold script-name
+			"^/*** in file:" mold to-local-file script-name
```
dockimbel: `mold to-local-file` will add unwanted double quotes, so `mold` should be removed in such case.

nc-x: @dockimbel done

nc-x:
@dockimbel Done, and I merged the two commits so as to not pollute the pull request (though I think the commits could have been squashed when you merged the PR, but I was not sure if it was possible, so I did it myself). But the appveyor and travis builds seem to be stuck and need to be manually stopped (I think).

dockimbel:
`when you merged the PR` I haven't merged it.

nc-x:
`when you merged the PR`
 *merge 😉 😄

PR #2316, opened 2016-11-13
Small improvement of help message and addition of a left logical shift operator

lkppo:

PR #2464, opened 2017-03-04
Small improvement of help message

lkppo:
- Increases padding of WHAT output to suite more function names ;
- Help message of 'as-color', 'as-ipv4', 'shift-right', 'shift-left',
  'shift-logical' functions ;
- Fix some typo.

Hello Nenad,

I spend some time to rehash my previous PR.

I hope this time it will be good for you.

Have a nice day,

-- 
LkpPo

### DIFF
--- a/environment/actions.red
+++ b/environment/actions.red
@@ -148,7 +148,7 @@ remainder: make action! [[
 round: make action! [[
 		"Returns the nearest integer. Halves round up (away from zero) by default"
 		n		[number! time! pair!]
-		/to		"Return the nearest multiple of the scale parameter"
+		/to		"Returns the nearest multiple of the scale parameter"
 		scale	[number!] "Must be a non-zero value"
 		/even		"Halves round toward even results"
 		/down		"Round toward zero, ignoring discarded digits. (truncate)"
@@ -170,7 +170,7 @@ subtract: make action! [[
 ]
 
 even?: make action! [[
-		"Returns true if the number is evenly divisible by 2"
+		"Returns TRUE if the number is evenly divisible by 2"
 		number 	 [number! char!]
 		return:  [number! char!]
 	]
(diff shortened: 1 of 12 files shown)


lkppo:
Hello Nenad,

Have you seen my PR is ready?

Regards,

-- 
LkpPo

geekyi:
Overall looks good to me, only changes in doc strings. However, I have some concerns:

## Suggestion

>  - Increases padding of WHAT output to suite more function names ;
>  - Help message of 'as-color', 'as-ipv4', 'shift-right', 'shift-left', 'shift-logical' functions ;
>  - Fix some typo.

You have 3 different things which should be in 3 different commits IMHO, because [Git commits should be atomic](https://www.google.com/search?q=what+are+atomic+git+commit)
I admit what is atomic is subjective; given also in this case, they are all just typographical fixes.
However, if you go by the rule of revertible changes, I think you can split to at least 2 commits:
- Increases padding of WHAT output to suite more function names
- Help messages for some words


## For the future
This commit also has some stylistic fixes:
- Plurals: `Returns` instead of `Return`
- Casing: `TRUE` instead of `true`

So my question is in general for @all: *What is the criteria, or the style guide to refer to?* I know we have the bare minimum of a style guide (use tabs of width 4)
If it's not a well defined, enforceable set of rules, such changes may just end up as noise

lkppo:
At start, I had three or four commits. Since the start, Nenad has said he accepted it. I will not split my PR in two commits because it's the third time I rework it for reasons independent of my will, and I'm a bit bored (not angry).

Maybe it's naught, it's not a good PR at all or not in step with the preoccupations of the moment, and removing it will be a better accomplishment.

geekyi:
Yeah, probably preoccupied as the release is so close.
Sorry, I had forgot about the previous pull-request #2316 
From there:
> Another change that will be rejected is adding help strings to shift-right, shift-left, shift-logical routines, as those routines are not meant for end users, but just as intermediary routines for the shifting operators. Therefore, they should not be documented.

I'd also wondered about a style guide as your PR seemed relevant.. sorry for hijacking it for that

lkppo:
I forgotten that and reintroduce them by mistake. I got my feet in the carpet with a rebase. So I will just revoke my PR. I 'm not in time.

Issue #2128, opened 2016-07-20
help should mention string argument form


dander:
Using `help` with a string argument is a very useful tool for language discovery because it aids in finding related methods. It would be really worthwhile to add mention of it in the default help information.

perhaps

```
red>> help
Use HELP or ? to see built-in info:

    help insert
    ? insert

To see all functions containing a specific substring:

    ? "file"

To see all words of a specific datatype:

    ? native!
    ? function!
    ? datatype!

Other useful functions:

    ?? - display a variable and its value
    probe - print a value (molded)
    source func - show source code of func
    what - show a list of known functions
    about - display version number and build date
    q or quit - leave the Red console
```

I'm not sure if this is the best/correct terminology, but I think "file" is a nice example. I'd be happy to submit a PR if desired.

pekr:
I am really glad, someone found the string form useful. But - I am against that. What I propose is - please, pretty pretty please, turn it into R2 functionality - no string needed, just type whatever you want and please find it as a substring, like a grep. Ppl will sort results out. Simply addition of - please use "file" will cure it only partially. It really affects my live console sessions significantly ...

dander:
not ever being a Rebol user, I was never aware of that functionality :)

greggirwin:
R2 only searches by name, while R3 also searches by doc string.

pekr:
I really want to be able to type just: help to- and get all related functions reported, without the need to type it as a: help "to-" .... most ppl will imo miss it, even if it would be mentioned in a docstring - it pushes you to use a special form. You normally search for: help insert, and not: help "insert" too ....

greggirwin:
@dander and @pekr, can we close this ticket with the latest help that's in place?

geekyi:
Nice. Now both `? file` and `? "file"` works, giving different output.
I think this can be closed, because @pekr's wish of `? to-` also works (`? "to-"` incidentally, gives the same output in this case)

Thanks so much Gregg, Oldes and anyone else I forgot to mention!

Issue #2127, opened 2016-07-20
help output format for functions refinements: USAGE is misleading


dander:
The usage format displayed by help shows each refinement coming after the function arguments, which is not valid Red syntax.

ex:

```
red>> help insert

USAGE:
    insert series value /part length /only /dup count

DESCRIPTION:
     Inserts value(s) at series index; returns series past the insertion. 
     insert is of type: action!

ARGUMENTS:
     series  [series! bitset! map!]
     value  [any-type!]

REFINEMENTS:
     /part => Limit the number of values inserted.
         length  [number! series!]
     /only => Insert block types as single values (overrides /part).
     /dup => Duplicate the inserted values.
         count  [number!]
```

It would be clearer to remove the refinements altogether (since there is a refinement section below), or to have them on separate lines.
possibly:

```
insert series value
insert/part series value length
...
```

or

```
insert series value
      /part length
      /only
...
```

Oldes:
@dockimbel what is your opinion?

Oldes:
With the above modification the output is:
```
>> ? insert
USAGE:
    INSERT series value

DESCRIPTION:
    Inserts value(s) at series index; returns series past the insertion.
    INSERT is an action! value.

ARGUMENTS:
    series     -> [series! bitset! map!]
    value      -> [any-type!]

REFINEMENTS:
    /part      -- Limit the number of values inserted
       length  -> [number! series!]
    /only      -- Insert block types as single values (overrides /part)
    /dup       -- Duplicate the inserted values
       count   -> [number!]
```

greggirwin:
I agree that we could leave out refinements in the usage string. Building combinations of them is certainly doable, but may be more confusing than helpful, once you get beyond 2 refinements.

greggirwin:
@dander, would you please check this with the latest version of `help` and sign off on it if it looks good.

dander:
Oh, yeah, I have been loving the new `help`. My complaint for this issue was about the refinements on the usage string. As you mentioned above, it has been addressed. Thanks to you both for all the work on improving this!

Issue #2123, opened 2016-07-17
HELP adds trailing dot if func doc string ends with a dot.


greggirwin:
```
red>> dirize: func [
[        "Returns a copy of a filespec with a trailing path marker."
[        path [file! string! url!]
[    ][
[        ; Always return a copy of the path.
[        path: copy path
[        either slash = last path [path][append path slash]
[    ]
== func [
    {Returns a copy of a filespec with a trailing path marker.} 
    path [file! string! url!]
][path: copy path 
    either s...
red>> ? dirize

USAGE:
    dirize path

DESCRIPTION:
     Returns a copy of a filespec with a trailing path marker.. 
     dirize is of type: function!

ARGUMENTS:
     path  [file! string! url!]

REFINEMENTS:
```

dockimbel:
That is the convention we adopted, docstrings do not need an ending dot, as the end is delimited by the closing double quote. Adding a dot when printing it is up to the docstring processor, which is currently only HELP function.

greggirwin:
I know it's the convention (some doc strings may need cleaning), but some people may want to write their doc strings to be more grammatically correct. If they do, this will look like an error on HELP's part.

greggirwin:
Unless you want to be strict in this convention, no matter whose code it is. i.e. is the convention that they MUST NOT have an ending dot, or that they MAY have one?

dockimbel:
Making HELP a little smarter about that wouldn't hurt. You can make a PR to take an eventual ending dot into account. Though, the convention stays for red/red repo.

dockimbel:
Also, maybe rewriting the HELP code would be nice too, as it was contributed by someone, by directly porting the R2 version, not taking into account R3's improvements. I tried to put back some of those improvements, but the whole HELP code is not very well architectured, a cleaner and shorter version would be welcome.

greggirwin:
I'm good with Red deciding its own repo rules. :^)

Oldes:
@dockimbel I can give it a try to rewrite the current help.

greggirwin:
I stopped working on my version when Ingo's was included with Red. I can post it somewhere if you want to collaborate Oldes.

greggirwin:
Or I could tackle this so you can focus on deeper issues from the R/S side. You've been doing a lot there lately.

Oldes:
It's ok.. I already have it.. just some clean up needed.

rebolek:
I have help parse done for `helpr`, VID help, it would would probably make sense to share the code.

PR #2497, opened 2017-03-24
FEAT: help recoded (issue #2123, issue #2127, issue #2128)

Oldes:
This modification provides more Rebol3 like help output.

### DIFF
--- a/environment/console/help.red
+++ b/environment/console/help.red
@@ -1,187 +1,17 @@
 Red [
 	Title:	"Console help functions"
-	Author:	["Oldes" "Ingo Hohmann" "Nenad Rakocevic"]
+	Author:	["Oldes" "Gregg Irwin" "Ingo Hohmann" "Nenad Rakocevic"]
 	File:	%help.red
 	Tabs:	4
-	Rights:	"Copyright (C) 2014-2017 Oldes, Ingo Hohmann, Nenad Rakocevic. All rights reserved."
+	Rights:	"Copyright (C) 2014-2017 All Mankind. All rights reserved."
 	License: {
 		Distributed under the Boost Software License, Version 1.0.
 		See https://github.com/red/red/blob/master/BSL-License.txt
 	}
 ]
 
-ctx-help: context [
-	datatypes: [
-		action!     "datatype native function (standard polymorphic)."
-		binary!     "string series of bytes."
-		bitset!     "set of bit flags."
-		block!      "series of values."
-		char!       "8bit and 16bit character."
-		datatype!   "type of datatype."
-		date!       "day, month, year, time of day, and timezone."
-		float!      "64bit floating point number (IEEE standard)."
-		email!      "email address."
-		error!      "errors and throws."
-		event!      "user interface event (efficiently sized)."
-		file!       "file name or path."
-		function!   "interpreted function (user-defined or mezzanine)."
-		get-path!   "the value of a path."
-		get-word!   "the value of a word (variable)."
-		handle!     "arbitrary internal object or value."
-		hash!       "."
-		image!      "RGB image with alpha channel."
-		integer!    "32 bit integer."
-		issue!      "identifying marker word."
-		library!    "external library reference."
-		lit-path!   "literal path value."
-		lit-word!   "literal word value."
-		logic!      "boolean true or false."
-		map!        "name-value pairs (hash associative)."
-		money!      "high precision decimals with denomination (opt)."
-		native!     "direct CPU evaluated function."
-		none!       "no value represented."
-		object!     "context of names with values."
-		op!         "infix operator (special evaluation exception)."
-		pair!       "two dimensional point or size."
-		paren!      "automatically evaluating block."
-		path!       "refinements to functions, objects, files."
-		percent!    "special form of decimals (used mainly for layout)."
-		pair!       "two dimensional point or size."
-		port!       "external series, an I/O channel."
-		refinement! "variation of meaning or location."
-		routine!    "native function."
-		set-path!   "definition of a path's value."
-		set-word!   "definition of a word's value."
-		string!     "string series of characters."
-		struct!     "native structure definition."
-		tag!        "markup string (HTML or XML)."
-		time!       "time of day or duration."
-		tuple!      "sequence of small integers (colors, versions, IP)."
-		typeset!    "set of datatypes."
-		unset!      "no value returned or set."
-		url!        "uniform resource locator or identifier."
-		vector!     "high performance arrays (single datatype)."
-		word!       "word (symbol or variable)."
-	]
-
-	clip-str: func [str] [
-		if (length? str) > 80 [str: append copy/part str 80 "..."]
-		str
-	]
-
-	form-val: func [val] [
-		val: case [
-			any-function? :val [
-				spec: spec-of :val
-				either any [
-					string? val: spec/1
-					string? val: spec/2	;-- attributes block case
-				][	
-					if #"." <> last val [append val #"."]
-					val
-				][
-					clear find spec /local
-					val: trim/lines mold spec
-				]
-				;@@ TODO: wrap long description lines
-			]
-			any-block? :val [ form reduce ["length:" length? val] ]
-			image?     :val [ form reduce ["size:" val/size] ]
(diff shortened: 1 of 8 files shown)


qtxie:
Looks nice! 👍

Oldes:
I removed datatype definitions as these were taken from Rebol source. Could be added back if someone will write what should be printed when one uses: `help datatype!`.

The `help` is again re-coded to reduce it's size and efficiency. I removed `fetch-help` function and instead added refinement `into` in the main `help` function. With this refinement it is possible to get help result into string instead of being printed in the console.

Block listed in the object is now showing partial content instead of block length information as it is in Rebol.

Oldes:
I modified the output of the help function... main difference is, that it crop description lines in object listing according width of the console (gui version only so far) and it correctly handle multiple strings in function descriptions... so:
```
my-function: func [
	"This is the best function"
	"It can handle almost everything!"
	arg        "It takes this argument"
	           "Don't know, why I write this?"
	/path      "Also path can be used"
	 p [path!] "Use this to specify location or whatever"
	return: [any-string!]
][]

? my-function
```

will show:
```
USAGE:
    MY-FUNCTION arg

DESCRIPTION:
    This is the best function. 
    It can handle almost everything! 
    MY-FUNCTION is a function! value.
    Returns value of type: [any-string!]

ARGUMENTS:
    arg        [any-type!] It takes this argument. Don't know, why I write this?

REFINEMENTS:
    /path      <- Also path can be used
       p       [path!]    Use this to specify location or whatever.
```

Oldes:
Just for comparison.. if used without the `return:` part, which is not supported in Rebol, this would be output in Rebol 2:
```
USAGE:
    MY-FUNCTION arg /path p

DESCRIPTION:
     This is the best function
     MY-FUNCTION is a function value.

ARGUMENTS:
     arg -- It takes this argument (Type: any)

REFINEMENTS:
     /path -- Also path can be used
         p -- Use this to specify location or whatever (Type: path)
```
and this in Rebol 3:
```
USAGE:
        MY-FUNCTION arg /path p

DESCRIPTION:
        This is the best function
        MY-FUNCTION is a function value.

ARGUMENTS:
        arg -- Don't know, why I write this?

REFINEMENTS:
        /path -- Also path can be used
                p -- Use this to specify location or whatever (path!)
```

environment/console/help.red
```diff
@@ -59,226 +36,305 @@ Other useful functions:
...
-					]
-					any [
-						/local to end
-						| set w [word! | lit-word! | get-word!] (prin-out out " " prin-out out w)
-						| set w refinement! (prin-out out " " prin-out out mold w)
-						| skip
-					]
+
+		catch [case/all [
+			if unset? :word [									;-- HELP with no arguments
```
dockimbel: Seems like that `if` is a leftover of older code, it should be removed.

environment/console/help.red
```diff
@@ -59,226 +36,305 @@ Other useful functions:
...
+				buffer: insert buffer HELP-USAGE
+				throw true
+			]
+
+			if word? :word [
```
dockimbel: Same remark for this `if` as for previous one.

environment/console/win32.reds
```diff
@@ -197,37 +199,48 @@ fd-read: func [
 		key 	[key-event!]
 		n	 	[integer!]
 		keycode [integer!]
+		size    [red-pair!]
 ][
 	n: 0
 	while [true] [
```
dockimbel: `while [true]` => `forever`

dockimbel:
> it correctly handle multiple strings in function descriptions

Multiple strings are allowed, but they are not meant for splitting the same docstring into multiple pieces (that would be a waste), it is meant to allow for extra informations to be encoded in the future in additional strings (we could then define a convention for the meaning of the 1st, 2nd, ... docstrings). It's an option for future use.

environment/console/win32.reds
Oldes: The `while [true]` part is from the original... also... this part of code should not be part of this pull request probably.. it leaked there from the `system/console/limit` changes here: https://github.com/red/red/pull/2514

Oldes: Although the help already depends on the value from the `system/console/size`

Oldes:
> Multiple strings are allowed, but they are not meant for splitting the same docstring into multiple pieces (that would be a waste). It is meant to allow for extra information to be encoded in additional strings (we could then define a convention for the meaning of the 1st, 2nd, ... docstrings). It's an option for future use.

So what is expected output for such a situation? Above are 3 versions.. there can be also 4. version where the multiple doc strings are printed on own lines.

Oldes:
@dockimbel I was also thinking about different output for this situation:
```
quit: func [
	"Stops evaluation and exits the program"
	/return status	[integer!] "Return an exit status"
][
	quit-return any [status 0]
]
```
It would now show:
```
REFINEMENTS:
    /return
       status  [integer!] Return an exit status.
```
Maybe it would be better to detect such a situation and display:
```
REFINEMENTS:
    /return    <- Return an exit status.
       status  [integer!] 
```
Or maybe the change should be done in the function source... there are many undocumented or ill documented functions.

Oldes:
Btw... there is a bug in above situation as `/return` is detected as a `returning` result . I will fix it.

dockimbel:
As @greggirwin version was merged, it would be good to port the fixes provided by this PR for the mentioned tickets above to the current version in master. @Oldes I'm closing this PR now, feel free to submit a new one with extra fixes/features required for closing those tickets.

Oldes:
It looks that the most important difference - the real console size - was incorporated by @greggirwin already, so I'm probably fine.

In the new version I don't like the pipe character, especially when there is no following content and long name, like here:

image

Oldes:
@dockimbel @greggirwin from the mentioned issues, only issue #2127 is not fixed with current version.

greggirwin:
I'll review comments here and in other tickets. Changing the | is easy, if we just want a different char. Also easy to leave it out if there is no value. 

The two things on the top of my list are to fix alignment/padding issues and add the ability to get help output as a string, if we still want that feature. On padding, the big question is how important it is to have a fully polished output, which means we'd have to find the longest name in any list and format based on that. Not hard, just overhead.

greggirwin:
One of the downsides with padding to the longest name, especially with system/words, is that one really long name pushes everything way out.

dockimbel:
@greggirwin In such cases, I usually limit the padding to a fixed size, truncate the long names and add a "..." sequence at tail.

PeterWAWood:
You could make use of Red's Unicode pedigree and save a couple of characters by using …  - #"^(2026)"

dockimbel:
@PeterWAWood Good suggestion, though a special case would need to be made for the CLI console on Windows, as such character would not display properly.

greggirwin:
I'll see how truncating names looks. The problem I forsee is that asking for help and getting only most of name may not seem all that helpful.

greggirwin:
I considered using the ellipsis char in the other truncation scenario, but didn't, for the reason Doc cited.

qtxie:
@greggirwin 
> get help output as a string

Yes, we need it as the VSCode plugin use it when showing the help window.

greggirwin:
I'll make that a priority then Qingtian. Does this also affect what we want, WRT #2127?

greggirwin:
Oldes, one of the things I like about `|` is that it makes the output table-like, though I agree alignment is an issue no matter what. Something like `==` doesn't look as nice to me. `:` is nice, but then there is no hope of making the output loadable, if we wanted to do that. I have added back buffered output here, and experimenting made me think about how we might use that capability.

greggirwin:
I'm adding padding, so object words align nicely, but a ripple effect for `system/words` is that the first column may be wider than our old fixed-width model. That means the console size check can't be static anymore, but will be dependent on the current output (longest word). The console nicely rewraps as you resize, but any new output will have the same issue. And it affects long doc/def strings for all words in the object. One long word, created by any user, will mess up `system/words` output, unless we truncate words. In which case we're back to whether that's acceptable.

greggirwin:
I'm not going to add extra padding, based on longest word, to datatype help right now, because it iterates over `system/words` and would have to do a second pass to do it. It's a single pass op right now, emitting as it finds matches.

PR #2501, opened 2017-03-26
Add GUI Console Inspector

maximvl:
A GUI tool to inspect words & values with filters and more features planned.

inspector

### DIFF
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -297,6 +297,7 @@ parse-trace: func [
 		limit [integer!]
 	return: [logic! block!]
 ][
+	clear p-indent
 	either case [
 		parse/case/trace input rules :on-parse-event
 	][
--- a/tests/source/units/all-tests.txt
+++ b/tests/source/units/all-tests.txt
@@ -34,6 +34,7 @@ case-folding-test.red
 throw-test.red
 vector-test.red
 map-test.red
+hash-test.red
 pair-test.red
 date-test.red
 convert-test.red
--- a/tests/source/units/hash-test.red
+++ b/tests/source/units/hash-test.red
@@ -1,0 +1,121 @@
+Red [
+	Title:   "Red hash test script"
+	Author:  "Maxim Velesyuk"
+	File: 	 %hash-test.red
+	Tabs:	   2
+	Rights:  "Copyright (C) 2017 Maxim Velesyuk. All rights reserved."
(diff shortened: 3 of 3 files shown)


maximvl:
Tests took to much time on appveyor, but travis one was fine.

environment/console/inspector.red
```diff
@@ -0,0 +1,87 @@
...
+]
+
+system/console/inspector: context [
+   initial-words: sort keys-of system/words
+   groups: #()
+
+   group: func [words /local acc s t] [
+      acc: copy #()
+      forall words [
+         error? try [
```
dockimbel: What you want to use here is `attempt` instead of `error? try`.

environment/console/gui-console.red
```diff
@@ -103,6 +104,9 @@ gui-console-ctx: context [
 		]
 		apply-cfg
 		win/selected: console
+
+    if system/console/inspector [ system/console/inspector/init ]
+		system/view/auto-sync?: yes
```
dockimbel: You should not "force" the mode globally, but rather set it when the app is about to be displayed, and set the `auto-sync?` value back on exit.

environment/console/inspector.red
```diff
@@ -0,0 +1,87 @@
...
+   make-words: func [types /local words] [
+      words: copy []
+      if types/system [ append words initial-words ]
+      if types/user   [ append words make-diff ]
+      ; TODO add functions, datatypes, constants etc
+      forall words [ words/1: to-string words/1 ]
+      words
+   ]
+
+   show: func [/local spec words sys-words user-words search list desc] [
```
dockimbel: Just a remark, you don't need to change it if you don't want to: [`show`](https://doc.red-lang.org/en/view.html#_show_function) is a key function in the View public API. If you have to use it somewhere in that context in the future, this might bite you in the back.

geekyi:
@maximvl If you're looking for inspiration, there's [anamonitor](http://rebol2.blogspot.com/2011/11/anamonitor-2-check-block-contents.html)

geekyi:
@maximvl maybe your new commits (hash! tests, etc)  should be in a new pull request?

greggirwin:
The GUI console code itself doesn't seem to be included in the commits for this PR. Something may have changed, as requested changes can't find their source. Closing this PR.

@maximvl, if you pick this back up, let's chat on Gitter and maybe start by adding some of our tools (I have some too) in support repos as we flesh them out.

Issue #2513, opened 2017-03-29
`system/console/limit` does not provide correct value in the CLI console mode


Oldes:
The console width in number of chars is correct in the gui-console, but not set in the CLI version.

dockimbel:
We should rename `limit` to something more appropriate. It was meant only for the GUI console, though it is desirable to support CLI console too.

PR #2514, opened 2017-03-29
FEAT: issue #2513 - providing information about console width in the …

Oldes:
…CLI mode

### DIFF
--- a/environment/console/POSIX.reds
+++ b/environment/console/POSIX.reds
@@ -367,6 +367,7 @@ get-window-size: func [
 	/local
 		ws	 [winsize!]
 		here [integer!]
+		size [red-pair!]
 ][
 	ws: declare winsize!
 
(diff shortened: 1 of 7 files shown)


dockimbel:
All platforms should behave the same way, this feature needs to be done for all platforms supported or none.

Oldes:
With this modification, the `system/console/size` is newly updated in CLI version when user changes size from properties menu.

Oldes:
@dockimbel it should be possible to add this feature to other platforms. See http://stackoverflow.com/a/1022961/494472

Oldes:
@qtxie could you please test if the POSIX change from the last commit is correct.. I have just Windows available now so it was a blind change.

dockimbel:
@qtxie Please review the changes in this PR.

Oldes:
There is still one issue... at this moment the CLI version reports size as a size of buffer and not the screen as it is in the GUI version... I think it should be changed as the second is more useful. I will do it soon.

PR #2522, opened 2017-03-31
FEAT: basic support for ANSI escaping sequences in Windows CLI console

Oldes:
This modification enables ANSI escape sequences to be processed in the Windows CLI console. It would be probably fine to add at least cursor support into `gui-console` while stripping the escape sequences there when not applied (gui console cannot have colors).

### DIFF
--- a/runtime/platform/win32-ansi.reds
+++ b/runtime/platform/win32-ansi.reds
@@ -1,0 +1,329 @@
+Red/System [
+	Title:   "ANSI Escape sequences support in Windows CLI console"
+	Author:  "Oldes"
+	File: 	 %win32-ansi.reds
+	Tabs:	 4
+	Rights:  "Copyright (C) 2011-2017 Nenad Rakocevic. All rights reserved."
+	License: {
+		Distributed under the Boost Software License, Version 1.0.
+		See https://github.com/dockimbel/Red/blob/master/red-system/runtime/BSL-License.txt
+	}
+]
+
+#define FOREGROUND_BLACK 		   00h
+#define FOREGROUND_BLUE 		   01h
+#define FOREGROUND_GREEN 	 	   02h
+#define FOREGROUND_CYAN 	 	   03h
+#define FOREGROUND_RED 		 	   04h
+#define FOREGROUND_MAGENTA	 	   05h
+#define FOREGROUND_YELLOW	 	   06h
+#define FOREGROUND_GREY		 	   07h
+#define FOREGROUND_INTENSITY	   08h
+#define FOREGROUND_WHITE	 	   0Fh
+#define BACKGROUND_BLUE			   10h
+#define BACKGROUND_CYAN 	 	   30h
+#define BACKGROUND_GREEN		   20h
+#define BACKGROUND_RED			   40h
+#define BACKGROUND_MAGENTA	 	   50h
+#define BACKGROUND_YELLOW	 	   60h
+#define BACKGROUND_GREY		 	   70h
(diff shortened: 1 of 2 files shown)


Oldes:
With above modification one can get color output like here:
https://github.com/Oldes/media/raw/master/screens/ansi-color-test.PNG

using this code:
```
	print "^[[2J" ;reset screen
	foreach [int des] [
		"00" "none"
		"39" "foreground"
		"49" "background"
		"01" "all"
	] [
		print ["INTENSITY:" des]
		repeat b 8 [
			repeat c 8 [
				pen: 29 + c
				bac: 39 + b
				code: to-string reduce ["[" int ";" pen ";" bac "m"]
				prin to-string reduce [#"^[" code "^^[" code "^[[0m  "	]
			]
			print ""
		]
		print ""
	]
```

dockimbel:
You need to provide same feature on Linux and macOS platform in order to be accepted. Moreover, be aware that such code will be dropped or at least, heavily changed once we pass 0.7.0 and implement the console:// port.

Oldes:
*nix terminals support ansi sequences by default. Just with some differences in colors.

Oldes:
Old Rebol2 console suported at least the cursor movement and erasing.

Oldes:
All console versions should support it at least in a minimal form as it is standard used many years.

dockimbel:
I know ANSI escape codes and how Unix terminals, though, I was not aware that the current Red CLI console on Unix let them pass unchanged (was not the case before).

StephaneVeneri:
With the last Automated builds on Windows, works fine with gui-console but problem with cli-console :
```
C:\red>red-02apr17-7f0f958.exe  --cli
Compiling Red console...
*** Loading Error: file access error: %win32-ansi.reds
*** in file: %/C/red/win32-cli.reds
*** at line: 18
```

Oldes:
Ah.. the new file must be registered to be included in the packer.

Oldes:
Here is a fix: https://github.com/red/red/pull/2534

runtime/platform/win32-ansi.reds
```diff
@@ -0,0 +1,400 @@
...
+			str: str + 8
+			bytes: 8
+		]
+	]
+	state:   1
+	value1:  0
+	value2:  0
+	attribute: -1
+	until [
+		cp: str/1
```
qtxie: Same as FIX-1, should use `string/get-char` to get the codepoint.

runtime/platform/win32-ansi.reds
```diff
@@ -0,0 +1,400 @@
...
+		state   [integer!]
+		value1  [integer!]
+		value2  [integer!]
+		command [integer!]
+		attribute [integer!]
+		cursor  [coord!]
+		col     [integer!]
+		row     [integer!]
+][
+	switch unit [
```
qtxie: Should use `string/get-char` here instead of the `switch` code. See [here](https://github.com/red/red/commit/a706ee89884f9a9db87161fdbdb27e119ab69535#diff-aebd7f72bc8cf92ed72bf130e69b59eeR208).

runtime/platform/win32-ansi.reds
Oldes: @qtxie thanks, sounds reasonable... the fix is in this commit: https://github.com/red/red/pull/2670/commits/9e807029e4732b5dc7e9f073ae35d200982ba085

qtxie: @Oldes Thanks for the fix. 👍

Issue #2539, opened 2017-04-03
Interpreter has problem with /extern function definition if lit-word is used too


Oldes:
This code:
```
buffer: copy ""
test: function [
	'word [any-type!]
	/extern buffer
][]
```
compiles fine, but fails in console (interpreted):
```
*** Script Error: invalid /extern value: 'word
*** Where: function
```

dockimbel:
Good catch!

dockimbel:
Fixed by PR #2540.

PR #2540, opened 2017-04-03
FIX: issue #2539 - lit-word used with /extern function definitions fa…

Oldes:
…ils in the interpreter


Oldes:
Although I'm not sure, if this should be allowed:
```
test: function [/extern buffer 'a :b][a: b: 1 ]
```

dockimbel:
Not that should not be allowed, but that is a different issue.

PR #2549, opened 2017-04-07
FEAT: debug function for displaying size of arguments and call stacks

Oldes:
This function could be useful to trace stack leaks.


dockimbel:
I don't remember ever needing that. Just doing a `probe stack/top - stack/bottom` would suffice (or the equivalent for the call stack).

Anyway, tracking eventual stack leaks requires much more than just dumping the stack size (a big stack size can be justified in some cases where a lot of recursivity happens, like in View).

PR #2608, opened 2017-04-21
Update actions.red

toomasv:
Removed map! from `insert` allowed series types.


dockimbel:
You should be submitting only one commit, instead of 3 (see above). Though, I can squash them into one on merging, so it's not a big issue.

toomasv:
Why the checks don't pass?

dockimbel:
It's not related to your commit, but master branch is broken since a few commits, it will be fixed asap.

PR #2638, opened 2017-04-28
Improvement for converting floats and integers from string

Zamlox:
Let me know if I placed correctly functions in ```common.reds```

Main reason of the change is to get rid of extra memory allocation during conversion.
Now it works also:
```
>> to integer! "15623%"
== 156
```

Error message displayed like:
```
>> to integer! "156#23%"
*** Script Error: cannot MAKE/TO percent! from: "156#23%"
*** Where: to
```

### DIFF
--- a/environment/lexer.red
+++ b/environment/lexer.red
@@ -146,113 +146,36 @@ system/lexer: context [
 		ret
 	]
 
-	make-number: routine [
+	string-number: routine [
 		start  [string!]
-		end	   [string!]
+		len	   [integer!]
 		type   [datatype!]
-		/local
-			str  [series!]
-			c	 [integer!]
-			n	 [integer!]
-			m	 [integer!]
-			len  [integer!]
-			unit [integer!]
-			p	 [byte-ptr!]
-			neg? [logic!]
+		/local 
+			err?	[red-logic!]
+			v 		[red-value!]
 	][
-		if type/value <> TYPE_INTEGER [
-			make-float start end type					;-- float! escape path
-			exit
+		err?: declare red-logic!
+		v: as red-value! red/string-number start len type err?
+		if err?/value [
+			fire [TO_ERROR(script bad-to-arg) datatype/push type/value v]
 		]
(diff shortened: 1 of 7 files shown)


dockimbel:
I was expecting a few lines changes only... but that went too far:
```
14 == to integer! "1423.5%"
```
Those are not the conversion rules defined in docs/conversion-matrix.xls

The rules for converting a string to a number (or any other datatype) are based on lexical rules, they should produce a result equivalent to `load`.

qtxie:
We can use a global block to save the converted value in `load-value`, right?
```red
load-value: func [
    str     [red-string!]
    return: [red-value!]
    /local
        blk   [red-block!]
        value [red-value!]
][
    #call [system/lexer/transcode/one/only str none no] ;@@ here we can pass a blk instead of none
    ...
]

```

Zamlox:
indeed, I didn't specify that my changes are small. I can take back the PR.
Related to matrix conversion, why is following conversion allowed:
```
>> to float! "1234.5%"
== 12.345
```
and not allowed for ```integer!``` type ?

Zamlox:
If the block is local to the function. must be allocated each time, which means memory allocation. If it's global, it must be protected in the future by synchronization object, when multithreading will be implemented.

Zamlox:
I'm closing this PR because the implementation is not as expected.

dockimbel:
@zamlox Good point about the `to float! "1234.5%"` case, that is not good, we need to fix that, it should return an error in such case.

dockimbel:
@qtxie No need for a global block, we just need to not call the lexer (so not call `load-value`), and just directly call `make-number` to do the deserialization.

qtxie:
I'm not sure it will be faster as we need to do the syntax checking twice. (In the lexer and in `make-number`)

Zamlox:
No need to do special syntax checking, especially if conversion from percentage in string format is not allowed. Just rely on the second parameter of ```strtod```, if it's not empty then it's not a valid number (check my changes)

PR #2693, opened 2017-05-17
FIX: Can be compiled on macOS now

rebolek:


dockimbel:
Closing it as it's not the right fix.

PR #2710, opened 2017-05-23
FEAT: removing old Red/System library code as this was moved to separate Red/Code repository

Oldes:


Oldes:
You should also merge the code in Red/code, where these files were moved.

PR #2720, opened 2017-05-28
Add none! to spec type, for recursive calls.

greggirwin:


dockimbel:
Ah, thanks, I removed it first, then forgot to put it back.

PR #2832, opened 2017-06-23
FEAT: REJOIN supports /ONLY for ignoring NONE values

rebolek:


geekyi:
Just an observation from my perspective; 
usually, the convention in Redbol is to optimize for the most common case.
This includes doing any additional things that would otherwise be done as boilerplate.
`/only` is used as an escape hatch when you want the *pure* case.

Here, it feels reversed. 
That is, I'd usually want `none` values ignored, and `/only` `rejoin`ed if I don't want to ignore none values.
Tell me if that feels the opposite of what you'd want.

I suppose @rebolek did it this way partly because this is the more compatible way.
Conceptually tho, this feels the opposite to how `/only` is done usually, for *me* at least

rebolek:
@geekyi you may want to have something like
```
>> x: none print rejoin ["Value of x is " x]
Value of x is none
```

As you wrote, I dit it because it is more compatible. But I think it's up to @dockimbel to decide, which way is better.

greggirwin:
Given the semantics of `/only` for appending blocks that exists already, I would use a different name if we want to do this. That leads to the question of whether `/only` for blocks has value with `rejoin`, which needs a good example to justify. For this behavior, is the change clearer than using `trim`?
```
rejoin/only [a b c d]
rejoin trim [a b c d]
```

rebolek:
@greggirwin you are trimming before reducing, so it won't help. It would need to be `rejoin trim reduce [a b c d]` which is kind of ugly and the block is now reduced twice.

greggirwin:
Ah, right.

rebolek:
@greggirwin Also, `/only` is not use exclusively for appending block, see its usage in current Red:
```
>> foreach word words-of system/words [spec: try [spec-of get word] unless error? spec [if pos: select spec /only [print [mold word #"-" pos]]]]
insert - Insert block types as single values (overrides /part)
set - Block or object value argument is set as a single value
change - Changes a series as a series.
random - Pick a random value from a series
mold - Exclude outer brackets if value is a block
append - Insert block types as single values (overrides /part)
find - Treat a series search value as a single value
select - Treat a series search value as a single value
compose - Compose nested blocks as blocks containing their values
clean-path - Do not prepend current directory
repend - Appends a block value as a block
```

greggirwin:
What I meant was that `rejoin` is appending, and `/only` in the context of `append` has a meaning.

greggirwin:
Maybe we can consider a `rejoin` redesign. There has been talk for years about a `/with` refinement as well.

rebolek:
@greggirwin Here is some code I wrote sometime ago for R3: https://gist.github.com/rebolek/da7f38e0e7a641000ad70a43660f2abc It supports both `/trim` and `/with`.

greggirwin:
Thanks. Do you think it helps to make the result string based on the length of the block? If it's all chars it would of course. :^)

rebolek:
Right, that actually makes no sense :) I've just copied my old code and only checked if it works in Red.
We should probably start a document in wiki to gather ideas and code.

greggirwin:
Experimental funcs from a while back:
```
; TBD: pair tuple
combine: func [
    "Merge values, modifying a if possible"
    a "Modified if series or map"
    b "Single value or block of values; reduced if `a` is not an object or map"
][
    if all [block? :b  not object? :a  not map? :a] [b: reduce b]
    case [
        series? :a [append a :b]
        map?    :a [extend a :b]
        object? :a [make a :b]
        'else      [append form :a :b]
    ]
]
join: func [
    "Concatenate/merge values"
    a "Coerced to string if not a series, map, or object"
    b "Single value or block of values; reduced if `a` is not an object or map"
][
    if all [block? :b  not object? :a  not map? :a] [b: reduce b]
    case [
        series? :a [a: copy a]
        map?    :a [a: copy a]
        object? :a []    ; form or mold?
        'else      [a: form :a]
    ]
    combine a b
]
```

greggirwin:
Crated https://github.com/red/red/wiki/JOIN-REJOIN-COMBINE-MERGE-EXTEND-design-ideas for discussion on this idea. Closing this ticket, but referenced it in the new wiki page.

PR #2836, opened 2017-06-24
Update run-all.r (removed 'ask' to avoid halting tests)

eranws:
I'm running a test suite on a debian docker container:
https://circleci.com/gh/eranws/red-docker/53

Is the `ask` necessary at the end of the tests? it causes the system to halt (for ~10minutes until timeout)


PeterWAWood:
You should run the tests with the ```--batch``` option.

PR #2838, opened 2017-06-25
FEAT: on-drop-files event handler on Windows

Oldes:
With this modification it is possible on Windows drag files from OS and drop them into Red opened windows. Result is an event with block of file names.

See `tests\view-drop-files.red` file for basic usage example.

So far it is only for Windows... other systems will just ignore this event so nothing will be fired there.

### DIFF
--- a/modules/view/backends/platform.red
+++ b/modules/view/backends/platform.red
@@ -172,6 +172,8 @@ system/view/platform: context [
 				EVT_TIME
 				EVT_DRAWING
 				EVT_SCROLL
+
+				EVT_DROP_FILES
 			]
 			
 			#enum event-flag! [
@@ -346,6 +348,7 @@ system/view/platform: context [
 			_time:			word/load "time"
 			_drawing:		word/load "drawing"
 			_scroll:		word/load "scroll"
+			_drop-files:	word/load "drop-files"
 
 			_track:			word/load "track"
 			_page-left:		word/load "page-left"
@@ -425,6 +428,7 @@ system/view/platform: context [
 					EVT_ROTATE		 [_rotate]
 					EVT_TWO_TAP		 [_two-tap]
 					EVT_PRESS_TAP	 [_press-tap]
+					EVT_DROP_FILES	 [_drop-files]
 				]
 			]
 			
@@ -471,6 +475,7 @@ system/view/platform: context [
 					sym = _rotate/symbol		[sym: EVT_ROTATE]
 					sym = _two-tap/symbol		[sym: EVT_TWO_TAP]
 					sym = _press-tap/symbol		[sym: EVT_PRESS_TAP]
+					sym = _drop-files/symbol 	[sym: EVT_DROP_FILES]
 					true [
 						fire [TO_ERROR(script not-event-type) word]
 					]
--- a/modules/view/backends/windows/events.reds
+++ b/modules/view/backends/windows/events.reds
@@ -261,6 +261,12 @@ get-event-picked: func [
 		msg	[tagMSG]
 		gi	[GESTUREINFO]
 		zd	[float!]
+		i   [integer!]
+		num [integer!]
+		sz  [integer!]
+		blk [red-block!]
+		buf [byte-ptr!]
+		bsz [integer!]
 ][
 	as red-value! switch evt/type [
 		EVT_ZOOM
@@ -291,6 +297,29 @@ get-event-picked: func [
 		EVT_SCROLL [
 			msg: as tagMSG evt/msg
 			integer/push get-track-pos msg/hWnd msg/msg = WM_VSCROLL
+		]
+		EVT_DROP_FILES [
+			msg: as tagMSG evt/msg
+			num: DragQueryFile msg/wParam FFFFFFFFh null 0 ;sets number of dropped files
+			blk: block/push* num ;creates block which will hold file names (as strings)
+			bsz: 64           ;initial temp buffer size
+			buf: allocate bsz ;allocates temp buffer used to query dropped file names
+			i: 0
+			while [i < num][
+				sz: 1 + DragQueryFile msg/wParam i null 0 ;returns the required size, in characters
+				if sz > bsz [
+					;reallocating temp buffer to hold result
+					free buf
+					bsz: sz
+					buf: allocate bsz
+				]
+				if 0 <> DragQueryFile msg/wParam i buf bsz [
+					string/load-in as c-string! buf sz - 1 blk UTF-16LE
+				]
+				i: i + 1
+			]
+			free buf
+			blk
 		]
 		default	 [integer/push evt/flags << 16 >> 16]
 	]
@@ -1143,6 +1172,9 @@ WndProc: func [
 				]
 			]
 		]
+		;WM_DROPFILES [
+		;	print-line "DROP FILES!! events.reds WndProc"
+		;]
 		default [0]
 	]
 	if ext-parent-proc? [call-custom-proc hWnd msg wParam lParam]
@@ -1164,6 +1196,10 @@ process: func [
 		evt?   [logic!]
 ][
 	switch msg/msg [
+		WM_DROPFILES [
+			make-event msg 0 EVT_DROP_FILES
+			EVT_DISPATCH
+		]
 		WM_MOUSEMOVE [
 			lParam: msg/lParam
 			x: WIN32_LOWORD(lParam)
--- a/modules/view/backends/windows/events.reds
+++ b/modules/view/backends/windows/events.reds
@@ -306,15 +306,15 @@ get-event-picked: func [
 			buf: allocate bsz ;allocates temp buffer used to query dropped file names
 			i: 0
 			while [i < num][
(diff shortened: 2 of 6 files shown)


greggirwin:
This would be a really nice feature on Windows. PR needs an update, because of other new events, but the main question is whether @dockimbel wants to include it, as it's Windows-only. I think it's valuable enough to include.

endo64:
It would also be nice to have `on-drop-file` event for specific widgets instead of whole window.

greggirwin:
In the grand scheme, we need to think about the DnD system, where `drag-start` can set info for the drag object, which can automatically be checked in the `drag` event, and easily map to face cursor changes, to indicate droppability. This then leads to generalized eventing/reactive handling, for which a number of us have experiments I believe.

greggirwin:
If we can find someone to champion this on MacOS, we can eliminate the platform specific issue.

greggirwin:
@qtxie @dockimbel, this has conflicts now, but I'd like to see it resolved one way or another.

greggirwin:
If we don't want to do it this way, but full I/O brings us something like R2's `system` port, people can add the functionality that way.

greggirwin:
Since we don't run on MacOS right now, let's get this in place. :^) @qtxie @dockimbel ?

greggirwin:
Ping. Ping. Ping. Ping. ...

It's silly to me that we still don't have this. Let's do it, and close the oldest PR in Red.

greggirwin:
I agree that having it for specific widgets would be great, maybe even necessary, but putting the basics in place is a start.

dockimbel:
So we drop GTK support and we go Windows-only for View, right? Otherwise, are you or people supporting this PR, willing to implement the GTK version, so that the burden is not put on the Core team?

GiuseppeChillemi:
Let it be Windows only Doc just this features if possible. Personally, I have my task bar based on Red blocked because this feature lacks. Then someone will port it to GTK.

dockimbel:
@GiuseppeChillemi Just merge that PR locally in your codebase and you'll get that feature for your specific needs. So there's no show-stopper there for you.

> Then someone will port it to GTK.

Sure...😆 Let's just build whatever we need personally, then let the rest of the world fill in the holes *eventually*... who cares for other users needs, as long as *my* needs are fulfilled, right?

GiuseppeChillemi:
No Doc, I am not alone. Gregg needs it, Oldes needs it, I need it, Hiiamboris needs it and of course different others. If personal I would not have asked but on Matrix this needing has appeared too, together with other most wanted: cursor management in areas and tables (The latter seems an user is trying to implement it).

dockimbel:
Sure, let's drop GTK support as you and a few others can't care less about it.

hiiamboris:
GTK is [lagging behind on a lot of features](https://github.com/red/red/issues?q=is%3Aissue+is%3Aopen+gtk), where a rarely used event type will unlikely make much difference. 

Some missing move/resize events, numerous serious glitches in keyboard events, size-text not working on multiline input, no shadow support in Draw, easily congested event loop, no DPI info. That's just off the top of my head.

So I don't think that GTK parity is a strong argument here :)

dockimbel:
> So I don't think that GTK parity is a strong argument here :)

So, that's supposed to be a good argument to make things worse?

If I accept this PR, nobody will ever implement it for GTK or macOS. That will fall on the Core team to do so and to implement to numerous additions related to that feature. People will keep opening tickets labelled as "bugs" because it does not work on GTK or does not have the same features... 

So if you want this PR merge, please add GTK support and provide maintenance support for further fixing issues and implementing new features related to that code. After 8 years, nobody cared enough to do so. That is why I'm asking if we should just drop GTK support. You seem to agree with that too, as you mentioned many minor behavior differences with Windows backend.

dockimbel:
> Also please consider that I understand the team has its schedules but this (and some other little features) have been waited for about 8 (eight!) years. I would like to create some products but I am stopped by the lack of them.

As I mentioned now several times above, you could have merged it locally *any time* in your local repo, so there's *no show-stopper* and never was!

>  [...] have been waited for about 8 (eight!) years [...]

That is just you not understanding how open source and github works...

dockimbel:
Here is my proposition: I agree to merge this PR if we drop the GTK backend for Red/View. Deal?

Oldes:
I'm unable to extend or update this PR, so it's better to close it.

GiuseppeChillemi:
This is the worst scenario. 

Personally I don't understand why a single actor would need GTK backend dropping. It is a kind of extortion to tell us "I will implement it you agree to loose this other". Will you pardon me if I use the right words.

 Open source works that we contribute when we have the skill to help. I don't have the one needed for some changes so we rely on you. If I would have been able to make all the changes needed by the community, I would have implemented them without asking. 

 I am here to receive sometime some sweet and see common basic neededings inplemented but I admit I am living just on my hope, and it is a long time I am stuck hoping for the basic things I never receive. Too long.

However, it has been closed, as I have said, with the worst outcome for the community.

🤷🏻

greggirwin:
<sigh> I certainly didn't want to start a fight about this.

Personally, while I don't use GTK, I like the idea of it being there. I hate to throw away all the stuff that does work, and when people run up against things, that's what issue reports are for.  That tells us who is using what, what they want and need.

@dockimbel is there any chance full I/O could support something like `system port/winmsg` like R2 did? It could even be more declarative, so the wparams are part of a hook call spec, which a callback func vets if there are multiples. On second thought, could it be done similar to `insert-event-func`, but for (potentially) unmapped OS events?

http://www.rebol.org/view-script.r?script=sys-port-drag-accept.r&sid=zh8tp

I understand that deeper hooks like this can let users mess things up, and conflict with each other, but, along with FFI, it also puts more power in their hands. I also know high level FFI needs `struct` work, but I believe putting that in users hands is a big win as well. It's an affordance over using R/S and is a nice feature to drive adoption IMO.

GiuseppeChillemi:
Gregg I don't want to start a fight too but this is the wrong way of telling the community "I don't have time for it, sorry".

A community will remain loyal to you if, from time to time, you spend a little of your time to help and solve their problem. It is called CARE. With a little CARE even in hard times people be on your side. 

This evening I have reached my limit. I am stuck, I have waited silently because I know it is not easy. But you can't answer me in this way. It is not just for the actors but for table and cursor in area too. I repeat: **I AM TOTALLY STUCK**. So you don't take care of me? I will  take care of myself and move to another language. Everything is faster than 5+ years for basic tools, even assembly.

 I LEAVE, goodbye.

hiiamboris:
*[sound of a door slam]*

dockimbel:
@Oldes The conflicts with current master branch in this PR are easy to solve, so it can be resurrected if needed.

@greggirwin Certainly, a `system:` port would be welcome. As it depends on the GUI sub-system only, it should be possible to implement it right now, as `port!` are there since a while. Though, it would need some design to determine the best way to interface it with the GUI event loop, and also upcoming /Library component (IO branch) to make such port really useful.

@GiuseppeChillemi I can understand your frustration, but in the context of an open source project, that's misplaced. Nobody owes anything to anyone in such context. As a user, you don't pay anything to access, use the software and get bugs fixed. We are not here to serve you or anyone else specifically. We have goals and priorities. You can request some features and we (the maintainers) decide if and when to implement them according to many factors (implementation complexity, relevancy against our goals, priority level, general usefulness, etc...). As it's an open source project, you are free to implement yourself the feature you want or ask someone else to do it for you (even paying someone if needed). PRs can be sent to the main repo and we (the maintainers) decide if we accept it or not. Given the decentralized aspect of git/github, you or anyone else can pick up any PR or any code anywhere and merge it locally to fulfill one's personal needs. You have total freedom, so do we. 

Concerning your personal complains about wished features:

- `drop-files` event: the PR is here since years, so you could have used it since day one if you needed so, nothing was stoppping you at all! It was not merged in master because it lacks support for our other backends. You personally don't care about other platforms, we do and other users do too. Arbitrary implementing features on one platform only goes against our goals, so we try to avoid that as much as possible. You can disagree with our goals but you can't blame me for working towards them. 

- `table` widget: probably the most complex widget you can think of in GUI framework. It's not easy nor quick to design and implement, it's a huge work! Moreover, unfortunately, the native backends on major platforms have a different set of features with only some core ones overlapping (see my early [screening](https://github.com/red/REP/issues/159) and prototype). It could be possible to implement a minimal one with support on all our backends, but it would still take some time (weeks?) and we would keep hearing complaints and requests for a full-featured one all the time... At this point of the project, that's not a priority.

- ability to control the cursor programmatically in `area` widget: the native `area` is not really designed for advanced editing as you want to implement (especially on Windows). You should have used a `rich-text` widget for that or build a Draw-based one where you would have total control on the look and behavior. We told you so before. Moreover, even if we would implement that on Windows, I'm not sure the other backends would support that (that needs checking). In the end, that's not something that can be quickly implemented, nor is it a priority (as there are better workarounds as explained above).

Last but not least, you can always try to propose bounties to community members for implementing features you would like to have (like the ones above) without waiting for us. I'm sure that someone would pick it up at some point. That's a common practice in many open source projects.

PR #2874, opened 2017-07-10
FEAT: improved area widget performance for strings with many lines

Oldes:
Instead of calling slow `DrawText` call with `DT_CALCRECT` it measures size for each line of given text using `GetTabbedTextExtent` and decides if scroll bars are needed or not. 

There is also a little optimization, which avoids double conversion into `utf16` string, which is needed for Windows API calls.


qtxie:
Nice improve, thanks. :+1:

Issue #2921, opened 2017-07-23
Missing .gitignore?


southwolf:
After running `rebol ./run-all.r`, I got these files

```
crush.dylib
quick-test/quick-test.log
quick-test/runnable/
system/tests/source/units/auto-tests/
tests/source/units/auto-tests/
```

It will be more clear if we have a `.gitignore` file to ignore such files, am I right?

dockimbel:
Usually, it is used only locally in your working folder, but it depends on the kind of git client you use. I personally don't use a `.gitignore` file at all (TortoiseGit client here) and I don't think most people in the Red team use one. Adding one to the red/red repo would create problems, as most people will customize it, and submit a custom version of that file along with their PR (we had such cases in the past), so I definitely think it's more convenient to keep it locally only.

Eventually, you could submit your `.gitignore` file for Red to https://github.com/github/gitignore so that we can put a link to it in our README for contributors to use it as a template if they need one.

PeterWAWood:
I do have a `.gitignore` file, among many other things it does include:
```
quick-test/quick-test.log
quick-test/runnable/
system/tests/source/units/auto-tests/
tests/source/units/auto-tests/
crush.dll
```

Specifically, it also includes `.gitignore` itself to avoid the problem of it being over-written by a `.gitignore` file in the repository.

Having a sample `.gitignore` file seems a good idea.

dockimbel:
@PeterWAWood Could you (or @9214, @greggirwin) please post a gitignore file to https://github.com/github/gitignore for Red, and reference it in the README?

9214:
@dockimbel the one in @southwolf's PR? How README file should be updated, specifically?

PeterWAWood:
@dockimbel I will pd this later today or tomorrow. 

I will need to change some of my work processes to account for a .gitignore file in the repo. The first entry in my .gitignore file is .gitignore :-)

@9214 The .gitignore file in this PR is not sufficient as it is for macOS only.

PeterWAWood:
@dockimbel I misread your request and prepared a .gitignore file for the red repo :-(

git does provide a way to have a repo wide .gitignore and to ignore other files locally (by using .git/info/exclude).

I got as far as committing the change and have now reverted it.

I will submit a PR to the GitHub/gitignore repo instead :-)

PeterWAWood:
I've submitted a PR to the GitHub/gitignore repo and made a note in the red repo README.

PeterWAWood:
[...]

PR #2922, opened 2017-07-23
Add .gitignore (#2921)

southwolf:
Added a `.gitignore` file to ignore file generated by tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,7 @@
+crush.dylib
+crush.dll
+
+quick-test/quick-test.log
+quick-test/runnable/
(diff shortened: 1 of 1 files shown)


greggirwin:
@PeterWAWood, was this handled by another PR, or is there a reason we *wouldn't* want to merge this?

PeterWAWood:
The Red Team's working processes would be affected if there was a .gitignore file in the repository. 

I have added https://github.com/red/red/blob/master/.gitignore-sample to the top level directory in the red repository. It can be renamed by anyone wanting to use it.

I will close this PR.

PR #2952, opened 2017-08-01
Gtk

rcqls:
Sorry, I realized that set-text did not set face/text properly


modules/view/backends/gtk3/handlers.reds
```diff
@@ -46,12 +46,13 @@ set-text: func [
...
 		str: as red-string! get-node-facet ctx FACE_OBJ_TEXT
 		if TYPE_OF(str) <> TYPE_STRING [
-			str/node: unicode/load-utf8 text size
+			string/make-at as red-value! str size UCS-2
 		]
 		if size = 0 [
 			string/rs-reset str
 			exit
 		]
+		str/node: unicode/load-utf8 text size
```
qtxie: We should reuse the string buffer instead of create one each time. You can use `load-utf8-buffer`.

Issue #2898, opened 2017-07-18
`source prompt` throws an error


greggirwin:
```
>> source prompt
*** Script Error: val needs a value
*** Where: set
*** Stack: source  
```

geekyi:
I see that `prompt` is unset in my 31 August console.
Did you mean `source any-unset-value` throws an error?

geekyi:
```red
>> source source
source: func [
    "Print the source of a function" 
    'word [any-word! any-path!] "The name of the function" 
    /local val
][
    val: get/any word   ; should check for unset here
...
```

greggirwin:
https://github.com/red/red/pull/2955

PR #2955, opened 2017-08-01
FIX: #2898 `source prompt` throws an error

greggirwin:


modules/view/view.red
```diff
@@ -854,7 +854,8 @@ dump-face: function [
 ][
 	depth: ""
 	print [
-		depth "Style:" face/type "Offset:" face/offset "Size:" face/size
+		depth "Type:" face/type "Style:" if face/options [face/options/style]
```
geekyi: Looks like something else got fixed too :p

greggirwin:
Yeah, the way github combines commits into open PRs makes it a bit of a pain sometimes, when I'd like to keep things separate.

geekyi:
@greggirwin you'd have to make a new branch (I'm guessing)

greggirwin:
Yes, which I will do if I have to, but is more work. I do try to keep from Making Doc's like miserable with my PRs, and have gotten better at that. :^)

PR #2960, opened 2017-08-03
GTK: first attempt for group-box

rcqls:
* radio added
* field and area initialized with buffer
* group-box is in early stage and has to be improved


modules/view/backends/gtk3/gui.reds
```diff
@@ -27,6 +27,9 @@ exit-loop:		0
 red-face-id:	0
 gtk-style-id:	0
 
+none-type: 		as red-word! symbol/make "none"
```
qtxie: It's wrong here. `symbol/make` will return the symbol ID, not a red-word!

modules/view/backends/gtk3/gui.reds
```diff
@@ -100,6 +103,25 @@ get-face-handle: func [
...
 
+parent-type?: func [
+	parent	[handle!]
+	return:	[red-word!]
+	/local
+		qdata	[handle!]
+		p-face	[red-object!]
+		p-type	[red-word!]
+][
+	p-type: none-type
```
qtxie: We can just initialize it to null: `p-type: null` here.

qtxie: Even better, we would resolve the type in the function and returns the symbol id. So we change the return type to `integer!`, then we intialize `p-type` to `-1`.

modules/view/backends/gtk3/gui.reds
```diff
@@ -100,6 +103,25 @@ get-face-handle: func [
...
+	qdata: g_object_get_qdata parent red-face-id
+	if qdata <> as handle! 0 [
+		p-face: as red-object! qdata
+		if TYPE_OF(p-face) = TYPE_OBJECT [
+			p-type: as red-word! get-node-facet p-face/ctx FACE_OBJ_TYPE
```
qtxie: We can resolve the type here:
```
w: as red-word! get-node-facet p-face/ctx FACE_OBJ_TYPE
p-type: symbol/resolve w/symbol
```

rcqls: Thanks! I miss that at the end symbol/resolve returns an integer!. Btw, I changed the name of the function since in fact it is more related to widget than parent in widget-face-symbol?

rcqls:
I also added some preliminary code for panel and very little for tab-panel.

rcqls:
added text-list. widget2 is a container introduced to add the main widget in order to add it to the container parent. Did not find a better name... will see later.

rcqls:
face/selected and face/text work for actor's text-list in a similar way of combobox.

rcqls:
Tab-panel added but without signal connection yet...

modules/view/backends/gtk3/gui.reds
```diff
@@ -27,6 +27,13 @@ exit-loop:		0
 red-face-id:	0
 gtk-style-id:	0
 
+group-radio:	as handle! 0
+tabs: context [
+	nb: 	0
+	cur: 	0
+	data:	as red-block! 0
```
qtxie: We should get the `data` facet directly from the face!. The `data` may point to another value once we have GC.

modules/view/backends/gtk3/gui.reds
```diff
@@ -529,15 +635,44 @@ OS-make-view: func [
...
-		gtk_fixed_put as handle! container/value widget offset/x offset/y
+		p-sym: get-widget-face-symbol as handle! parent
+		if widget2 = as handle! 0 [widget2: widget]
+		case [
+			p-sym = panel [
+				container: as handle! parent
+				gtk_widget_set_size_request widget2 size/x size/y
+				gtk_fixed_put as handle! container widget2 offset/x offset/y
+			]
+			p-sym = tab-panel [
```
qtxie: It would be better to put those codes for setup tab-panel into a function.

modules/view/view.red
```diff
@@ -745,7 +745,7 @@ show: function [
 			]
 			
 			switch face/type [
-				#if config/OS = 'Windows [				;@@ remove this system specific code
+				#if any [config/OS = 'Windows  config/OS = 'Linux] [				;@@ remove this system specific code
```
rcqls: Ok I checked that this can be avoided. I thought it was necessary because in any other backend (windows, macOS) `link-tabs-toi-parent` is called at least once. I interpreted that `panels` in `face/pane` with `face` being a `tab-panel` do not have `tab-panel` as its natural parent. But it seems this is done without this code....

modules/view/backends/gtk3/gui.reds
rcqls: In order to avoid, there is a more elegant solution in my opinion:  copy each label of each tab  directly in `face/text` of each `panel` created automatically. Like this, each `panel` would know  the label to the tab when the panel is created. This can be done in `view.red`. But for now, I need this trick to have the label tab properly set.
By the way, how a global variable could be cleaned by GC.  This does not make sense for me.

modules/view/backends/gtk3/gui.reds
```diff
@@ -102,7 +107,7 @@ get-face-handle: func [
 	as handle! int/value
 ]
 
-widget-face-symbol?: func [
+get-widget-face-symbol: func [
```
rcqls: I would do... but it seems that @dockimbel uses '?' at the end of a function when returning a boolean in red/system. So I tried to be homogeneous and `get-face-type` is maybe a better name if this last convention is true.

modules/view/backends/gtk3/gui.reds
```diff
@@ -529,15 +635,44 @@ OS-make-view: func [
...
 	if all [
 		sym <> window
 		parent <> 0
 	][
-		gtk_widget_set_size_request widget size/x size/y
-		container: gtk_container_get_children as handle! parent
-		gtk_fixed_put as handle! container/value widget offset/x offset/y
+		p-sym: get-widget-face-symbol as handle! parent
+		if widget2 = as handle! 0 [widget2: widget]
+		case [
```
rcqls: Maybe later when all the widgets would be created. I prefer first to wait. Refactoring this code later for elegancy would be easy. But I think it is too early for now as the code works for tab-panel and all other widgets already introduced.
All the stuff I do is for creation of widgets. Updating these widgets is to do next... So all this work is in not final and could be refactored later.

modules/view/backends/gtk3/gui.reds
rcqls: I would do later.... I do not think it is urgent since there is better solution as mentioned above.
Is it possible for you to test the code when the PR is not accepted?

modules/view/backends/gtk3/gui.reds
rcqls: BTW, I need the parent face pointed by tabs/data. Then I hope that  GC would not clean an existing parent.

qtxie: It's not cleaned by the GC. The GC maybe move the underly memory the pointer point to. Also we can get the `data` facet of the parent face with the same way we get the `type`.

modules/view/backends/gtk3/gui.reds
qtxie: Use `?` when returning a boolean, yes, that's one use case for `?`. Another idiomatic use for `?` is query some information. e.g. `rs-length?`.

modules/view/backends/gtk3/gui.reds
qtxie: Yes, of course. Just checkout your branch and test it, easy. :-)

modules/view/backends/gtk3/gui.reds
rcqls: I had this in mid too and I did it first like this but at some point I thought data was already accessible when tab-panel is created. But, I'll do what you propose.

modules/view/backends/gtk3/gui.reds
rcqls: Of course, I am silly...

rcqls:
I think this version follows mostly your comments and add actors for tab-panel too!

qtxie:
@rcqls Yes, you can.

rcqls:
BTW, view-test.red is currently not modified yet...

qtxie:
When test it on my ubuntu 15.10 with GTK 3.16.7, I got this error:
```
(console:2044): Gtk-CRITICAL **: gtk_entry_buffer_set_text: assertion 'chars != NULL' failed
```

1. open the console
2. do %tests/gtk3/vid.red
3. close it.
4. error shows in the console.

rcqls:
Currently, I have the same messages with `gtk 3.18.9`. Needs to fix that later.

rcqls:
After a long googling, I finally found a solution for request-file and request-dir used inside the red-console. Don't know if what I did to convert the c-string! to red-string! is the best way but it works. I miss maybe to append / at the end of the path for request-dir. But it is a first useable version.

rcqls:
After long reading of the gtk3 doc and the red source code, I realized that font design required to have font as a handle at least to update the font (cf. view.red). The best I imagined to fulfill this requirement was to save the css string as the font handle...
Good news, since I worked also a bit on change-* functions, `react-test.red` is working!

qtxie:
@rcqls I think we should use `Pango` for font handling, CSS is an layer build on top of those APIs, we should avoid it as much as possible.

rcqls:
@qtxie I wanted first to use pango only but after a really long reading of documentation along the different versions, a lot of functions (for example, `gtk_widget_modify_font`,...) disappear in recent versions or are commented as obsolete...  Just read the latest documentation [Text Widget Overview](https://developer.gnome.org/gtk3/stable/TextWidget.html) everything is made using css and it was not in older version of gtk...
Anyway, the goal is first to have a working version of the backend gtk3 on linux... If you are right about pango, it would be possible to change the `font.reds` in such a way but I think that the core team of gtk encourage the use of css which is pretty flexible by the way. I guess red would want one day to have editor with composite elements (different styled elements in a textview) and css is clearly the solution as it is in the html world...I guess this is what the core team of gtk has in mind...

qtxie:
@rcqls OK, I agree with you. Seems it's the right way for GTK3 backend.

rcqls:
This could be change later but I changed hFont to font in the get-text-size function used in size-text only for Linux. The reason is that font structure is very similar to pango_font_description which only concerns font and not css-style which is much more complete.
I also connect area for changed event and it was not as easy as I expected because of the GtkTextIter structure to get the full text of a gtk_text_buffer.
Good news, react-test6.red is working ...

modules/view/view.red
```diff
@@ -643,6 +643,9 @@ system/view: context [
 	auto-sync?: yes										;-- refresh faces on changes automatically
 	debug?: 	no										;-- output verbose logs
 	silent?:	no										;-- do not report errors (livecoding)
+	#if config/OS = 'Linux [
```
dockimbel: This is not good, and won't be accepted. The View engine code should not expose to end users a different API depending on the backend.

rcqls:
This last improvement allow us to have tests pretty smooth considering that not all the info are taking into account. The adjustment 2nd pass try do adjust the offsets of (mostly in the horizontal direction) the widgets just after a widget with allocated size transformed by the gtk system automatically.

qtxie:
The tests looks nice now. 👍 
Does it something like `DeferWindowPos` (win32 api) in GTK ? The purpose of this api is moving or resizing all the widgets once. We may have hundreds of widgets in a window, if we move it one by one, we may see flicker.

rcqls:
You mean that the different gtk_fixed_move calls have to be applied at the same time. Right ? We'll see when developing resizing part.
I tested the look of code inside the repo red/code and it is not so bad. I did not guess that this little trick could fix the look of most of the examples.

rcqls:
With this commit, eve-clock.red is working on linux... except a very small bug to fix later when leaving the application. Also, working on the calculator.red I did not manage yet to draw text on a GtkDrawingArea... (after following a lot of tuto to do so)

rcqls:
Just remember that I started working without any documentation (except of course the official one of red/system that I read very often). So I mimicked some existing code. If `as handle! 0` is equivalent to null, why GTKApp was initialized with `as handle! 0`. I thought it was different because of this. With my knowledge of C, I understand what you mean but at some point there `handle!` (don't think it is in the red/system documentation) and also `pointer!` (in the documentation).

rcqls:
`GtkTextIter!` is in `gtk.reds`. 
For `EVT_KEY_UP`, I understand now why it is not in View documentation related to `area` object. Is it not better in the `livecode.red` to react to `on-change` then which is the default actor?

qtxie:
I got this crash when running the vid.red.
```
qtxie@ubuntu:~/Red/red$ ./console tests/gtk3/vid.red 

(console:6285): Gtk-WARNING **: Theme parsing error: <data>:1:24: 'text-decoration-line' is not a valid property name

*** Runtime Error 98: assertion failed
*** in file: /D/EE/QW/red/modules/view/backends/gtk3/gui.reds
*** at line: 146
***
***   stack: gui/get-face-handle 00000062h
***   stack: gui/get-face-handle B4B420CCh
***   stack: gui/adjust-sizes 093098F8h
***   stack: gui/OS-show-window 154179832
***   stack: gui/OS-refresh-window 154179832
***   stack: ctx344~refresh-window B4D0DA14h
***   stack: show
***   stack: show
***   stack: view
***   stack: red/_function/call B4D0D6F4h B777688Ch
***   stack: red/interpreter/eval-code B4D0D6F4h B4AEF3C8h B4AEF3C8h false 00000000h 00000000h B4DA8E04h
***   stack: red/interpreter/eval-expression B4AEF3B8h B4AEF3C8h false false false
***   stack: red/interpreter/eval B4D0D6D4h true
***   stack: ctx436~try-do B4C66C2Ch
***   stack: ctx436~launch B4C66C2Ch
***   stack: ***_start
```
BTW, I compiled the console in debug mode.

rcqls:
Thanks, I can confirm this regression... I have to check what is the change in charge of this regression.

rcqls:
@qtxie The issue you discovered is because of the introduction of the last stuff about writing text (coming from FACE_OBJ_TEXT) for the base widget. I have to debug it... Thanks again!

rcqls:
`tests/vid.red` and `calculator.red` are both working without any error. The case of none value for font is now managed inside font-description function since it is used in render-text and get-text-size.

dockimbel:
@rcqls 
>  If as handle! 0 is equivalent to null, why GTKApp was initialized with as handle! 0. I thought it was different because of this.

You have to provide a type when declaring a global variable, `null` is untyped, that is why you need `as handle! 0` in order to create a global variable of type `handle!`, and initialize it with `null`.

> With my knowledge of C, I understand what you mean but at some point there is handle! (don't think it is in the red/system documentation) and also pointer! (in the documentation).

`handle!` is not a Red/System type, it's a macro defined at runtime.

rcqls:
@dockimbel Thanks for this clarification!

rcqls:
This is a preliminary work on image allowing`tests/gtk3/view-test.red` to show button icon, image drawing and image in some base face.

rcqls:
drag and drop is working on `tests/shield.red`, `worm.red`, `tests/gtk3/view-test.red`... However, code needs to be cleaned...

PR #3001, opened 2017-08-27
FEAT: [R/S] preliminary support for [bytes! <size>] definitions in struct

Oldes:
When working with system and external binding, you very quickly notice, that many structs use byte arrays... for example Windows Socket data:
```
typedef struct WSAData {
  WORD           wVersion;
  WORD           wHighVersion;
  char           szDescription[WSADESCRIPTION_LEN+1];
  char           szSystemStatus[WSASYS_STATUS_LEN+1];
  unsigned short iMaxSockets;
  unsigned short iMaxUdpDg;
  char FAR       *lpVendorInfo;
} WSADATA, *LPWSADATA;
```
Or `libmpg123`'s id3 tag:
```
typedef struct
{
	char tag[3];         /**< Always the string "TAG", the classic intro. */
	char title[30];      /**< Title string.  */
	char artist[30];     /**< Artist string. */
	char album[30];      /**< Album string. */
	char year[4];        /**< Year string. */
	char comment[30];    /**< Comment string. */
	unsigned char genre; /**< Genre index. */
} mpg123_id3v1;
```

To use these structs from current Red/System is possible, but not easy, so I tried to extend Red/System as you can see in this PR, so for the second example I could use:
```
mpg123_id3v1!: alias struct! [
	tag     [bytes!  3] ; Always the string "TAG", the classic intro.
	title   [bytes! 30] ; Title string. 
	artist  [bytes! 30] ; Artist string.
	album   [bytes! 30] ; Album string.
	year    [bytes!  4] ; Year string.
	comment [bytes! 30] ; Comment string.
	genre   [byte!]     ; Genre index.
]
```

### DIFF
--- a/system/compiler.r
+++ b/system/compiler.r
@@ -184,6 +184,7 @@ system-dialect: make-profilable context [
 			| 'c-string!
 			| 'pointer! into [pointer-syntax]
 			| 'struct!  into [struct-syntax] opt 'value
+			| 'bytes! integer!
 		]
 
 		type-spec: [
@@ -604,6 +605,7 @@ system-dialect: make-profilable context [
 					enum-type? name
 					type: [integer!]
 				]
+				name <> 'bytes!
 				not type: find-aliased name
 				any [
 					all [silent return none]
@@ -621,6 +623,7 @@ system-dialect: make-profilable context [
 				local?: all [locals select locals name]
 				select-globals name
 			]
+			if all [block? type type/1 = 'bytes!][ return type ]
 			if all [not type pos: select functions decorate-fun name][
 				if mark: find pos: pos/4 /local [
 					pos: copy/part pos mark			;-- remove locals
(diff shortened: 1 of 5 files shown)


Oldes:
The mentioned problem is there when I pass the path value as an argument into `typed` function (in above case `print-line`)... so I now dig into compiler if I can find the source of the problem myself.

Oldes:
@dockimbel Now the test script works as expected... let me know if you want me to make proper tests and possibly arm version.

greggirwin:
@PeterWAWood, and thoughts on this PR? I know @dockimbel and @qtxie need to weigh in as well, but it looks like a good idea to me, and includes tests. From the approach taken, it shouldn't affect old code that uses existing syntax.

greggirwin:
It also doesn't preclude further extension, from a brief evaluation.

PR #3021, opened 2017-09-08
FEAT: including fmod import in libc

Oldes:
`fmod` function was maybe not needed in Red runtime, but as `libc` is used also for Red/System, could you please include it? Btw... there is more math related functions missing.. I can add them if you would accept it.

### DIFF
--- a/system/runtime/libc.reds
+++ b/system/runtime/libc.reds
@@ -132,6 +132,11 @@ Red/System [
 			value		[float!]
 			return:		[float!]
 		]
+		fmod:		"fmod" [
+			x           [float!]
+			y           [float!]
+			return:     [float!]
+		]
 	]
 ]
 


greggirwin:
@Oldes, do you have a list handy of all you'd want to add? @dockimbel  or @qtxie will have to approve, but it would make sense to do them all at once, if the size addition isn't an issue. 

On one hand, it makes sense to only include what we need. On the other, if the cost is small, it doesn't open more security risks or surface area for attacks, and may benefit more people using R/S, it's worth considering.

greggirwin:
@Oldes, any thoughts on my 30-Apr comment?

greggirwin:
@Oldes @dockimbel @qtxie, what do we want to do with this PR?

dockimbel:
As we are switching to `fmod()` for the floating point remainder in R/S, I am accepting this PR.

Issue #2969, opened 2017-08-08
Doc-string wrongly placed in `what-dir` function


dockimbel:
https://github.com/red/red/blob/master/environment/functions.red#L572

PR #3036, opened 2017-09-15
FIX: issue #2969 (Doc-string wrongly placed in `what-dir` function)

9214:
Small step for the Red, but huge crawl for a toddler. :baby_bottle:


dockimbel:
Thanks!

PR #3082, opened 2017-10-18
Environment functions test

mahengyang:
unit test for environment/functions.red, some function can not test by red, marked by `TODO`

### DIFF
--- a/quick-test/quick-test.r
+++ b/quick-test/quick-test.r
@@ -199,7 +199,6 @@ qt: make object! [
     ;; red/system or red
     red?: false
     parse read src red?-rule
- 
     ;; compose and write compilation script
     either binary-compiler? [
     	if #"/" <> first src [src: tests-dir/:src]     ;; relative path supplied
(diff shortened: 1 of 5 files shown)


PeterWAWood:
@mahengyang Thanks for the tests. Sorry for the delay. I've now had time to look at run-all.r and updated it to accept tests outside of the tests/source/units/dir.

1. I've updated the "run-all" scripts. My changes are a little more general than the ones you suggested.

2. I've added  ../environment/scalars-test.red to all-tests.txt as an example (which I used to test the changes). It doesn't really add much value apart from that.

3. The list-dir function has been moved out of `functions.red` into one of the console files as it is designed to run only in the console.

You should now be able to submit a PR with just the functions tests and an update to `all-tests.txt'

PeterWAWood:
@mahengyang I've didn't accept the PR because of the conflicts. These were due to the changes to the test runners conflicting with the changes that I made.

I have committed your functions-test.red file in https://github.com/red/red/commit/bba78a52f56db9894040c8db56dde510297d9346.

Also, I commented out the probe test as we need to find away to test it without it's output leaking into the test output.

Issue #3117, opened 2017-11-25
path-thru doesn't sanitise illegal characters


pijoter:
### Expected behavior
path-thru hould sanitise characters illegal from filesystem point of view (remove them or change); characters like "?" or ":" are illegal for most of filesystems but allowed (required) in most of url adresses

### Actual behavior
>> path-thru http://rowery.olsztyn.pl/?r=10
== %/C/ProgramData/Red/cache/rowery.olsztyn.pl/?r=10

### Steps to reproduce the problem
>> read-thru/binary http://rowery.olsztyn.pl/?r=10
*** Access Error: cannot open: %/C/ProgramData/Red/cache/rowery.olsztyn.pl/?r=10
*** Where: write
*** Stack: read-thru  

### Red version and build date, operating system with version.
>> system/version
== 0.6.3
>> system/build/date
== 12-Aug-2017/17:41:41+02:00

dockimbel:
Good point.

dockimbel:
Some sanitization code added to `path-thru`:
```
>> path-thru http://rowery.olsztyn.pl/?r=10
== %/C/ProgramData/Red/cache/rowery.olsztyn.pl/
>> path-thru http://rowery.olsztyn.pl/#hashtag
== %/C/ProgramData/Red/cache/rowery.olsztyn.pl/
```

pijoter:
With these changes there will be the same cache-file for many urls! 
Maybe better approach is to store cache-files under md5 checksum name (like here https://github.com/red/red/pull/3124)?

Oldes:
I have same opinion as @pijoter ... the part after question mark must be used as a part of the url.

hiiamboris:
In continuation of this issue: :point_up: [July 5, 2018 8:42 PM](https://gitter.im/red/bugs?at=5b3e586a3572e970c179111c)

`load-thru https://picsum.photos/100/100/?random` doesn't do the caching, because it ends in a slash and tries to write into a directory: 
```
>> path-thru https://picsum.photos/100/100/?random
== %/C/Users/Gregg/AppData/Roaming/Red/cache/picsum.photos/100/100/
```

Hopefully it will be solved by #3124

hiiamboris:
This one should be closed, as the PR implementing it was merged.

greggirwin:
I'll close this, but need to revisit the chat on the PR, which was quite extensive.

PR #3124, opened 2017-11-30
path-thru filename as md5 checksum

pijoter:
no need to convert from url! to thru-cache dir structure
cache file is placed under path based on md5 checksum

>> path-thru http://www.red-lang.org/
== %/C/ProgramData/Red/cache/C3/C36ABFBA6FA8418E5BF6E31127F1DCFD

>> path-thru http://rowery.olsztyn.pl/?r=10
== %/C/ProgramData/Red/cache/B4/B4BCF480EF4282064228FDE953F255B7

Pijoter

### DIFF
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -825,7 +825,7 @@ path-thru: function [
 	so: system/options
 	unless so/thru-cache [make-dir/deep so/thru-cache: append copy so/cache %cache/]
 
-	hash: checksum form file 'MD5
+	hash: checksum form url 'MD5
 	file: head (remove back tail remove remove (form hash))
 	path: dirize append copy so/thru-cache copy/part file 2
     unless exists? path [make-dir path] 
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -824,12 +824,12 @@ path-thru: function [
 ][
 	so: system/options
 	unless so/thru-cache [make-dir/deep so/thru-cache: append copy so/cache %cache/]
-	
-	if pos: find/tail file: to-file url "//" [file: pos]
-	clear find pos charset "?#"
-	path: first split-path file: append copy so/thru-cache file
-	unless exists? path [make-dir/deep path]
-	file
+
+	hash: checksum form file 'MD5
+	file: head (remove back tail remove remove (form hash))
+	path: dirize append copy so/thru-cache copy/part file 2
+    unless exists? path [make-dir path] 
+	append path file
 ]
 
 exists-thru?: function [
--- a/tests/source/environment/functions-test.red
+++ b/tests/source/environment/functions-test.red
@@ -387,7 +387,7 @@ Red [
 
 ===start-group=== "path-thru tests"
 	--test-- "path-thru test"
-		--assert not none? find (path-thru http://red-lang.com) "/cache/red-lang.com"
+		--assert not none? find (path-thru http://red-lang.com) "/cache/33/334C4A4C42FDB79D7EBC3E73B517E6F8"
 ===end-group===
 
 ===start-group=== "exists-thru? tests"
--- a/tests/source/environment/functions-test.red
+++ b/tests/source/environment/functions-test.red
@@ -387,7 +387,7 @@ Red [
 
 ===start-group=== "path-thru tests"
 	--test-- "path-thru test"
-		--assert not none? find (path-thru http://red-lang.com) "/cache/33/334C4A4C42FDB79D7EBC3E73B517E6F8"
+		--assert not none? find (path-thru http://red-lang.com) "/cache/91/91DD75833FAA7FF66B9BD4638549782B"
 ===end-group===
 
 ===start-group=== "exists-thru? tests"


Oldes:
To have Travis pass, you must modify also the tests.. like: https://github.com/red/red/blob/master/tests/source/environment/functions-test.red#L388-L391

greggirwin:
@pijoter is there any other ticket or chat on this, describing the purpose for the change?

pijoter:
this PR is related to issue #3117; current solution is somewhat broken because many different urls could be mapped to the same cache file

greggirwin:
Thanks @pijoter! Much clearer now. As is often the case, this opens more questions. 

Today, there is no distinction based on URL scheme either, which is bad. You may get different content via FTP vs HTTP, with the rest of the URL being the same, yes? But they map to the same local file in the cache. In that context, this is also a good change.

But this change comes at a cost as well. We lose the ability to find things in the cache without tooling. That is, we can no longer look in the cache, find the sub-dir for a domain, and clear that or just a single item in it. Cached content will also not even be grouped together by domain.

There are arguments for both sides, that you should never look directly at the cache in the first place, and that being human friendly is a good thing with a lot of value. First, of course, we should be correct, which is only true in the simplest case today. However, including the query and fragment parts of the URL seem risky as, especially for the query, results may vary on each read which invalidates the purpose of the cache. i.e., once cached, you're just going to read stale data from there. We can argue that, if you know that a query is dynamic, you just shouldn't use `*-thru` funcs, or use `/update` when reading, but is it best to put that on the user?

What we need to do is define the goals and desired/intended behavior clearly.

1) State the purpose of the funcs. Why do they exist?
2) Be correct. Any difference in a URL should map to a different local cache resource.
3) Be smart. Should query params be considered, or never cached if they exist in the URL?
4) Be easy to manage. Use human friendly names and direct mappings as much as possible.
5) Be clear. Document how it works and why it works that way.

pijoter:
> But this change comes at a cost as well. We lose the ability to find things in the cache without tooling.

this is just (internal) cache - why it need to be user readable? we need a method to  clearly map file! or url! (or something else) to cache file. md5/sha1 hash functions surely meet this requierement better than current approach; 

sure, it would be nice to reuse cache files somehow  but we do not know a-priori that additional parameter in the url is important on not; IMHO it is faster and cheaper to just make cache files when needed and remove them if stalled (but clear-cache func is still missing)

Oldes:
I agree with @pijoter that searching cache is not needed in most cases. How often do you search your browser cache? I also think, that `*-thru` functions should use http header to get info, if there is a new version of the content (if one has internet connection).

greggirwin:
We have a +1 from @Oldes, with perhaps a further enhancement possible. It doesn't look like Red uses `-thru` funcs internally anywhere, and so would not be affected by this change. I'm going to post on Gitter, to see if others have comments, but I'm inclined to merge this now.

giesse:
I think it is a good idea in general. But @greggirwin does make a pretty good point. I am inclined  to believe that doing it correctly while keeping a human friendly dir structure is not worth the effort, however, there are definitely ways to do so if people do indeed poke into the cache often.

(I'm thinking that if you are poking into the cache manually, then doing so from the Red console rather than directly is not a big downside, so we could have something like `clear-thru` to remove anything that matches a pattern URL; it would require us storing metadata for the cache somewhere though.)

hiiamboris:
You need to poke into a cache dir you can obtain it with `path-thru`, what else is there needed? I liked the change from the beginning. Perhaps in a wiki or somewhere it should be noted that `-thru` funcs do not parse and sort the query arguments nor removes the anchors (not sure if latter was a good idea but whatever), so if user adds those randomly they might render same page into different cache points. It was clear from the unmangled case but now will require a mention or a bit of a console play.

giesse:
@hiiamboris let's say you want to delete everything in the cache from www.google.com - `path-thru` is not going to help you there. Unless we use a reversible function, the only way to do this is to store the inverse mapping somewhere (ie. metadata file next to the cache file, or global metadata file with list of all cached files etc.).

I honestly don't know if it's worth having.

greggirwin:
My original use case was a bit specialized. A desktop app that downloaded images for viewing, organized by %project/camera/. Of course, you can do this outside the cache, or use funcs to clear, but it was very convenient to delete data no longer needed via the file manager.

Oldes:
What about having directory as a domain name and hashed file name of the rest of the url? Btw.. the cash should also deal with content expiration.

Oldes:
Now I see I already mentioned it. The key thing is that the directory must contain also port replacing the `:` char with simething valid. It should be also easy to write the metadata into such a dir wchich could be used to clear expired content when required.

hiiamboris:
> with something valid like _at_

or like `%3A`? `_at_` can be a part of some peculiar domain name after all

dockimbel:
I am coming a bit late at the discussion, but why not simply sanitize the special characters and keep the names readable, instead of obfuscating them? Isn't our moto to be human-friendly?

Oldes:
If you count also the query part.. you can get too long file names very quickly. I think using compromise is the best way in this case.

greggirwin:
We could keep the base filename, and append the hash of any query params if they make the filename too long. That's easy to work around. I'll mock up new mezzanines, unless someone else wants to, to compare an improved version of the old behavior with this. 

I really like keeping things simple, and as much as we think people shouldn't poke the cache, or interact with it, by hashing the names we lose the ability for normal people to ever look at it. 
```
User: Hey, my data is wrong.
Tech support: Remove that file from the cache.
User: How?
Tech support: Open a Red console and type...
```
Since the cache dir is configurable, multiple apps could also each have their own cache. And once the filenames are hashed, you can never find out what he source name of a cached file was.

greggirwin:
Here are some very old R2 funcs that deal with a related issue. 
```
    safe-filename: func [
        [catch]
        "Change invalid characters to underscores to make them valid filenames; limit length. Do not pass in fully qualified filenames."
        file [any-string!]
        /local bad-char mark
    ][
        if any [
            empty? file
            #"." = first file
            #"." = last file
        ][
            throw make error! join [script invalid-arg] file
        ]
        bad-char: union charset [#"^(0)" - #"^(1F)"] charset {<>:"/\|?*}
        ; " commented single quote to reset syntax highlighting.
        file: copy file
        parse file [any [mark: bad-char (change mark #"_") | skip]]
        head clear at file 255
    ]

    safe-filename: func [
        [catch]
        "Changes invalid characters to underscores and limits length."
        file [any-string!] "Unqualified filenames."
        /local bad-char mark illegal-windows illegal-NTFS
    ][
        if any [empty? file   #"." = first file   #"." = last file] [
            ;throw make error! join [script invalid-arg] file
            return none
        ]
        ; The following are illegal in Windows:
        illegal-windows: [
            %CON %PRN %AUX %NUL
            %COM1 %COM2 %COM3 %COM4 %COM5 %COM6 %COM7 %COM8 %COM9
            %LPT1 %LPT2 %LPT3 %LPT4 %LPT5 %LPT6 %LPT7 %LPT8 %LPT9
        ]
        if find illegal-windows to file! file [
            ;throw make error! join [script invalid-arg] file
            return none
        ]
        ; The following are used internally by NTFS
        illegal-NTFS: [
            %$Mft %$MftMirr %$LogFile %$Volume %$AttrDef %$Bitmap %$Boot %$BadClus
            %$Secure %$Upcase %$Extend %$Quota %$ObjId %$Reparse
        ]
        if find illegal-NTFS to file! file [
            ;throw make error! join [script invalid-arg] file
            return none
        ]
        
        bad-char: union charset [#"^(0)" - #"^(1F)"] charset {<>:"/\|?*}
        ; " commented single quote to reset syntax highlighting.
        file: copy file
        parse file [any [mark: bad-char (change mark #"_") | skip]]
        head clear at file 255
    ]

    safe-dir-name: func [
        "Change invalid characters to underscores to make them valid directory names; limit length. Do not pass in fully qualified dir names."
        file [any-string!]
    ][
        safe-filename file
    ]
```

Issue #3132, opened 2017-12-03
replace/all infinite loop


pijoter:
### Expected behavior
>> replace/all "test" charset [#"t" #"e"] ""
== "s"
### Actual behavior
infinite loop 
### Steps to reproduce the problem
as above
### Red version and build date, operating system with version.
red 0.63
25-Nov-2017/16:24:59+01:00
Windows 10.0.16299.64

meijeru:
Not a recent problem, confirmed also on W10 of 9-Aug-2017.

PR #3134, opened 2017-12-03
FIX: replace/all infinite loop #3132

endo64:


meijeru:
You replace `change` by `insert remove`. But both have the same effect, don't they??

PR #3138, opened 2017-12-09
FIX: OS-request-dir/dir and OS-request-file/file ignore string! values

endo64:
Fixes issue 3074 (`request-dir` and `request-file` ignore string! values)


dockimbel:
@qtxie So what about this PR?

qtxie:
I think it's ok.

PR #3181, opened 2018-01-02
Added error for unsupported save/as formats

gltewalt:
Currently it just uses exit and fails silently.


greggirwin:
Looks like a good fix @gltewalt.

qtxie:
@gltewalt Good fix. :+1:

PR #3182, opened 2018-01-03
Fix `save/as` with none! format, and add `save` tests.

greggirwin:


qtxie:
@greggirwin Thanks! 😄

PR #3189, opened 2018-01-04
Added cause-error to load. Updated docstrings to reflect available codecs.

gltewalt:


qtxie:
@gltewalt Good work. Thanks!

Issue #3190, opened 2018-01-05
WISH: /ONLY and /CASE refinements for ALTER


9214:
I wish I could use `alter` with `/only` refinement to achieve something like this:
```Red
>> alter [foo bar][foo bar]
== false
>> alter x: [foo bar][foo bar]
== false
>> x
== [bar]
>> alter/only x: [foo bar][foo bar]
== true
>> x
== []
```

[R3's `alter`](http://www.rebol.com/r3/docs/functions/alter.html) also has `/case`, which could be added as a nice bonus.

greggirwin:
If we want to do that, can we de-obfuscate the body as well? :^\

dockimbel:
@greggirwin Not my code. ;-P

greggirwin:
I'm also going to question the design of `alter`, looking at the tests. If given a value that is more than one item long, it only removes the first if the item is found. `Alter` is currently only used once in %VID.red and once in %view-test.red. 

I'm happy to take this one on Doc. Care to collaborate on the design @9214?

greggirwin:
Should we rename `alter` to `toggle`? I have old notes on that.

greggirwin:
Hmmm, opens up the question of `/only` as well.

greggirwin:
Actually, `/only` should be implied. We should be toggling a single value in all cases, yes?

greggirwin:
I think I have this in good shape, with bitset!, /case, and /only support. Will post shortly.

greggirwin:
```
toggle: function [
	"Toggle a value (append/remove), returning the toggled state (on/off)"
	input [series! bitset!] "If value is found, remove it; otherwise, append it"
	value "Must be an integer or char if input is a bitset"
	/case "Perform a case-sensitive match; ignored for bitset inputs"
][
	; /case is safely ignored for bitsets
	pos: either case [find/only/case input :value] [find/only input :value]
	; The easy case is that the value wasn't there. /only is safely ignored for bitsets.
	; Use `not` instead of `none?`, though, because `find` on bitsets returns a
	; logic value, not an index.
	either not pos [append/only input :value  (true)][
		; Bitset handling is a bit...different. They don't have a current
		; index, so the `/part` arg for `remove` is the bit index.
		either bitset? input [remove/part input :value][
			; If we want to support strings as both inputs and values, we
			; have to remove more than one item (a char) if value is a string.
			; But we DON'T want to remove more than one part if the input
			; isn't a string.
[...]

PR #3191, opened 2018-01-08
WISH: #3190 /ONLY and /CASE refinements for ALTER

greggirwin:
`/only` is implied with this new version of `alter`, and now works "correctly". The original was a bad design and would produce changing results if given a block of values or a string as the value. Bitset support also added. Was going to suggest changing the name to `toggle`, but holding off on that. 

The only places `alter` is used today:

- https://github.com/red/red/blob/master/modules/view/VID.red#L211
- https://github.com/red/red/blob/master/tests/view-test.red#L440

Many more tests added, making sure those use cases are covered.

### DIFF
--- a/modules/view/backends/windows/draw.reds
+++ b/modules/view/backends/windows/draw.reds
@@ -669,11 +669,11 @@ draw-short-curves: func [
 		pt: pt + 1
 		pt/x: ( 2 * ctx/other/path-last-point/x ) - control/x
 		pt/y: ( 2 * ctx/other/path-last-point/y ) - control/y
+		control/x: pt/x
+		control/y: pt/y
 		pt: pt + 1
 		pt/x: either rel? [ ctx/other/path-last-point/x + pair/x ][ pair/x ]
 		pt/y: either rel? [ ctx/other/path-last-point/y + pair/y ][ pair/y ]
-		control/x: pt/x
-		control/y: pt/y
 		pt: pt + 1
 		loop nr-points - 1 [ pair: pair + 1 ]
 		if pair <= end [


PeterWAWood:
@greggirwin I took a quick look at the PR:

1. (From my point of view) it would be better to use "list" as a pseudonym for [series! bitset!]. Logically a bitset! is a series but is implemented very differently. List seems appropriate in these circumstances.

2. `list` would be a better name than 'input' for me.

3. Re: comments - [Comment style](https://doc.red-lang.org/en/style-guide.html#_comments)

4. The tests look good though we should probably include checks of these cases:
    String changing from ISO-8859-1 to UTF-16 to UTF-32 and vice-versa
    Case handling with non-ASCII characters (e.g é matching E)
    (I can help with the tests if you need.)

5. I found this behaviour a little unexpected:
```text
>> a: [1 2]
== [1 2]
>> alter a [1 2]
== true
>> a
== [1 2 [1 2]]
```
6. The comment say that `value` must be an integer! or char! with a bitset! but `alter` processes a string:
```text
>> b: charset[#"a"-#"z"]
== make bitset! #{0000000000000000000000007FFFFFE0}
>> alter b #"A"
== true
>> find b #"A"
== true
>> alter b "BCDE"
== true
>> find b "ABCDE"
== true
```

PeterWAWood:
I can see that  having "/only" as the default makes sense for `after` is that it makes `alter` inconsistent with other functions which insert values. I think there is more benefit in having consistency than the benefit to the alter function. The `/only` refinement as used extensively, is itself inconsistent but I don't think it is going to get changed.

Also, from a language consistency perspective, it would seem best to return the modified series rather than true or false.

As for the coding standards, I defer to @dockimbel

PeterWAWood:
Yes, I see the problem with expanding series upon insertion. It seems that `alter` has been designed with scalar values in mind. I'm not really sure of the use case for it especially when it comes to inserting and removing non-scalar values. Non-scalar values throw up some interesting design issue.

For example, how to handle objects:
```text
>> blk: []
== []
>> o: make object! [a: 1]
== make object! [
    a: 1
]
>> oo: make object! [a: 1]
== make object! [
    a: 1
]
>> equal? o oo
== true
>> same? o oo
== false
>> alter blk o
== true
>> blk
== [make object! [
    a: 1
]]
>> alter blk oo
== false
>> blk
== []

```

greggirwin:
Indeed. For objects you may want to alter based on either sameness or equality.

Issue #3196, opened 2018-01-10
Red console not compiling on a Linux 32bits


gregit:
### Expected behavior
First time running RED executable should compile the console.

### Actual behavior
`Compiling compression library...
Compiling Red console...
*** Runtime Error 23: illegal operand
*** at: 0806202Dh
`
note: using wine and windows binary works

### Steps to reproduce the problem
launch linux executable:
 `./red-063`

### Red version and build date, operating system with version.

Debian Stretch

`$uname -r
4.9.0-5-686-pae`

`$ lscpu
Architecture :        i686
Mode(s) opératoire(s) des processeurs : 32-bit
Boutisme :            Little Endian
Processeur(s) :       1
Liste de processeur(s) en ligne : 0
Thread(s) par cœur : 1
Cœur(s) par socket : 1
Socket(s) :           1
Identifiant constructeur : GenuineIntel
Famille de processeur : 15
Modèle :             2
Nom de modèle :      Intel(R) Pentium(R) 4 CPU 2.40GHz
Révision :           7
`

dockimbel:
The automatic check for SSE3 presence has only been implemented on Windows so far, so the compiled code generates some SSE3 instructions, resulting in a crash.

It should just need a couple of lines of code to implement such detection reading from `/proc/cpuinfo` from `red.r`.

In the meantime, you can also manually compile the Red console:
1. Add `cpu-version: 1.0` entry into the `Linux` target definition in `system/config.r`
2. Compile the console using manually using:
`do/args %red.r "-r %environment/console/console.red`

gregit:
```
./red-13jan18-a27989f 
** Script Error: Invalid argument: flags any space :
** Where: context
** Near: SSE3?: parse cpuinfo [any [thru ["flags" any space ":"] to "sse3" to newline to end]]
```

PR #3197, opened 2018-01-11
Add check for SSE3 instruction on Linux

Zamlox:
**FIX: added two functions to export list: decode-utf8-char and form-signed**
   - this commit adds two functions to the export list of libred. I couldn't compile red console on my Ubuntu 16.04 without this change

**FIX: issue #3196 (missing check for SSE3 instructions on Linux)**
   - this commit will check for string "sse3" in /proc/cpuinfo using a parse rule and it will set SSE3? to true if it founds or false otherwise


qtxie:
Thx! 👍

PR #3209, opened 2018-01-24
FEAT: query action (WIP - only returns file modification date on Windows so far)

Oldes:


Oldes:
@dockimbel as there is already `size?` as a native, I'm quite not sure, if I should not create `modified?` as a native too.

Oldes:
But it was not finished!

Issue #3156, opened 2017-12-21
Crash when assigning result of function in object to a word


rebolek:
### Expected behavior

```
>> c: context [f: func[a][""]]
== make object! [
    f: func [a][""]
]
>> t: c/f []
== ""
```

### Actual behavior

```
>> c: context [f: func[a][""]]
== make object! [
    f: func [a][""]
]
>> t: c/f []

*** Runtime Error 1: access violation
```

### Steps to reproduce the problem

See above. Also, here's stack trace:

```
*** Runtime Error 1: access violation
*** in file: /home/sony/Code/red/runtime/datatypes/context.reds
*** at line: 283
***
***   stack: red/_context/set F6A0D7ECh F6A0D7FCh
***   stack: red/word/set
***   stack: red/interpreter/eval-expression F6805564h F6805564h false false false
***   stack: red/interpreter/eval F6A0D7CCh true
***   stack: red/natives/catch* true 1
***   stack: ctx357~try-do F7F50D6Ch
***   stack: ctx357~do-command F7F50D6Ch
***   stack: ctx357~eval-command F7F50D6Ch
***   stack: ctx357~run F7F50D6Ch
***   stack: ctx357~launch F7F50D6Ch
***   stack: ***_start
```

### Red version and build date, operating system with version.

Custom build from latest sources (21-Dec-2017/10:47:01+01:00), Ubuntu 17/10.

meijeru:
Looks like regression after this commit: [9178838](https://github.com/red/red/commit/9178838f7569b6e3d28fedcf070f04ab7e6e4e33)

rebolek:
@meijeru Actually, it's this commit: https://github.com/red/red/commit/bca86785936b0750b994702daa15734361237f1f

meijeru:
So @dockimbel and @qtxie have enough food for thought :smile:

dockimbel:
Thanks for identifying those culprit commits, that's a significant time-saver for us! :+1:

dockimbel:
Relates to #3128

rebolek:
You're welcome!
Actually, I needed working console, so I had to find which commit works ;)

toomasv:
works with explicitly set type:
```
t: to-string c/f []
```

9214:
Maybe related (crashes the console)
```Red
graph: make deep-reactor! [
    data: [
        red  10x10 10
        blue 40x40 20
    ]
]

view/no-wait [
    base 500x500 with [
        draw: collect [
[...]

PR #3237, opened 2018-02-28
Fix for #3156 and formatting for input/ask

gltewalt:


environment/console/input.red
```diff
@@ -579,7 +579,7 @@ ask: function [
 ][
 	buffer: make string! 1
 	_set-buffer-history buffer head system/console/history
-	_read-input question
+	_read-input append question space
```
Oldes: What is someone don't want to have a space as a part of the question for some reason?

Issue #3128, opened 2017-12-01
Problem with functions from within objects


toomasv:
### Expected behavior
```
n: object [l: none x: none asc: has [a l][l: copy [] a: x while [a][insert l a a: a/x] copy l]]
n1: make n [l: "1" x: none]
n2: make n [l: "2" x: n1]
foreach y n2/asc [probe y/l]
"1"
== "1"
```
### Actual behavior
```
n: object [l: none x: none asc: has [a l][l: copy [] a: x while [a][insert l a a: a/x] copy l]]
n1: make n [l: "1" x: none]
n2: make n [l: "2" x: n1]
foreach y n2/asc [probe y/l]
*** Script Error: invalid argument: make object! [l: "1" x: none asc: func [/local a l][l: copy [] a: x while [a] [insert l a a: a/x] copy l]]
*** Where: foreach
*** Stack:  
```
### Steps to reproduce the problem
As above
### Red version and build date, operating system with version.
0.6.3 
```
>> system/build
== make object! [
    date: 23-Nov-2017/9:38:11+02:00
    config: make object! [
        config-name: 'Windows
        OS: 'Windows
        OS-version: 0
        [...]
```
### Additional notes
```
>> type? n2/asc
== block!
```
Whith `copy` and `to-block` the function works:
```
>> foreach y copy n2/asc [probe y/l]
"1"
== "1"
>> foreach y to-block n2/asc [probe y/l]
"1"
== "1"
```
Also, it works in global scope:
```
n: object [l: none x: none]
asc: func [n /local a l][l: copy [] a: n/x while [a][insert l a a: a/x] copy l]
n1: make n [l: "1" x: none]
n2: make n [l: "2" x: n1]
>> foreach y asc n2 [probe y/l]
"1"
== "1"
```
But not in another object:
```
>> m: object [asc: func [n /local a l][l: copy [] a: n/x while [a][insert l a a: a/x] copy l]]
== make object! [
    asc: func [n /local a l][l: copy [] a: n/x while [a] [insert l a a: a/x] copy l]
]
>> foreach y m/asc n2 [probe y/l]
*** Script Error: invalid argument: make object! [l: "1" x: none]
*** Where: foreach
*** Stack:
``` 
And worst, while trying to set a word to the output of `n2/asc` the console crashes.

toomasv:
[...]

PR #3240, opened 2018-02-28
FIX: issue #3128  

gltewalt:


dockimbel:
I have renamed the PR, as #3156 was fixed by another commit (different fix). This PR is still valid though for fixing #3128.

PR #3257, opened 2018-03-13
Made `replace` to accept parse rules while replacing in strings

toomasv:


nedzadarek:
`replace` doesn't accept function as the third argument (`value`). It insert `?function?` instead of calling the function.

I changed this line https://gist.github.com/toomasv/399e3507c11a51954cfaa00eaa04237c#file-replace-red-L29 to:
`p: compose [to pattern change pattern (value)]`
Here is my updated version:
https://gist.github.com/nedzadarek/828cff5b24dae68b2a81f7193623c55d
It needs testing.

dockimbel:
@nedzadarek It seems to work fine here:
```
>> replace [1] 1 :to-word
== [func ["Convert to word! value" value][to word! :value]]
```

@toomasv Can you confirm that it still works after your changes in this PR?

dander:
@toomasv looks like you converted from tabs to spaces

toomasv:
@dockimbel It works, but not with functions, as @nedzadarek noticed.
With 
```
p: compose [to pattern change pattern (value)] ; instead of `p: [to pattern change pattern value]`
```
it works also with functions (but not with words copied in pattern -- this is similar to how `replace` with block works).

Example:
```
>> replace' "Seems not to work" [#"n" thru "t "] ""
== "Seems to work"
>> replace' "Seems not to work" [#"n" thru "t "] does [""]
== "Seems to work"
```
@dander Yeah, somehow happened, didn't notice.

greggirwin:
@toomasv if you fix the tabs issue, and add a couple tests for the new behavior, I think we can merge this. Then we can review if other changes are desired.

toomasv:
@greggirwin I made a new pull-request for `functions` with changed `replace` and for `replace-tests`. But I am not sure I made everything correctly.

toomasv:
Locally all tests were ok.

greggirwin:
@toomasv sorry this sat long enough to become subject to merge conflicts. If you fix those, I think we can merge this, assuming all tests pass.

endo64:
> @toomasv sorry this sat long enough to become subject to merge conflicts. If you fix those, I think we can merge this, assuming all tests pass.

@greggirwin I think this functionality is already implemented: 
```
>> replace/all "a1b2c3d4" [d] "-"
== "a-b-c-d-"
```

and @toomasv 's tests: 

```
>> replace "Seems not to work" [#"n" thru "t "] ""
== "Seems to work"
>> replace "Seems not to work" [#"n" thru "t "] does [""]
== "Seems to work"
>> replace/all "Seems not to not work" [#"n" thru "t "] does [""]
== "Seems to work"
>> replace/all "Seems not to not work" [#"n" thru "t "] ""
== "Seems to work"
```

What @toomasv added to the function is already there now:

image

@toomasv can you also check, if so @greggirwin we are good to close this PR.

toomasv:
Yes, I checked the code, my addition is included already.

PR #3281, opened 2018-03-25
FIX: extend system/console context

9214:
Without `terminate: []` compilation in release mode with this included file will fail (in case user wants to use `ask`).
```
*** Compilation Error: word terminate not defined in system/console/terminate 
```


9214:
I'm closing this PR since @qtxie pushed a better fix.

PR #3335, opened 2018-04-13
Re-add fix for to-UTC-date + remove uneeded compose in about func

x8x:
Commit https://github.com/red/red/commit/be7ff3a992ff8e128e2c051ed474c70ec55792de was added to fix issue with `to-UTC-date` added at commit https://github.com/red/red/commit/1a598cd1d04085cdc062025ee68866777b6b4eca that crashed GUI Console.
Commit https://github.com/red/red/commit/a60656d0074dee4b1377bb8ae3302350555f02fc reverted that commit and the crash came back.

Also removed unneeded `compose` from `about` function.

### DIFF
--- a/environment/console/help.red
+++ b/environment/console/help.red
@@ -598,14 +598,12 @@ help-ctx: context [
 		plt: system/platform
 		either debug [
 			print either git [
-				compose [
-					"-----------RED & PLATFORM VERSION-----------" lf
+				[
 					"RED: [ branch:" mold git/branch "tag:" mold git/tag "ahead:" git/ahead
 					"date:" to-UTC-date git/date "commit:" mold git/commit "]^/"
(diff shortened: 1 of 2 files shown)


greggirwin:
@x8x, I don't want to make you explain every detail, but I'm not clear on the fixes in the low level code, why they're needed, and if they're safe. My lack of understanding. e.g. disabling owner checking now occurs both before and after the stack value is set, which seems like a big change, and limited testing on the mezz to-UTC-date doesn't show the problem, so I'm trying to connect the dots.

qtxie:
So you add back commit be7ff3a ? But this fix is wrong.

x8x:
@greggirwin Right! I really have no clue, must be some low level binding or something in `date!` or `time!`, I don't see anything wrong with `to-local-date` function. May need @qtxie or @dockimbel expertise.

PR #3345, opened 2018-04-23
FIX: HELP function fixed for map! and object! values

endo64:
Help string "is an object! value" was missing:

```
>> o: context [a: 1]
```

Before fix:

```
>> ? o
     a  integer!      1
```

After fix

```
>> ? o
O is an object! with the following words and values:
     a  integer!      1
```

Same for `map!` values.

Also `if not`s replaced by `unless`.

### DIFF
--- a/environment/console/help.red
+++ b/environment/console/help.red
@@ -205,7 +205,7 @@ help-ctx: context [
 		if system/words/all [not opt  not all][
 			clear find blk refinement!
 		]
-		if not all [
+		unless all [
 			remove find blk to set-word! 'return
 			clear find blk /local
 		]
@@ -262,7 +262,7 @@ help-ctx: context [
 			;	the param type spec, but they are just words in Red func specs.
 			param-type=: [
 				set =val block! (emit 'type =val) (
-					if not any [
+					unless any [
 						parse reduce =val [some [datatype! | typeset!]]
 						parse =val ['function! block!]
 					][
@@ -358,7 +358,7 @@ help-ctx: context [
 				]
 			]
 		]
-		if not found-at-least-one? [
+		unless found-at-least-one? [
 			_print ["No" type "values were found in the global context."]
 		]
 	]
@@ -381,14 +381,14 @@ help-ctx: context [
 		word [word! path!]
 	][
 		fn: either any-function? :word [:word][get :word]
-		if not any-function? :fn [
+		unless any-function? :fn [
 			print "show-function-help only works on words that refer to functions."
 			exit
 		]
 
 		; Convert the func to an object with fields for spec values
 		fn-as-obj: parse-func-spec :fn
-		if not object? fn-as-obj [
+		unless object? fn-as-obj [
 			print "Func spec couldn't be parsed, may be malformed."
 			print mold :fn
 			exit
@@ -412,12 +412,12 @@ help-ctx: context [
 			DENT_1 word-is-value-str/only word
 		]
 
-		if not empty? fn-as-obj/params [
+		unless empty? fn-as-obj/params [
 			_print [newline "ARGUMENTS:"] 
 			foreach param fn-as-obj/params [_print [DENT_1 form-param param]]
 		]
 		
-		if not empty? fn-as-obj/refinements [
+		unless empty? fn-as-obj/refinements [
 			_print [newline "REFINEMENTS:"] 
 			foreach rec fn-as-obj/refinements [
 				_print [DENT_1 as-arg-col mold/only rec/name DOC_SEP either rec/desc [dot-str rec/desc][NO_DOC]]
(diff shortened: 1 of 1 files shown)


greggirwin:
@endo64, I'm fine with adding the new help text. That's a good change. But arbitrarily changing all `if not`'s to `unless` isn't justified. It's a subjective call, unless it's in a performance critical area where it will make a measurable difference.

If you want to submit a PR with just the help text change, I'll review again and merge that.

Thanks!

meijeru:
I would agree the wholesale change of `if not` to `unless` was uncalled-for, but in the spirit of "egoless programming" I could have *suggested* the same change. If @endo64 would have merely suggested it, without doing the actual changes, I could even imagine @greggirwin would have seriously considered it.

Oldes:
@greggirwin Instead of writing: 
_It's a subjective call, **unless** it's in a performance critical area where it will make a measurable difference._

maybe you should write:
_It's a subjective call, **if** it's **not** in a performance critical area where it will make a measurable difference._

greggirwin:
TL;DR

- I don't believe egoless programming is always a good thing
- We should have a reason for changing code stylistically
- If you stylize existing code, be prepared to take ownership
- `Unless` feels different at the head of a sentence, versus the middle
- I didn't say we should *never* use `unless`.


I feared this would be an issue. I'll defend my stance a bit more, as this 
topic will last for the rest of our lives. We can make notes in the 
style guide, and put it under the eternal "tabs vs spaces" war. I
think this is worth the discussion, and will go into some detail here.

Let me start by saying that I consider @endo64 a friend. I've known him
for a long time, value his opinion, and deeply respect his technical
ability. Nothing I say here should be taken as a criticism toward him,
in any way. I understood his changes to be an attempt to improve the
code.

When we worked on `help` in the past, Oldes and I had different styles
and approaches to some things, and I think we ended up with a blend of
the two. We all have our biases and styles. It's a strength, and core
principle of Red that we can express things in different ways, rather
than treating code as a commoidity that should always look the same,
as other language designs encourage.

When `unless` was added to Rebol, Carl made this note:

> This function can take some getting used to. It has been provided to 
make PERL programmers happier, and it's marginally simpler and faster 
than writing an if not expression.

While I defer to Carl in almost every case, some later changes to Rebol
weren't made with the same deep thought as the core language (IMO). This
is one of those cases. In an appeal to authority, Perl Best Practices
even recommends against its use. If people use it because it seems clearer
to them, which it doesn't to me, that's fine. But if people use it in the
general case just because it's faster, I think that's a poor use of it.
Save it only for the cases where the performance warrants it. Otherwise
we end up with a mix of 2 equivalent constructs where the negation isn't
always clear. That's my bias. `Unless`, at the beginning of a sentence,
does not read well to me, but that may be just me. It may also work better
for non-native English speakers, because of word order in their native 
language. And there are cases where `unless` will be clearer.

Rewriting code is not the same as improving or fixing code, but may be 
purely stylistic changes. If you're truly egoless, and I have yet to meet
a programmer who is, you may not care if someone changes the style of your
code, but those changes may still affect the effort required to maintain it.
If someone changes code, with no clear benefit, that is also not egoless
programming.

We need to balance Ownership Mentality versus changing things for the sake
of changing them. And we've probably all had the experience of going back
to old code we wrote, and dropping right into but, compared to code we 
wrote, which others have tinkered with, and makes us doubt if we wrote it
in the first place.

Q. How many other negative branch constructs do we have?

I wonder if my bias against is relates to this. It is an exception to the
rule of control constructs. 

Perl wasn't the first language to have `unless`. It goes way back. But it's
also used as a control modifier in some languages that support it, which 
I...need to form an opinion on. :^)

- https://softwareengineering.stackexchange.com/questions/65939/whats-the-first-language-that-had-the-unless-conditional-loop-built-into-itse
- http://www.perltutorial.org/perl-unless/#Perl_unless_statement_guidelines
- http://www.railstips.org/blog/archives/2008/12/01/unless-the-abused-ruby-conditional/
- https://en.wikipedia.org/wiki/BASIC-PLUS

endo64:
> I consider @endo64 a friend

I definitely am, and didn't take any of your words as criticism on me. No worry on this.
And thank you for the kind words.

> If you stylize existing code, be prepared to take ownership

To be honest I never thought about the ownership of code, as I see the whole Red project as one, that everyone makes changes piece by piece and it is accepted or rejected by the project owner/team.
But you're right this should be considered especially if the changes are about the coding styles.

I saw a similar commit that changes `if not`s to `unless` long time ago, so I thought that it could be desired.

> It may also work better for non-native English speakers

Actually for me, as a Turkish speaker, using `unless` was very difficult for me in the beginning and still is in some cases. Because my brain has to compute the sentences when I see `unless` as it is more English-ish than a programming language expression. But that may be just me.

But lately I force myself to use more `unless` instead of `if not` as it is native and one word. But this of course is a personal preference.

Thanks for the discussion!

PR #3355, opened 2018-04-28
Added unofficial Red builds link to README

gltewalt:


PeterWAWood:
I don't think this should be added to the Read Me. It is more appropriate for the wiki.

greggirwin:
I told Greg to go ahead and add that, thinking it was better than adding it to the red-lang downloads page.

What if we add a link to the official downloads above this in the readme? I don't want to bury it too much. Maybe Greg's other changes to the wiki will make a good location work as well.

dockimbel:
It's a bit dangerous for us to endorse external Red builds from servers we have zero control on.

greggirwin:
OK. @gltewalt, we'll close this, and just do it in the wiki.

greggirwin:
While I thought this would be a good thing to support, being an Open Source project, I looked at a few other langs, and most of them only have a brief readme that lays out the license and basics. I can also see that if we do it for one, it sets a precedent, and we do want to have a limited set of things we need to monitor and maintain, to ensure we don't link to unsafe resources.

Issue #3248, opened 2018-03-08
WISH: extract should be more compatible with Rebol


meijeru:
The `extract` function has a refinement `/index` with argument `pos [integer!]` which is superfluous because it can be replaced by the usage of `at`. Thus `extract/index some-series 2 4` is equivalent to `extract at some-series 2 4`. The refinement is perhaps there because of compatibility with Rebol, but in Rebol the refinement's argument is of `any-type!` instead of just `integer!` in Red. This would allow a series argument for the position, and that could be a reason for it to be there.

Thus, I would propose to broaden the type of the `pos` argument and thereby (a) make the refinement non-superfluous and (b) increase compatibility with Rebol.

An alternative could be to abolish the refinement altogether.

Rebol's `extract` has a further refinement missing from Red:

     /default -- Use a default value instead of none
         value -- The value to use (will be called each time if a function) (Type: any-type!)

but it is not clear to me what its use is, therefore I do not include that in this wish.

dockimbel:
Agreed for `/index` support for sake of compatibilty.

greggirwin:
>  but in Rebol the refinement's argument is of any-type!

@meijeru, I think this is a bug in R2. It was meant to be `[number! logic!]` but the doc string was put first, so the type was ignored. What is the use case/example for having a series argument as the index position?

meijeru:
@greggirwin The use case is the same as with any series function that allows an index position to be indicated by a series not at the head, think of copy/part etc.

greggirwin:
I understand the purpose of `/index` in the simple case, but the PR made for this (#3360) doesn't work that way, nor does R3 (which it's based on). They assume a block of indexes, to extract multiples. *That's* the case I doubt is justified at this point. And if I'm correct that R2 is bugged, and shouldn't allow `any-type!`, then we should only support `series!` for `pos`, if `/index` is not needed; and we should name the refinement `/part`.

However, I'd still like to see where this has ever been used, or what a real world use case is, where a `series!` value for `pos` helps. If we name the refinement `/part`, then `pos` becomes `length` and we can extract multiple sequential values in one shot. I still dont' see how a series val works for length there though. Rebol never had that, and I don't recall it ever coming up in the past.

meijeru:
[...]

PR #3360, opened 2018-05-01
extract/index using block! for pos argument by Heroide

Heroide:
Issue #3248

### DIFF
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -633,41 +633,28 @@ make-dir: function [
 	path
 ]
 
-forskip: function [
-	"Evaluates a block for periodic values in a series."
-	'word [word!] {Word set to each position in series and changed as a result}
-	skip-num [integer!] "Number of values to skip each time"
-	body [block!] "Block to evaluate each time"
-	/local orig result
-][
-	if 0 > skip-num [skip-num: negate skip-num]
-	if skip-num = 0 [skip-num: 1]
-	orig: get word
-	while [any [not tail? get word (set word orig false)]] [
-		set/any 'result do body
-		set word skip get word skip-num
-		get/any 'result
-	]
-]
-
 extract: function [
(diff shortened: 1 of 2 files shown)


greggirwin:
@Heroide thanks so much for the PR! The code looks clean, and you included tests, which is great. The problem is that it also includes `forskip` which is unrelated. I know reworking PRs can be a pain (at least for me), but `forskip`, while useful, has come up a lot in discussion but not yet been approved for inclusion in Red. Tough call. I've used it quite a bit myself in the past. We'll start a wiki page for things intentionally *not* carried over from Rebol and their discussion. 

From there we can decide more formally on `forskip`, and write a full set of tests for it if we decide to include it.

As a side note, `forskip` in R2 was used to make `forall` easy to write, as it's just a special case of `forstkip`. In Red `forall` is a native.

Heroide:
Yes, I did not like that too. Forskip is no good name. But code became messy without. And not working, even more not good. I thought of renaming but now it is more clear where source came from. I understand if forskip is in, people will find out and use. I see the problem.

greggirwin:
On further review, the formatting and some structure seemed odd for Red, and I now see it's basically a paste of the R3 version. With the question I just asked on the original wish ticket behind this, maybe we need to step back and look at `extract`'s design afresh.

greggirwin:
Since `/default` isn't there (which I agree with removing), it's `value` param is now a local, giving us both `val` and `value`, which is confusing.

Heroide:
I removed the bad function. And replaced confusing val and value with better names. Maybe still not Red like, but better.

greggirwin:
Thanks for the updates @Heroide. Let's move the chat to https://github.com/red/red/issues/3248 and look at `extract`'s design, to make sure we're building what we really need.

greggirwin:
I'm closing this, and the related ticket. See the more extensive chat there.

Issue #3369, opened 2018-05-07
accessing series over the board


hiiamboris:
### Test script
```
a: "12345678"
b: skip a 6
c: tail a
d: skip a 2
remove/part a 4
b
copy b
c
copy c
d
copy/part d b
clear a
d
copy d
copy/part d b
```
### Expected behavior
R2 crashes, but R3 got it right:
```
>> a: "12345678"
== "12345678"
>> b: skip a 6
== "78"
>> c: tail a
== ""
>> d: skip a 2
== "345678"
>> remove/part a 4
== "5678"
>> b
== ""
>> copy b
== ""
>> c
== ""
>> copy c
== ""
>> d
== "78"
>> copy/part d b
== "78"
>> clear a
== ""
>> d
== ""
>> copy d
== ""
>> copy/part d b
== ""
```
### Actual behavior
```
>> a: "12345678"
== "12345678"
>> b: skip a 6
== "78"
>> c: tail a
== ""
>> d: skip a 2
== "345678"
>> remove/part a 4
== "5678"
>> b
== ""
>> copy b
== "56"
>> c
== ""
>> copy c
== "5678"
>> d
== "34"
>> copy/part d b
== "3456"
>> clear a
== ""
>> d
== ""
>> copy d
== "12"
>> copy/part d b
== "3456"
```
### Red and platform version
```
-----------RED & PLATFORM VERSION----------- 
RED: [ branch: "master" tag: #v0.6.3 ahead: 688 date: 5-May-2018/11:50:50 commit: #9ce2aacce1509943ac11efa3b13457e8a66f03b4 ]
PLATFORM: [ name: "Windows 7 Service Pack 1" OS: 'Windows arch: 'x86-64 version: 6.1.1 build: 7601 ]
--------------------------------------------
```

greggirwin:
https://github.com/bitbegin/red/commit/c02cf06d2fe1e35d150021e293859e673b39bbea doesn't produce your expected output @hiiamboris. We may need to clarify each piece, to make sure we don't introduce regressions as well. Can we also name this ticket to include `series` and `copy` so we know exactly where the problem lies?

greggirwin:
Ah, not merged yet. My oversight. 

@bitbegin, could you sync fresh with the main Red repo master, so your small change doesn't involve 250+ commits in the merge?

greggirwin:
Nevermind, looks like already done in the PR. Thanks!

greggirwin:
[...]

PR #3372, opened 2018-05-07
#3369: fix series copy overflow

bitbegin:


greggirwin:
I'll go ahead and merge this, but please still review when you have a chance @qtxie.

greggirwin:
Also, @bitbegin and @hiiamboris, would one of you add some tests to go with the ticket that led to this fix, so we can catch regressions? Thanks.

Issue #3374, opened 2018-05-08
trim/with corrupts trim function


PeterWAWood:
### Expected behavior
>> trim/all {   a ^-1^/    ab2^-  ^/  ac3  ^/  ^/^/}
== "a1ab2ac3"
### Actual behavior
>> mstr: {   a ^-1^/    ab2^-  ^/  ac3  ^/  ^/^/}
== {   a ^-1^/    ab2^-  ^/  ac3  ^/  ^/^/}
>> trim/all {   a ^-1^/    ab2^-  ^/  ac3  ^/  ^/^/}
== "a1ab2ac3"
>> "    ^-1^/    2^-  ^/  c3  ^/  ^/^/" = trim/with copy mstr "ab"
== true
>> trim/all {   a ^-1^/    ab2^-  ^/  ac3  ^/  ^/^/}
== {^-1^/2^-^/c3^/^/^/}
>> 
### Steps to reproduce the problem
run the above code
### Red and platform version
commit 9ce2aacce1509943ac11efa3b13457e8a66f03b4 (HEAD -> master, upstream/master)
macOS 10.12.16

meijeru:
Looks like unintentional global?

PR #3378, opened 2018-05-09
FIX: issue #3374 (a simple fix method)

bitbegin:


bitbegin:
tests have been added.

PR #3387, opened 2018-05-11
FIX: patch KillTimer call after  #6fdf75bb

9214:
As always, more of a blind guess than anything. I'm naively assuming that `handle!` is just a shortcut for `int-ptr!`. Fix ix related to [this](https://github.com/red/red/commit/6fdf75bb) commit, also see [original bug report](https://gitter.im/red/bugs?at=5af602d6f04ce53632ce6080) on Gitter.


qtxie:
The `SetTimer` and `KillTimer` functions should use the same `nIDEvent`, so this fix is not correct. 
I have pushed a fix for it.

Issue #3385, opened 2018-05-10
NOW silently ignores some refinements


meijeru:
### Expected behavior

`now` should raise an error if it has more than one refinement and they are incompatible
### Actual behavior

`now/time/day` and `now/day/time` both act as if `/time` was not there, giving the result of `now/day`
### Steps to reproduce the problem

execute `now/time/day` in GUI console
### Red and platform version
```
Red 0.6.3 for Windows built 27-Apr-2018/12:07:23+02:00 commit #4c36108
```

endo64:
R2, R3 and Red have different behaviors on multiple refinements:
```
>> now/time/day
; Red== 11
; R2 ==  ** Script Error: Too many refinements
; R3 == 10:10:09
```

meijeru:
R2 seems the most sensible one

bitbegin:
I wanted fix this by disallow more than one refinements for now, but found another issue: "now/time/precise is allowed"

bitbegin:
i set these refinements (year month day time zone weekday yearday) to be uniqued, and `date precise utc` can be existed with above refinements at the same time.

dockimbel:
Too many combinations to check:
```

>> now/time
== 21:19:47
>> now/time/utc
== 19:19:49
>> now/utc
== 20-Aug-2019/19:19:52
>> now/utc/precise
== 20-Aug-2019/19:19:56.783413
>> now/time/utc
== 19:20:14
>> now/time/utc/precise
== 19:20:40.798931
```

PR #3390, opened 2018-05-14
FIX: #3385 (`now` no need support more than one refinements)

bitbegin:
>> now
== 14-May-2018/12:17:53+08:00
>> now/day
== 14
>> now/time
== 12:17:58
>> now/time/day
*** Script Error: Too many refinements
*** Where: now
*** Stack:

>> now/day/time
*** Script Error: Too many refinements
*** Where: now
*** Stack:

>> now/day/time/year
*** Script Error: Too many refinements
*** Where: now
*** Stack:

>>

### DIFF
--- a/environment/system.red
+++ b/environment/system.red
@@ -122,6 +122,7 @@ system: context [
 				bad-func-extern:	["invalid /extern value:" :arg1]
 				no-refine:			[:arg1 "has no refinement called" :arg2]
 				bad-refines:		"incompatible or invalid refinements"
+				too-many-refines:	"Too many refinements"
 				bad-refine:			["incompatible refinement:" :arg1]
 				word-first:			["path must start with a word:" :arg1]
 				empty-path:			"cannot evaluate an empty path value"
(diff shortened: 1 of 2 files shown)


bitbegin:
@greggirwin 

1.  `bad-refine` support only one argument,  but in some cases, there maybe are more than 2 args.
For example: `now/year/day/time/zone`.  I don't known how to show error messages for these situations

2. incompatible-refinements? is better name

greggirwin:
@bitbegin if you can collect the list of incompatible refinements, you can put them in a block, in the arg. e.g. 
```
>> cause-error 'script 'bad-refine [[/x /y /z]]
*** Script Error: incompatible refinement: [/x /y /z]
*** Where: do
*** Stack: cause-error  
```
The error isn't a perfect match, then as we're talking about more than one. The alternative is to show only the first one that's a problem. That leads to iterative debugging, if they have more than one. For now, I'd go with the above approach. We can improve it later if needed.

bitbegin:
need better solution

PR #3405, opened 2018-05-31
FIX: issue #3369 (accessing series over the board)

haolloyin:
Partial fix for edge case when `clear` or `remove` a series which points to another series's tail.

**`change` action also has the similar bug in such edge case, and I am working on it.**

@qtxie @bitbegin Please review this fix. Thanks!


qtxie:
Thanks. :+1:

PR #3414, opened 2018-06-09
gui-sys-call is only defined for Windows

x8x:
One less step to compile GUI Console by default for macOS

### DIFF
--- a/red.r
+++ b/red.r
@@ -522,7 +522,7 @@ redc: context [
 		]
 		exe: safe-to-local-file exe
 
-		either gui? [
+		either all [Windows? gui?] [
 			gui-sys-call exe any [all [file form-args file] ""]
 		][
 			if with [repend exe [" " form-args file]]


greggirwin:
I'll leave it to Doc or Qtxie to merge this, as currently it's guaranteed to be Windows, because that's set in `run-console` and used as the `gui?` arg (see line 696 and the `run-console` calls after line 800). What it may want to dispatch on, instead, is `gui-target`, which is set at line 457.

greggirwin:
Given the pressures of the team right now, @x8x, let's see if you and I can nail this down. Given my above comment, do you still want this merged? I'm not clear on the improvement it makes for Mac, but it sort of acts as documentation that the logic branch is for Windows.

x8x:
@greggirwin My assumption is, to get GUI console built by default on macOS, one needs to set `gui?` to `true` by changing line 696 from:
`gui?: Windows?									;-- use GUI console by default on Windows`
to:
`gui?: any [Windows? macOS?]									;-- use GUI console by default on Windows`

But if `guy?` is `true`, line 526 will return an error because `gui-sys-call` is defined only for Windows.
Line 456 shows that `gui?` could be defined also for macOS.
Unless there is a better way to set GUI Console by default for macOS.. :-)

x8x:
There is this a command line switch at line 727:
`| "--cli"						(gui?: no)`
which is useful only on Windows as GUI is default, there should then be a
`| "--gui"						(gui?: yes)`
switch for macOS, so this PR would still be valid.

greggirwin:
@x8x, I'm going to merge this. If they want to change it later, they can. Sound good?

x8x:
@greggirwin Sounds good, what about adding the `--gui` flag so that on macOS no source code need be changed to build the GUI Console? That would make it easier to test.

greggirwin:
Gah! Seems I forgot to merge this. Sorry for the delay.

Issue #3413, opened 2018-06-09
Missing DOC strings


endo64:
DOC strings are missing for below functions:

Note that I already fixed all of them with help of @greggirwin and will PR soon, opened this issue just to track it more easily.

Update, added doc strings here.

```
<<                   : Shift bits to the left
>>                   : Shift bits to the right
>>>                  : Shift bits to the right (unsigned)
_read-input          : Internal Use Only
_set-buffer-history  : Internal Use Only
as-color             : Combine R, G and B values into a tuple
as-ipv4              : Combine a, b, c and d values into a tuple
as-rgba              : Combine a, b, c and d values into a tuple
ask                  : Prompt the user for input
bind                 : Bind words to a context; returns rebound words
charset              : Shortcut for `make bitset!`
comment              : Consume but don't evaluate the next value
common-substr        : Internal Use Only
construct            : Makes a new object from an unevaluated spec; standard logic words are evaluated
context              : Makes a new object from an evaluated spec
create-dir           : Create the given directory
debug-info?          : Internal use only
dir                  : Display a single column directory listing, for the current dir if none is given
dir?                 : Returns TRUE if the value looks like a directory spec
do-actor             : Internal Use Only
do-file              : Internal Use Only
do-safe              : Internal Use Only
eval-set-path        : Internal Use Only
event?               : Returns true if the value is this type
exists?              : Returns TRUE if the file exists
find-flag?           : Checks a flag in a face object
get-current-dir      : Returns the platform’s current directory for the process
halt                 : Stops evaluation and returns to the input prompt
input                : Wait for console user input
last-lf?             : Internal Use Only
link-sub-to-parent   : Internal Use Only
link-tabs-to-parent  : Internal Use Only
ll                   : Display a single column directory listing, for the current dir if none is given
ls                   : Display a single column directory listing, for the current dir if none is given
make-face            : Make a face from a given style name or example face
normalize-dir        : Return the contents of the system clipboard
object               : Makes a new object from an evaluated spec
on-face-deep-change* : Internal use only
on-parse-event       : Standard parse/trace callback used by PARSE-TRACE
os-info              : Returns detailed operating system version information
parse                : Process a series using dialected grammar rules
pwd                  : Displays the active directory path (Print Working Dir)
quote                : Return but don't evaluate the next value
read-clipboard       : Return the contents of the system clipboard
red-complete-file    : Internal Use Only
red-complete-input   : Internal Use Only
red-complete-path    : Internal Use Only
[...]

PR #3416, opened 2018-06-09
Added missing doc strings, #3413

endo64:
Added missing doc strings, #3413

```
<<                   : Shift bits to the left
>>                   : Shift bits to the right
>>>                  : Shift bits to the right (unsigned)
_read-input          : Internal Use Only
_set-buffer-history  : Internal Use Only
as-color             : Combine R, G and B values into a tuple
as-ipv4              : Combine a, b, c and d values into a tuple
as-rgba              : Combine a, b, c and d values into a tuple
ask                  : Prompt the user for input
bind                 : Bind words to a context; returns rebound words
charset              : Shortcut for `make bitset!`
comment              : Consume but don't evaluate the next value
common-substr        : Internal Use Only
construct            : Makes a new object from an unevaluated spec; standard logic words are evaluated
context              : Makes a new object from an evaluated spec
create-dir           : Create the given directory
debug-info?          : Internal use only
dir                  : Display a single column directory listing, for the current dir if none is given
dir?                 : Returns TRUE if the value looks like a directory spec
do-actor             : Internal Use Only
do-file              : Internal Use Only
do-safe              : Internal Use Only
eval-set-path        : Internal Use Only
event?               : Returns true if the value is this type
exists?              : Returns TRUE if the file exists
find-flag?           : Checks a flag in a face object
get-current-dir      : Returns the platform’s current directory for the process
halt                 : Stops evaluation and returns to the input prompt
input                : Wait for console user input
last-lf?             : Internal Use Only
link-sub-to-parent   : Internal Use Only
link-tabs-to-parent  : Internal Use Only
ll                   : Display a single column directory listing, for the current dir if none is given
ls                   : Display a single column directory listing, for the current dir if none is given
make-face            : Make a face from a given style name or example face
normalize-dir        : Return the contents of the system clipboard
object               : Makes a new object from an evaluated spec
on-face-deep-change* : Internal use only
on-parse-event       : Standard parse/trace callback used by PARSE-TRACE
os-info              : Returns detailed operating system version information
parse                : Process a series using dialected grammar rules
pwd                  : Displays the active directory path (Print Working Dir)
quote                : Return but don't evaluate the next value
read-clipboard       : Return the contents of the system clipboard
red-complete-file    : Internal Use Only
red-complete-input   : Internal Use Only
red-complete-path    : Internal Use Only
routine              : Defines a function with a given Red spec and Red/System body
set-current-dir      : Sets the platform’s current process directory
set-flag             : Sets a flag in a face object
shift-left           : Shift bits to the left
shift-logical        : Shift bits to the right (unsigned)
shift-right          : Shift bits to the right
to-red-file          : Converts a local system file path to a Red file path
update-font-faces    : Internal Use Only
write-clipboard      : Write content to the system clipboard
write-stdout         : Write data to STDOUT
```

### DIFF
--- a/environment/console/engine.red
+++ b/environment/console/engine.red
@@ -310,10 +310,10 @@ expand: func [
 	probe expand-directives/clean blk
 ]
 
-ls:		func ['dir [any-type!]][list-dir :dir]
-ll:		func ['dir [any-type!]][list-dir/col :dir 1]
-pwd:	does [prin mold system/options/path]
-halt:	does [throw/name 'halt-request 'console]
+ls:		func ["Display a single column directory listing, for the current dir if none is given" 'dir [any-type!]][list-dir :dir]
+ll:		func ["Display a single column directory listing, for the current dir if none is given" 'dir [any-type!]][list-dir/col :dir 1]
+pwd:	func ["Displays the active directory path (Print Working Dir)"][prin mold system/options/path]
+halt:	func ["Stops evaluation and returns to the input prompt"][throw/name 'halt-request 'console]
 
 cd:	function [
 	"Changes the active directory path"
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -10,12 +10,12 @@ Red [
 	}
 ]
 
-routine: func [spec [block!] body [block!]][
+routine: func ["Defines a function with a given Red spec and Red/System body" spec [block!] body [block!]][
 	cause-error 'internal 'routines []
 ]
 
 also: func [
-	"Returns the first value, but also evaluates the second."
+	"Returns the first value, but also evaluates the second"
 	value1 [any-type!]
 	value2 [any-type!]
 ][
@@ -34,7 +34,7 @@ attempt: func [
 	]
 ]
 
-comment: func ['value][]
+comment: func ["Consume but don't evaluate the next value" 'value][]
 
 quit: func [
 	"Stops evaluation and exits the program"
@@ -75,6 +75,7 @@ probe: func [
 ]
 
 quote: func [
+	"Return but don't evaluate the next value"
 	:value
 ][
 	:value
@@ -260,6 +261,7 @@ math: function [
 ]
 
 charset: func [
+	"Shortcut for `make bitset!`"
 	spec [block! integer! char! string!]
 ][
 	make bitset! spec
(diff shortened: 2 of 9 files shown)


DideC:
`dir` and `ls` produce a "...**multi columns** directory listing...", while `ll` produce a "...single column..." one !!!

DideC:
In this list, the spec for `normalize-dir` is also wrong.

environment/functions.red
```diff
@@ -564,9 +568,10 @@ to-red-file: func [
 	dst
 ]
 
-dir?: func [file [file! url!]][#"/" = last file]
+dir?: func ["Returns TRUE if the value looks like a directory spec" file [file! url!]][#"/" = last file]
 
 normalize-dir: function [
+	"Return the contents of the system clipboard"
```
DideC: Not really what it does, isn't it ?

environment/console/engine.red
```diff
@@ -310,10 +310,10 @@ expand: func [
 	probe expand-directives/clean blk
 ]
 
-ls:		func ['dir [any-type!]][list-dir :dir]
-ll:		func ['dir [any-type!]][list-dir/col :dir 1]
-pwd:	does [prin mold system/options/path]
-halt:	does [throw/name 'halt-request 'console]
+ls:		func ["Display a single column directory listing, for the current dir if none is given" 'dir [any-type!]][list-dir :dir]
```
DideC: this is a "multi columns directory listing" !

greggirwin:
Thanks for reviewing @DideC ! Do you want to make a PR to fix them, or have @Endo do it?

DideC:
I let Endo do it ;-)
Don't know what to put for `normalize-dir`.

meijeru:
@DideC @endo64 what `normalize-dir` does:
`normalize-dir` will invoke `clean-path` and prefix the current working directory if the file path indicated by its argument (`word!` `path!` or `file!`) is _relative_, and add a trailing `/` if necessary
`clean-path` will normalize its `file!` argument to an absolute path where `.` and `..` are eliminated

greggirwin:
It was just a bad entry. Endo and I developed the list, and that one just ended up in the wrong slot. Normalize-dir will say "Returns an absolute directory spec" when fixed.

greggirwin:
`Ls` will be "Display a directory listing, for the current dir if none is given"

endo64:
@DideC Thanks for catching! 
@greggirwin I fixed them and made a PR #3419

DideC:
@endo64 @greggirwin  thanks for the averall fix and this correction.

PR #3428, opened 2018-06-20
Add: 'random/seed now' test in the date-test.red

hyzwhu:
This is a quick test for the #3252 .  
Because the code speed is very fast, and the ```now``` is not changed when the  ```random/seed now```  second called. I add a ```wait 1 ``` between the two calls in the test.

### DIFF
--- a/tests/source/units/date-test.red
+++ b/tests/source/units/date-test.red
@@ -851,6 +851,16 @@ Red [
 			26-Feb-6712/17:07:12-10:00
 		]
 
+	--test-- "misc-5"
+		res1: make block! 5
+		res2: make block! 5 
+		random/seed now 
+		res1: collect [loop 5 [keep random 10]]
+		wait 1
+		random/seed now 
+		res2: collect [loop 5 [keep random 10]]
+		--assert res1 <> res2 
+
 ===end-group===
 
 ~~~end-file~~~


PeterWAWood:
This test should not be added to the date tests. It is a test of random not date.

greggirwin:
I haven't look at `random` tests yet, but it's an action, so we should
decide where those tests live.

greggirwin:
Misc-3 and misc-4 tests test `random` already.

PeterWAWood:
I don't think it makes much sense to have tests of `random` inside the tests for `date`. In the future, when somebody is working on `random`, how will they now the tests are hidden amongst the date tests?

These tests should be moved to a `random-test.red` file.

PeterWAWood:
There is no need to use `now` and `wait` in these tests. Simple date literals in the same form as that returned by `now` would suffice.

greggirwin:
The question stands, I think, as to whether action tests should live in their own test file or with their host type.

PeterWAWood:
I think this issue shows that having action tests split across datatypes makes no sense in many cases;
```
>> random/seed now/time                                     ;; this line should be in date-test.red
>> collect [loop 5 [keep random 10]]                   ;; this line should be in integer-test.red
== [4 8 8 2 8]
```

greggirwin:
I think that just shows we don't have a plan for this yet. :^) We can just as easily argue that working on a datatype will be a lot more work if you have to go to a separate test file for every action that might be affected.

meijeru:
I agree with @greggirwin : the runtime is split by datatype and not by action, with all actions for one datatype inside the `%runtime/datatypes/<type>.reds` file (barring inheritance of course).

greggirwin:
@PeterWAWood, can we make a note to keep test location in mind, and make recommendations? With that, we can action this PR on its own.

greggirwin:
There is a conflict now, which needs to be addressed. @PeterWAWood what file do you want this test added to, and @hyzwhu will you make that change once we know?

greggirwin:
@PeterWAWood what do you want to do with this PR?

PeterWAWood:
Personally, I feel these tests should be in a random-test.red file but if it is easier to go ahead with this PR, please do so after resolving the file conflict.

greggirwin:
It looks like @endo64 submitted a test that obsoletes this.

PR #3434, opened 2018-06-22
Add tuple! support to `last`, including new tests.

greggirwin:


greggirwin:
That's how `pick` works for dates, so it's consistent with that. R2 used different indexes, with the year be 1, rather than the date part. Wrappers and named indexes seem the best solution. Not likely you'd want to use ordinals much for this, but at least it will be consistent.

hiiamboris:
In R3 and Red you could use `last` without caring if the series is at it's end:
```
>> last tail "abc"
== #"c"
```
Note that you're breaking this:
```
>> pick s: tail "abc" length? s
== none
```
So it's likely that some calls to `last s` will transform into `last head s` in practice. For instance, one might rely on `dir?` func working regardless of the index, which in turn relies on `last`. Or when you implement a stack you might wanna keep it's pointer at the tail and use `last` in the `pop` func without using an extra `back`.

On the other hand it will be consistent with current `take/last` implementation:
```
>> take/last tail "abc"
== none
```

There was also talk that `pick 0` should not return `none` but raise an error. :point_up: [June 14, 2018 12:35 PM](https://gitter.im/red/bugs?at=5b2236df1ee2d149ecc133ec) `0 = length? tail s`

greggirwin:
@hiiamboris, `last tail "abc"` returning a value seems wrong to me. In R2 it's an error:
```
>> last tail "abc"
** Script Error: Out of range or past end
** Near: last tail "abc"
```
Your arguments for it are pretty weak, as we can just say the opposite. Maybe others will weigh in. If you can find examples of code that break with this change, that's the best way to discuss it. i.e., why do they do that?

hiiamboris:
@greggirwin to be precise, I'm not advocating the current behavior, just drawing your attention to an edge case and related matters.

greggirwin:
Got it, thanks. :^)

greggirwin:
With Endo's endorsement, if you're on board @hiiamboris, I'll merge this. With luck, @PeterWAWood will chime in soon and can do the merge so there's no self-favoritism on my part. :^)

Issue #3460, opened 2018-07-06
DRAW/TRANSPARENT returns a TUPLE! rather than an IMAGE!


9214:
### Expected behavior
`draw/transparent` returns an image with transparent background.
### Actual behavior
`draw/transparent` returns a `255.255.255.0` `tuple!`. 
### Steps to reproduce the problem
```red
draw/transparent 0x0 []
```
### Red and platform version
```
Red 0.6.3 for Windows built 5-Jul-2018/20:38:08+05:00 commit #963ec55
```

PR #3461, opened 2018-07-06
FIX: issue #3460 (DRAW/TRANSPARENT returns a TUPLE! rather than an IM…

9214:


greggirwin:
The failing check for this is just appveyor timing out, yes? If so, and you can confirm a successful build on your end, I'll merge this.

9214:
@greggirwin I confirm a successful build on my machine. Tests timed out on R/S suite, so I don't think it's related to the change in Draw module.

greggirwin:
Thanks @9214!

Issue #3453, opened 2018-07-03
Case folding should use latest Unicode data


meijeru:
### Expected behavior
the table in `%runtime/case-folding.txt` should correspond to the latest Unicode version (currently 11)
### Actual behavior
the table uses version 7
### Steps to reproduce the problem
N/A
### Red and platform version
all

endo64:
The file name should be `%runtime/case-folding.reds`

meijeru:
Right you are!

bitbegin:
1. use https://github.com/bitbegin/generate-unicode-table to generate two tables in one file, one for to-uppercase, another for to-lowercase.

2. #include this file to case-folding.reds, directly use these two tables without sort.

meijeru:
I continue to think that pulling directly from `unicode.org`  is the most desirable solution, notwithstanding the anomalies that @bitbegin found. These should be reported to Unicode instead of being circumvented. As to reading directly, I dod not need `ftp://` for that, as my Gist shows. But if the current solution passes all tests, let it stand...

bitbegin:
@meijeru  this file CaseFolding.txt cannot give whole information for case folding direction.
For example, `00B5; C; 03BC; # MICRO SIGN` cannot tell us which is uppercase, which is lowercase. But we can get it from UnicodeData.txt.  Maybe we can also parse this file to get two tables if we have time. (I found the UnicodeData.txt from here https://www.unicode.org/Public/UCD/latest/ucd/UnicodeData.txt)

PR #3468, opened 2018-07-13
FIX: issue #3453 ( update case folding table from Unicode 7.0 to Unic…

bitbegin:
…ode 11.0)


meijeru:
For an automatic conversion of the Unicode table, see [this Gist](https://gist.github.com/meijeru/02bc24b9797af9ad6d1c770095ee1947)

greggirwin:
Thanks for the PR @bitbegin, though it seems to fail some tests.

@meijeru, if you've run your gist, do you know if the modified table passes all tests? 

Would be great to see how the two differ, so we know. And we should put tools like Rudolf's script in a special area

bitbegin:
The unicode case folding is not easy to implement, because of

* the unicode case folding file maybe not corrent.
For example:

00B5; C; 03BC; # MICRO SIGN
03C2; C; 03C3; # GREEK SMALL LETTER FINAL SIGMA

but this http://www.fileformat.info/info/unicode/char/b5/index.htm show, it's uppercase is 039c not 03bc,
and this http://www.fileformat.info/info/unicode/char/03c2/index.htm show, it's uppercase is 03a3 not 03c3.

* for now, Red's case conversation exists some limits, so some test case is not correct.

Issue #3444, opened 2018-06-24
User errors receive code 600 instead of 800


meijeru:
### Expected behavior

`e: make error! "abc"` should have `e/code = 800` since `system/catalog/errors/user/code = 800`

### Actual behavior
```
>> probe make error! "abc"
make error! [
    code: 600
    type: 'user
    ....
```
### Steps to reproduce the problem
execute above code
### Red and platform version
```
presumably all, tested on W10
```

nedzadarek:
I can confirm on Win 8.1 64b `Red for Windows version 0.6.3 built 26-Mar-2018/1:14:22+02:00` and `cli` `Red 0.6.3 for Windows built 18-Jun-2018/15:11:17+02:00 commit #6753157`.

bitbegin:
good catch!

greggirwin:
With that merged, it looks fixed to me. Thanks @bitbegin !

PR #3475, opened 2018-07-18
FIX: issue #3444 (error code need be continuous)

bitbegin:


greggirwin:
The jump in error numbers from 500 to 800 was the problem for #3444. Some logic must expect all "slots" to be filled, when indexing and producing an error code.

greggirwin:
Red chose compatible codes, but doesn't have `command` errors yet, and Rebol had 700 as reserved itself.

PR #3476, opened 2018-07-18
High res time query on Windows

hiiamboris:
Current implementation of `now/precise/time` uses `GetSystemTime` winapi call, which is limited to 1ms. I replace it with 2 calls: `GetSystemTimeAsFileTime` and `FileTimeToSystemTime`. Both are supported by WinXP.

From FileTime I can get not just milliseconds, but 100ns-rounded data. Tests/performance are unaffected.

Since Windows timer is still capped to 1ms period, the time reported by `now/precise/time` will increment by roughly 1ms if called frequently:
```
>> loop 10 [probe now/time/precise loop 3000 ['do 'nothing]]
23:38:40.548467
23:38:40.549467
23:38:40.549467
23:38:40.549467
23:38:40.550467
23:38:40.550467
23:38:40.550467
23:38:40.551467
23:38:40.551467
23:38:40.551467
```
However when called less frequently, the microsecond part changes, showing that the OS timer period is not precisely 1ms but bigger than that:
```
>> loop 10 [probe now/time/precise wait 1]
23:34:02.414616
23:34:03.414673
23:34:04.41473
23:34:05.414787
23:34:06.414844
23:34:07.414902
23:34:08.414959
23:34:09.415016
23:34:10.415073
23:34:11.41513
```

100-ns precision at best:
```
>> t: now/time/precise
== 23:34:44.390016
>> t/second
== 44.39001639999333        ; "double" for 44.390'016'400
```

I expect this to be a bit better suited for profiling tasks and might yield better timer query resolution once the OS supports it.

### DIFF
--- a/runtime/platform/win32.reds
+++ b/runtime/platform/win32.reds
@@ -143,6 +143,11 @@ platform: context [
 		SuppressExternalCodecs		[integer!]
 	]
 
+	tagFILETIME: alias struct! [
+		dwLowDateTime	[integer!]
+		dwHighDateTime	[integer!]
+	]
+
 	tagSYSTEMTIME: alias struct! [
 		year-month	[integer!]
 		week-day	[integer!]
@@ -255,6 +260,14 @@ platform: context [
 			FreeEnvironmentStrings: "FreeEnvironmentStringsW" [
 				env			[c-string!]
 				return:		[logic!]
+			]
+			FileTimeToSystemTime: "FileTimeToSystemTime" [
+				filetime	[tagFILETIME]
+				systemtime	[tagSYSTEMTIME]
+				return:		[integer!]
+			]
+			GetSystemTimeAsFileTime: "GetSystemTimeAsFileTime" [
+				time			[tagFILETIME]
 			]
 			GetSystemTime: "GetSystemTime" [
 				time			[tagSYSTEMTIME]
@@ -481,20 +494,24 @@ platform: context [
 		return:  [float!]
 		/local
 			tm	[tagSYSTEMTIME value]
+			ftime	[tagFILETIME value]
 			h		[integer!]
 			m		[integer!]
 			sec		[integer!]
-			milli	[integer!]
+			nano	[integer!]
 			t		[float!]
 			mi		[float!]
 	][
-		GetSystemTime tm
+		GetSystemTimeAsFileTime ftime
+		FileTimeToSystemTime ftime tm
 		h: tm/hour-minute and FFFFh
 		m: tm/hour-minute >>> 16
 		sec: tm/second and FFFFh
-		milli: either precise? [tm/second >>> 16][0]
-		mi: as float! milli
-		mi: mi / 1000.0
+		nano: either precise? [
+			ftime/dwLowDateTime % 10'000'000 * 100
+		][0]
+		mi: as float! nano
+		mi: mi / 1e+9
 		t: as-float h * 3600 + (m * 60) + sec
 		t: t + mi
 		t
--- a/runtime/platform/win32.reds
+++ b/runtime/platform/win32.reds
@@ -498,7 +498,13 @@ platform: context [
 			h		[integer!]
 			m		[integer!]
 			sec		[integer!]
+			ms-int 	[integer!]
+			bits0-31 	[integer!]
+			bits32-47 	[integer!]
+			bits48-63 	[integer!]
 			nano	[integer!]
+			hi		[integer!]
+			n 		[integer!]
 			t		[float!]
 			mi		[float!]
 	][
@@ -508,9 +514,20 @@ platform: context [
 		m: tm/hour-minute >>> 16
 		sec: tm/second and FFFFh
 		nano: either precise? [
-			ftime/dwLowDateTime % 10'000'000 * 100
+			ms-int: tm/second >>> 16
+			; let x0 = x1 + x2<<32 + x3<<48 (x2-x3 are 2-byte parts, to avoid overflow when multiplied by 1e4)
+			; then x0%n = (x1%n + (x2 * 1<<32%n) + (x3 * 1<<48%n)) % n
+			hi: ftime/dwHighDateTime
+			bits0-31: ftime/dwLowDateTime
+			if bits0-31 < 0 [hi: hi + 1] 		;-- `%` treats bits0-31 as signed, while it is really not, have to work around
+			bits32-47: hi and FFFFh
+			bits48-63: hi >>> 16 and FFFFh
+			n: 10'000
+			; overflow check: 9999 + (65535 * 8000) = 524'289'999
+			nano: (bits32-47 * 7'296) + (bits48-63 * 0'656) // n + (bits0-31 % n) 	;-- raw part, in 100ns units
+			nano * 100 + (ms-int * 1'000'000)
 		][0]
-		mi: as float! nano
+		mi: as-float nano
 		mi: mi / 1e+9
 		t: as-float h * 3600 + (m * 60) + sec
 		t: t + mi


PeterWAWood:
This PR is missing three things:
1. Linux, macOS and BSD implementations
2. Any agreement that `time/precise` should provide 100ns precision
3. Some minimal testing such as a the value from a subsequent call returning a larger number than a previous one.

greggirwin:
If we want max profiling precision, shouldn't we use QueryPerformanceCounter? That's not for `time` though. Modern Windows have GetSystemTimePreciseAsFileTime, but it doesn't go far back. 

`now/precise` wasn't intended as a profiling tool, so lack of precision isn't critical there. In most cases, I think having a fixed precision is much better here (which means we'd be rounding anyway) for normal time purposes.

We should certainly look at the profiling side, but separately.

hiiamboris:
Thank you both for reviewing it!
@PeterWAWood to clarify the point of this PR: enhance the precision on Windows with **minimum** effort and disturbance. Why not?
1. on POSIX systems `now/precise` already has 1 microsecond precision https://github.com/red/red/blob/master/runtime/platform/POSIX.reds#L494 so it's a windows-only issue
2. 100ns is just what that function provides. I don't discard any data, no reason.
3. Can do if it's agreed that the PR is welcome.

@greggirwin I agree, but as MS puts it
> QPC is independent of and isn't synchronized to any external time reference. To retrieve time stamps that can be synchronized to an external time reference, such as, Coordinated Universal Time (UTC) for use in high-resolution time-of-day measurements, use GetSystemTimePreciseAsFileTime.

And the latter starts with Windows 8. So using QPC in `now/precise` is out of question I guess. It will be a separate interface, and hopefully *monotonic* :)

hiiamboris:
By the way, my testing indicates that it's possible to use `NtSetTimerResolution` to tune the clock period to 0.5ms. Not that every Red process should do it for itself, but if I decide to do that for some program then with this PR I get 0.5ms intervals reported with `time/precise` out of the box.

greggirwin:
@hiiamboris if we agree that QPC, et al is the way to go for profiling, and it's also not a fit for `time!`, is there still value in this PR? That is, do we need more precision for what is roughly "wall clock" time? Next question: how do we align time and QPC so there's not a big mismatch in using them together?

hiiamboris:
> if we agree that QPC, et al is the way to go for profiling, and it's also not a fit for time!, is there still value in this PR? That is, do we need more precision for what is roughly "wall clock" time?

@greggirwin well, idk, do we *need* a 1ms precision? ;) We can profile with 1sec precision, just increase the number of test runs. The value is that it gives you all the precision info that OS makes available. On *nix it's 1us, on Windows - 500-1000us. *Why lose the information that is given?* Why introduce a rounding error? Worst case this PR is the same as without it, best case is 2+ times better. For now. As for the future, maybe W11 will have 1us precision timers, who knows?

> Next question: how do we align time and QPC so there's not a big mismatch in using them together?

I don't think we should. People made a whole project out of it: http://www.windowstimestamp.com/description . For profiling we need something fast, monotonic and stable. QPC is all that. It has some drift, so what? Profiler won't notice that. See "Fig. 4.4.3. Windows 7: Continuous adjustment interrupted by a three minutes drift phase." for the actual drift values. 3ms per 3 minutes? `3e-3 / 180 = 0.0016%` No big deal.

On the other hand, see "Fig. 4.4.1: Drift and autoadjust on a Windows 8 system.." on how system clock auto-adjustment behaves. This won't do profiling any good. But it can certainly help in situations where multiple systems have to communicate in time-dependent manner, over a network or what not.

hiiamboris:
@greggirwin last time I checked it was microseconds on ubuntu

greggirwin:
> Why lose the information that is given? Why introduce a rounding error? 

For the same reason we say "I'll meet you at 5:00", and why `now` only returns to the second. There's a point of diminishing returns, and a cognitive overhead to more precision. 1 second is very coarse, if you're doing something like logging, so my question is how often there is value in μs vs ms. Ms is *really* precise for almost any human need, IMO. It may be, though, that in the modern world of distributed computing, more precision is useful. Then we have to ask if that holds true for physical time, and not just logical clocks, ala Lamport.

giesse:
I think that there is a reason why `/precise` is a refinement, and I think when people use it we should give the most accurate value we have available. I don't think there is any downside to being "too precise". For normal uses people won't use `/precise`.

For things like `delta-time`, I'm all for using something different and even more accurate in the short term (but not good enough for time-of-day in general), eg. see http://www.rebol.net/r3blogs/0293.html

The more accurate the time source the less iterations we need to profile things. Additionally, a built in profiler that accumulates time spent in each function will have to use something like that since looping is not an option.

greggirwin:
But is `/precise` even the mechanism we want to use for profiling (eventually)? I'm all for a `stats/timer` being as precise as possible, but as a semi-normal user, my instinct is that those extra numbers on my eyeballs will all add up. Logs are my standard use case. Of course, we can round for that, as desired.

giesse:
I'm not saying we should use `/precise` for profiling - it's just what we happen to have right now.

My logic is simply that I don't see the reason in rounding `/precise` and making life harder for the few people that need the higher accuracy for some reason (say, you're writing a NTP server, or something like Google's Spanner DB). If you just need the time, don't use `/precise`, if you need more, use `/precise` and then format the result however works for you (like in your logging example - in fact in most cases you won't even need `/precise` for logging).

PeterWAWood:
I think the more important question is do we want to have the same behaviour across multiple platforms or different behaviour? 

We currently have inconsistent behaviour, microsecond precision on Windows and millisecond precision on other platforms. I believe it would be better to have consistent behaviour.

giesse:
@PeterWAWood it's the other way around. On Windows this returns 0:

```
>> s: now/precise difference now/precise s
== 0:00:02.4577e-5
```

(Not that the above is really accurate, but clearly the timer is incrementing.)

In a more readable format:

```
>> s: now/precise format-time difference now/precise s
== "25μs"
>> s: now/precise format-time difference now/precise s
== "25μs"
>> s: now/precise format-time difference now/precise s
== "16μs"
>> s: now/precise format-time difference now/precise s
== "49μs"
>> s: now/precise format-time difference now/precise s
== "16μs"
```

Wouldn't it make sense to bring Windows to the same accuracy of the other platforms?

greggirwin:
Thanks for all the consideration and arguments everyone. It sounds like this change has support, so if you add some tests @hiiamboris, I'll merge it.

greggirwin:
Also, I did confirm that ISO8601 doesn't impose a ms limit in the standard:

> There is no limit on the number of decimal places for the decimal fraction. However, the number of decimal places needs to be agreed to by the communicating parties.

PeterWAWood:
@giesse Yes. I would be happy with `/precise` providing the same precision on all platforms.

I believe in the principle that, as much as is possible, Red should have consistent behaviour cross-platform. I feel this is especially true at this stage whilst we don't have modules. Once we do, I would support putting platform specific behaviour in OS specific modules.

qtxie:
With `/precise`, we have millisecond precision on Windows and microsecond precision on other platforms for now. 
This PR provide 100ns precision on Windows. I think it would be better to make it microsecond precision, so we have consistent behaviour cross-platform.

greggirwin:
@qtxie +1. 

@hiiamboris if you make that change, add tests, and nobody complains, we can merge this.

runtime/platform/win32.reds
```diff
@@ -481,20 +494,24 @@ platform: context [
...
+		GetSystemTimeAsFileTime ftime
+		FileTimeToSystemTime ftime tm
 		h: tm/hour-minute and FFFFh
 		m: tm/hour-minute >>> 16
 		sec: tm/second and FFFFh
-		milli: either precise? [tm/second >>> 16][0]
-		mi: as float! milli
-		mi: mi / 1000.0
+		nano: either precise? [
+			ftime/dwLowDateTime % 10'000'000 * 100
```
qtxie: This is not correct. The `ftime` is a 64bit integer, you cannot just use half of it to do the calculation.

hiiamboris: Indeed! Thanks!

hiiamboris:
I incorporated a 64-bit modulo N // 10000 there. Sorry for the previous overlook.
I didn't make it into a separate func, so can't write tests for it. But you can test it in the Red console as I did:
```
mod10k: function [hi lo] [
	if lo < 0 [hi: hi + 1]
	bits32-47: hi and FFFFh
	bits48-63: hi >>> 16 and FFFFh
	n: 10'000
	(bits32-47 * 7'296) + (bits48-63 * 0'656) // n + (lo % n)
]

loop 100 [n: random 7FFFFFFFh  ok: n // 10000 = r: mod10k 0 n  print [ok n r]]
repeat i 30 [ok: (n: 2 ** i - 1) // 10000 = r: mod10k 0 (n2: FFFFFFFFh >>> (32 - i))  print [ok n n2 r]]
ok: (n: 2 ** 31 - 1) // 10000 = r: mod10k 0 7FFFFFFFh  print [ok n r]
ok: (n: 2 ** 32 - 1) // 10000 = r: mod10k 0 -1  print [ok n r]
repeat i 30 [ok: (n: 2 ** (32 + i) - 1) // 10000 = mod10k n2: 1 << i - 1 -1  print [ok n 2 ** 32 * (n2 + 1) - 1 r]]
```

What other tests can I possibly make? I can't test it against an unknown input (`GetSystemTimeAsFileTime`). Best I could do is test the additivity of time intervals:
```
>> s: 0:0:0  t: now/time/precise  loop 1000 [wait 1e-3 t': now/time/precise  s: t' - t + s  t: t']  ? s
S is a time! value: 0:00:01.000057
```
But I don't think including this into test suite is a good idea, since it's platform dependent and is expected to break on heavy load. If you wish I might set the timer (via WinAPI) to 500ns (best resolution) and show that the interval will be just that.

greggirwin:
@PeterWAWood do you have thoughts on how we should doc things that are hard to test, or require special testing frameworks?

PeterWAWood:
@greggirwin This needs to be tested by mocking the GetSystemTime and/or  GetLocalTime and/or GetTimeZoneInformation functions in Win32.reds (and their equivalents in other operating systems.)

It will need to be run separately from the current tests as it will require building a custom runtime.

greggirwin:
Looks like the appveyor build timed out on this, but otherwise can we merge this @qtxie? I think we have to live with the outstanding test issues for the time being.

qtxie:
@greggirwin So now we have 100-ns precision on Windows and microsecond precision on other platforms. I think it's fine to merge it.

Issue #3338, opened 2018-04-17
unset crash console


planetsizecpu:
### Expected behavior
If a word can't be unset issue an error

### Actual behavior
hang for a moment and crash console

### Steps to reproduce the problem
unset word [block]

### Red and platform version
Red 0.6.3 for Windows built 16-Apr-2018/11:21:12+02:00 commit #7c6d67f

```
>> about
Red 0.6.3 for Windows built 16-Apr-2018/11:21:12+02:00 commit #7c6d67f
a: [1]
unset a
```

meijeru:
It will also happen with `unset [1]`.

planetsizecpu:
Instead  `unset 'a ` works properly, as usual.

meijeru:
Yes, that is because in `unset a`, there is an evaluation of `a`! The moral is: probably anything other than a `word!`  value leads to a crash.

9214:
I can reproduce this with `unset` and `block!` argument containing at least one non-`any-word!` value.

meijeru:
Conclusion: anything other than a word or block of words leads to a crash.

PR #3480, opened 2018-07-22
FIX: issue #3338 (not allow non-word to be unset)

bitbegin:


tests/source/units/all-tests.txt
```diff
@@ -30,6 +30,7 @@ integer-test.red
 float-test.red
 same-test.red
 words-of-test.red
+unset.red
```
greggirwin: Suggest adding "-test" to the filename, before the suffix, to match other files.

greggirwin:
Thanks!

Issue #3489, opened 2018-08-04
Inconsistency between reduce and compose on any-function! values


meijeru:
### Expected behavior

`reduce` and `compose` should behave equally on non-block values
### Actual behavior
`reduce` and `compose` may both be applied to a non-block argument (BTW one could ask oneself why allow this at all)
for arguments *not* of type `any-function!` they just yield the result of evaluation
BUT `reduce :add` gives the `action!` value, `compose :add` gives an error, since it tries to *apply* the action
### Steps to reproduce the problem
execute above examples (with any native, action, function or routine -- `compose <op>` gives: not yet implemented)

### Red and platform version
all

greggirwin:
I have this change made for function types, but haven't found `compose` tests to update yet. In addition, `Compose` evaluates in Red, but not in Rebol. That seems more consistent with `reduce` in Red, so I'll leave it as is.
```
>> compose quote (1 + 1)
== 2
>> compose 'system/words
== make object! [
    datatype!: datatype!
    ...
```
Rebol:
```
>> compose quote (1 + 1)
== (1 + 1)
>> compose 'system/catalog
== system/catalog
```
`Reduce` checks for function types and does not evaluate them.

greggirwin:
Found the tests in %evaluation-test.red.

Issue #3487, opened 2018-07-29
FEAT: Make `help-string` available without console


greggirwin:
### Expected behavior

You should be able to use `help-string`, if not `help`, in non-console apps. If there is no console object, output should not wrap.

### Actual behavior

`help-string` crashes, because it depends on the console object existing to determine when to wrap output. 

### Steps to reproduce the problem

e.g. `print [help-string "-thru"]`

### Red and platform version
```
-----------RED & PLATFORM VERSION-----------
Should be the same on all.

RED: [ branch: "master" tag: #v0.6.0 ahead: 3561 date: 24-Jul-2018/23:29:28 commit: #f77e751ad310318172fd5b714c6f41c4b7004e9b ]
PLATFORM: [ name: "Windows 10" OS: 'Windows arch: 'x86-64 version: 10.0.0 build: 17134 ]
--------------------------------------------
```

greggirwin:
Or maybe it should format just as the console would, truncating if needed, at a configurable width in characters.

greggirwin:
I have this change made, but won't push a PR until I have tests, if we have a solution for that.

PeterWAWood:
@greggirwin It should work if you include %help.reds in a #system block and then define a routine! to access it. 

I hope that it won't require the `-r` option to compile.

greggirwin:
Sorry, should have been %help.red, not %.reds. Just wasn't sure if #include-ing  was how you wanted things done. Thanks.

PR #3490, opened 2018-08-04
FIX/FEAT: #3487 #3489 help-string and compose

greggirwin:
Apologies for unrelated commits in the PR.

### DIFF
--- a/environment/console/help.red
+++ b/environment/console/help.red
@@ -3,13 +3,12 @@ Red [
 	Author:  ["Gregg Irwin" "Oldes"]
 	File: 	 %help.red
 	Tabs:	 4
-	Rights:  "Copyright (C) 2013-2017 All Mankind. All rights reserved."
+	Rights:  "Copyright (C) 2013-2018 Red Foundation. All rights reserved."
 	License: {
 		Distributed under the Boost Software License, Version 1.0.
 		See https://github.com/dockimbel/Red/blob/master/BSL-License.txt
 	}
 	Notes: {
-		TBD: Emit to output buffer so help can be returned as a string.
 		TBD: Determine what useful funcs to export from help-ctx.
 	}
 ]
@@ -24,6 +23,7 @@ help-ctx: context [
 	RT_MARGIN: 5			; How close we can get to the right console margin before we trim
 	DENT_1: "    "			; So CLI and GUI consoles are consistent, WRT tab size
 	DENT_2: "        " 
+	NON_CONSOLE_SIZE: 120	; Where to truncate, if not running in a console
 	
 	;---------------------------------------------------------------------------
 	;-- Buffered output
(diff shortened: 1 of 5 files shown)


greggirwin:
Looks like quick-test wasn't enough. Need to see what's failing.

greggirwin:
@PeterWAWood I changed the #include to use an absolute path, which solved finding the file, but I don't know what the relative pathing should be for other includes to make that work normally. However, even with that in place, `help-string` is not seen during a %run-all.r pass. I haven't tracked that down yet, and may not spend more time on it now. I stubbed those tests out in the meantime.

greggirwin:
@PeterWAWood, any more thoughts on this?

PeterWAWood:
@greggirwin I'm not sure why you are having path problems with including help.red.

I added the following lines to `functions-test.red` on the `master` branch to check the `#include`:
```text
#include  %../../../environment/console/help.red
===start-group=== "help-ctx"
 	--test-- "help-ctx-1"
 		--assert help-ctx/DOC_SEP == "=>"
===end-group===
```
Here are the results:
```text
peter$ rebol -qws quick-test/run-test.r tests/source/environment/functions-test.red
Quick-Test v0.12.0
Running under REBOL 2.7.8.2.5
~~~started test~~~ functions
===group=== path-thru tests
--test-- path-thru test FAILED**************
~~~finished test~~~  functions
  Number of Tests Performed:      92
  Number of Assertions Performed: 134
  Number of Assertions Passed:    133
  Number of Assertions Failed:    1
****************TEST FAILURES****************
```
 How are you compiling the test when you get the paths issues?

greggirwin:
If I just run a single test file, it's fine. It's when I do `do %tests/run-all.r` in a REBOL console that I get the failure.
```
>> do/args %run-test.r to-rebol-file "c:\dev\greggirwin\red\tests\source\environment\functions-test.red"
Script: "Builds and Runs a single Red/System Tests" (none)
Script: {Simple testing framework for Red and Red/System programs} (none)
Script: "Untitled" (none)
Quick-Test v0.12.0
Running under REBOL 2.7.8.3.1
~~~started test~~~ functions
~~~finished test~~~  functions
  Number of Tests Performed:      102
  Number of Assertions Performed: 146
  Number of Assertions Passed:    146
```
```
>> cdr "C:\dev\greggirwin\red\"
== %/C/dev/greggirwin/red/
>> do %tests/run-all.r
Script: "Builds and Runs the Red Tests" (none)
checking to see if interpreter test files need generating
1 files were generated
Quick-Test v 0.12.0
REBOL  2.7.8.3.1
This test started at 5-Aug-2018/13:19:33.71-6:00
Red Test Suite

ok - lexer...........................................29 / 29
ok - preprocessor tests..............................22 / 22
ok - run-all-comp1.red.............................4039 / 4039
ok - run-all-comp2.red.............................6156 / 6156
ok - Red print........................................9 / 9
ok - Red run time errors..............................8 / 8
ok - Red compile errors...............................1 / 1

** - Red Test Suite*******************************10356 / 10357 **
       in 0:01:57.221

The test finished at 5-Aug-2018/13:21:30.931-6:00
The test output was logged to /C/dev/greggirwin/red/quick-test/quick-test.log
hit enter to finish

== 1
```

PeterWAWood:
@greggirwin What did the quick-test.log show. The 0/1 suggest that run-all-interp.red either did not compile or crashed (segfault etc.). 

I suspect the problem is that the `make-run-all.r` script is not copying the include of help.reds. None of the other tests include other source files.

For me, the best solution would be to put the tests of `help.red` into a `help-test.red` file.

greggirwin:
It said it couldn't find the include file, with the relative path.

PeterWAWood:
@greggirwin I now understand the issue, the relative path is correct when the program is compiled but when the test is run from the quick-test/runnable directory, the runtime will resolve the relative path differently.

I could come up with a fix to the code that generates the interpreter tests so that any #include, if place directly below the file header, is processed by the compiler rather than the interpreter. 

It is a problem of having code that is going to be both compiled and interpreted. I don't think that there is an easy answer until we have modules.

greggirwin:
Moved (stubbed) tests, and CI didn't time out. @PeterWAWood OK to merge?

PeterWAWood:
@greggirwin Looks good to me

PR #3507, opened 2018-08-14
FIX: build/git.r file looks modified after each build on Windows

endo64:
git.r file looks modified after each build on Windows because of line endings difference. Might be possible to ignore by changing git settings, but better to keep the original line endings.

### DIFF
--- a/build/build.r
+++ b/build/build.r
@@ -91,7 +91,7 @@ if Windows? [
 ]
 
 ;-- Restore git file
-attempt [write git-file "none^/"]								;-- tests require it!
+attempt [write/binary git-file "none^/"]								;-- tests require it!
 
 ;-- Remove temporary files
 attempt [delete ts-file]


greggirwin:
It seems that this is tied to git settings on a user's system, which may or may not alter line endings. Ah, the joy. @dockimbel doesn't want to use a %.gitignore, which is fine. So how can we work around this?

I'm on Windows with autoCrLf on, and every time I build, I have to revert this file when I go to submit a PR. If it's not just me, that affects people's ability to contribute easily.

x8x:
Speaking with @endo64, it looks like there is no real default for `autoCRLF` on Windows as it depends on which sauce of git one installs.
The best solution we came up is probably to add a `.gitattributes` file at the root of the repo, with `* text=auto` and revert this PR.
This will enforce autoCRLF for the repo also for users that have it disabled globally.
@endo64 can you confirm this works for you?

endo64:
@x8x @greggirwin `.gitattributes` is a good solution for line ending problems generally. However in our case, `git.r` file is changed by `build.r` so whatever solution we apply `git.r` would look modified.

I tested with different `core.autocrlf` settings, but `git.r` file always look modified on Windows after `build.r` executed.

However, why do we need a newline in `git.r` file anyway? I just removed the newline from `git.r`, committed, changed `attempt [write git-file "none^/"]` like to `attempt [write git-file "none"]` (no newline at the end), after executing `build.r` the `git.r` file doesn't look modified anymore. Tested also compiling `red.exe` and executed all the tests as well.

So the best solution is:
1. Remove the newline from `git.r`,
2. Change `attempt [write git-file "none"]` line.
3. Commit and push these changes to the repo.
4. Optionally add `.gitattributes` file with the following content to prevent further line ending issues:

```
*.red  text
*.reds text
*.r    text
*.txt  text
*.md   text
*.yml  text
*.svg  text
*.java text
*.html text
*.css  text
*.js   text

*.png  binary
*.gif  binary
*.jpg  binary
*.jpeg binary
*.ico  binary
```

Here some references: https://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/ and https://help.github.com/articles/dealing-with-line-endings/

greggirwin:
I like the "remove newline" solution.

PR #3509, opened 2018-08-15
Toomasv patch 2

toomasv:
Added parse rules for string matching in `replace`.


greggirwin:
Later PR replaced this one.

PR #3510, opened 2018-08-15
Added parse rules for string `replace` and tests

toomasv:


environment/functions.red
```diff
@@ -203,15 +203,16 @@ replace: function [
 	series [series!] "The series to be modified"
 	pattern "Specific value or parse rule pattern to match"
 	value "New value, replaces pattern in the series"
-	/all  "Replace all occurrences, not just the first"
+	/all "Replace all occurrences, not just the first"
 	/deep "Replace pattern in all sub-lists as well"
+	/local p rule s e many? len pos
```
greggirwin: Since `function` is used, not all these locals need to be declared.

Issue #3511, opened 2018-08-16
Response of function ASK contains question


DVL333:
Response of function ASK contains not only entered text, but a question also.

### Steps to reproduce the problem
```
>> name: ask "Enter your name "
Enter your name Sam
== "Sam"
>> print name
Enter your name Sam
```

It's wrong. Must be only "Sam"

### Red and platform version
-----------RED & PLATFORM VERSION----------- 
RED: [ branch: "master" tag: #v0.6.3 ahead: 839 date: 16-Aug-2018/9:37:36 commit: #22cf5fe67414ae44a9d731df3a9f8285d64c4a85 ]
PLATFORM: [ name: "Windows 10" OS: 'Windows arch: 'x86-64 version: 10.0.0 build: 14393 ]
--------------------------------------------

9214:
?
```Red
>> name: ask "enter your name "
enter your name Sam
== "Sam"
>> name
== "Sam"
>> print name
Sam
>> about
Red 0.6.3 for Linux built 16-Aug-2018/14:00:15+05:00 commit #8cc5422
```

greggirwin:
@bitbegin was too fast! :^) Duped here on GUI console, but not CLI console. Fixed already. Thanks for the report @DVL333!

greggirwin:
I'm on simple-gc branch right now. @DVL333 please close this when you confirm the fix.

PR #3512, opened 2018-08-16
FIX: issue #3511 (line is a special arg, should return a copy of it)

bitbegin:


greggirwin:
Appveyor seems to have hung. This is a safe and small change, so I'll merge it. Travis passed.

qtxie:
It is a bit strange. Should not need to copy the line as `ask` works fine in the console engine.

qtxie:
If it's needed, better to use `move` to move the content to the head of the string, avoid making a new string.
As we have GC now, it is not a big issue.

qtxie:
@bitbegin It's not the right fix. The bug is in `print`, not in `ask`.
```
>> s: skip "enter your name Sam" 16
== "Sam"
>> s
== "Sam"
>> print s
enter your name Sam
```

bitbegin:
@qtxie i'll fix it

qtxie:
Why not just remove the `head`?

bitbegin:
https://github.com/red/red/blob/04f07dfc243d8e1bc2f363287704f21bf66fc63f/environment/console/GUI/core.red#L625  like here, the line at tail but include promote.

qtxie:
Then we should add a `head` there: `add-line head line`. It's not efficient and would be costly to copy the string each time we printing.

bitbegin:
@qtxie please review this https://github.com/red/red/pull/3514

qtxie:
Looks good to me. :+1

PR #3513, opened 2018-08-21
FIX: empty id field when make error will crash console

bitbegin:


greggirwin:
Looks like an appveyor timeout again, otherwise good.

Issue #3518, opened 2018-08-23
Ordinal number lookup for the error throw is not correct


gltewalt:
### Expected behavior
I expect `make error! 0` to return "no loop to break" 

In:
```
SYSTEM/CATALOG/ERRORS/THROW is an object! with the following words and values:
     code        integer!      0
     type        string!       "Throw Error"
     break       string!       "no loop to break"
     return      string!       "return or exit not in function"
     throw       block!        length: 2  ["no catch for throw:" :arg1]
     continue    string!       "no loop to continue"
     while-cond  string!       {BREAK/CONTINUE cannot be used in WHILE condition block}
```
### Actual behavior
>> make error! 0
*** Script Error: value out of range: 0
*** Where: make
*** Stack: 
### Steps to reproduce the problem
`make error! 0`
### Red and platform version
```
-----------RED & PLATFORM VERSION----------- 
RED: [ branch: "master" tag: #v0.6.3 ahead: 847 date: 21-Aug-2018/3:45:37 commit: #04f07dfc243d8e1bc2f363287704f21bf66fc63f ]
PLATFORM: [ name: "Windows 7 Service Pack 1" OS: 'Windows arch: 'x86-64 version: 6.1.1 build: 7601 ]
--------------------------------------------
```

greggirwin:
Maybe this in %error.reds / `make`?
```
				if any [
					int/value <= 0
					(sym + object/get-size errors) <= as red-value! w
				][
					fire [TO_ERROR(script out-of-range) spec]
				]
```
`<= 0` should allow 0?

PR #3519, opened 2018-08-23
FIX: issue #3518 (0 can be error code number)

bitbegin:


greggirwin:
Looks like my guess was too simplistic, as something failed in the build.

gltewalt:
int/value < 0
(sym + object/get-size errors) < as red-value! w

bitbegin:
@greggirwin ci pass
@gltewalt `(sym + object/get-size errors) <= as red-value! w` is right.  As get-size errors return 10 (error code number from 0 to 9), w be assigned with 10 (1000 / 100) will not be allowed.

bitbegin:
@greggirwin ci issue, @x8x has fixed
@gltewalt `(sym + object/get-size errors) <= as red-value! w` is right.

object/get-size errors return 10 (0-9), `int` = 900, this expression will return false, and `int` >= 1000 will retrun true, so this is right.

However, we can improve this to 
```
object/get-size errors <= int/value
```
, to simplize the code.

PR #3556, opened 2018-10-10
FEAT: use macro to automatically track number of datatypes in all code

moliad:
used a long and explicit macro name (```TYPE_TOTAL_COUNT```) so it doesn't get mixed up with a type name within code.
```
TYPE_COUNT    TYPE_MAX   
```
could look like data type names and not be as obvious in intent.


dockimbel:
Good improvement! 👍

greggirwin:
Agreed. Nice, simple improvement.

Issue #3517, opened 2018-08-22
Poking an element of an image is partially bounds-checked


meijeru:
### Expected behavior
Using an out-of-bounds index to set a component of an image should give an error
### Actual behavior
```
>> i: make image! 2x2
== make image! [2x2 #{FFFFFFFFFFFFFFFFFFFFFFFF}]
>> i/size
== 2x2
>> length? i
== 4
>> i/5: #{00000000}
== #{00000000}
>> poke i 5 #{00000000}
== #{00000000}
>> i/(2x3): #{00000000}
*** Script Error: value out of range: 2x3
*** Where: set-path
*** Stack:  
>> i/(3x2): #{00000000}
== #{00000000}
```
### Steps to reproduce the problem
Execute above code
### Red and platform version
```
-----------RED & PLATFORM VERSION----------- 
RED: [ branch: "master" tag: #v0.6.3 ahead: 845 date: 20-Aug-2018/0:39:55 commit: #318673b7ae1c33b5852087f80905f2293ce5bd3f ]
PLATFORM: [ name: "Windows 10" OS: 'Windows arch: 'x86-64 version: 10.0.0 build: 17134 ]
```

greggirwin:
Same issue as #3516 (off by one at image end)?

PR #3560, opened 2018-10-12
FIX: issue #3517(access image out of range)

bitbegin:


bitbegin:
As image! is only valid for windows and mac, and invalid for linux, so i can't add the test file to `all-tests.txt`

PR #3573, opened 2018-10-21
FEAT: Added case refinement to REPLACE function

endo64:
`case` refinement added to `replace` function. See https://github.com/red/REP/issues/18

```
>> replace/all/case/deep quote (a (A a (A a))) ['a] 'x
== (x (A x (A x)))
```

### DIFF
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -205,54 +205,59 @@ replace: function [
 	value "New value, replaces pattern in the series"
 	/all "Replace all occurrences, not just the first"
 	/deep "Replace pattern in all sub-lists as well"
-	/local p rule s e many? len pos
+	/case "Case-sensitive replacement"
+	/local p rule s e many? len pos do-parse do-find
 ][
-	if system/words/all [deep any-list? series][
-		pattern: to block! either word? pattern [to lit-word! pattern][pattern]
-		parse series rule: [
-			some [
-				s: pattern e: (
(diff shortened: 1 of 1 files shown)


greggirwin:
Have to think about this a bit. Starting notes:
- Code is clean, though need to make sure tabs are used to indent.
- Could use comments (in the original), since there are a lot of interactions by refinement and input type, and some subtlety. e.g. at a glance, why does `binary!` force `many?`?
- The new `do reduce` parts in the `while` tests will add quite a bit of overhead. 
- Most importantly, needs tests.

Palaing:
Is there a reason not to use `change/part` instead of `remove/part` followed by `insert` ?

endo64:
> Is there a reason not to use `change/part` instead of `remove/part` followed by `insert` ?

Yes, please see the comments and the changes by this PR https://github.com/red/red/pull/3134

greggirwin:
@endo64 let's not add `/tail` support yet. I'm not sure of its value. Maybe others will chime in. I understand the idea is for chaining, as with `insert`, but I think it's a bit confusing in `replace` and have never seen it used.

`Do reduce` could be replaced with a specializer/apply, but we don't have an official one of those yet. I'll go ahead and merge this, so it's there, then please write or have someone else write some tests and do another PR with comments (where we can note that `do reduce` should be done away with.

Thanks!

Issue #3590, opened 2018-11-16
parse-func-spec in %help.red should be local


meijeru:
**Describe the bug**
The definition of `parse-func-spec` (an auxiliary function for `help`) is visible in the global context.
**To Reproduce**
Call `source parse-func-spec` in the console
**Expected behavior**
The definition should not be visible.

PR #3591, opened 2018-11-16
FIX: #3590 parse-func-spec in %help.red should be local

greggirwin:


greggirwin:
Appveyor just timed out. No other errors.

Issue #3380, opened 2018-05-09
foreach should complain about the empty block of words


hiiamboris:
Since it's a probable indication of a coder's mistake? Otherwise one would use `forever`?

### Steps to reproduce the problem
`foreach [] [1] []`

### Expected behavior
R3:
```
>> foreach [] [1] []
** Script error: invalid argument: []
** Where: foreach
** Near: foreach [] [1] []
```

### Actual behavior
it deadlocks

### Red and platform version
```
-----------RED & PLATFORM VERSION----------- 
RED: [ branch: "master" tag: #v0.6.3 ahead: 688 date: 5-May-2018/11:50:50 commit: #9ce2aacce1509943ac11efa3b13457e8a66f03b4 ]
PLATFORM: [ name: "Windows 7 Service Pack 1" OS: 'Windows arch: 'x86-64 version: 6.1.1 build: 7601 ]
--------------------------------------------
```

PR #3611, opened 2018-11-24
3380

hiiamboris:
Fix for #3380:
- a sanity check against deadlocking in `foreach-next-block` so that it will return false in case the block of words is empty and thus the input cannot possibly be advanced
- a specific error when `foreach` 1st argument is empty `[]`

A couple of related tests

### DIFF
--- a/runtime/natives.reds
+++ b/runtime/natives.reds
@@ -297,6 +297,7 @@ natives: context [
 		
 		either TYPE_OF(value) = TYPE_BLOCK [
 			size: block/rs-length? as red-block! value
+			if 0 >= size [fire [TO_ERROR(script invalid-arg) value]]
 			
 			while [foreach-next-block size][			;-- foreach [..]
 				stack/reset
@@ -3002,7 +3003,7 @@ natives: context [
 		]
 		assert TYPE_OF(blk) = TYPE_BLOCK
 
-		result: loop? series
+		result: (loop? series) and (0 < size)
 		if result [
 			switch type [
 				TYPE_STRING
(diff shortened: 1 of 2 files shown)


greggirwin:
Thanks @hiiamboris! It looks good to me at a glance, but I'll let @qtxie give the final sign-off and merge.

hiiamboris:
Sure. The more eyes the better.

runtime/natives.reds
```diff
@@ -3002,7 +3003,7 @@ natives: context [
 		]
 		assert TYPE_OF(blk) = TYPE_BLOCK
 
-		result: loop? series
+		result: (loop? series) and (0 < size)
```
Oldes: I believe it would be more readable when you would use `size > 0` instead of `0 < size`. At least for my brain.

greggirwin: :+1:

meijeru: Such an idiom can be quite convenient when the right hand side is an expression with operators in it, and one wants to avoid parentheses (otherwise necessary due to left-association). However I agree that in this case it is unnecessary.

qtxie: `and` is bitwise operator, here we want `logical and`.

meijeru: `and` also works on `logic!` values, so it is not really wrong, but I agree `all` is more idiomatic.

hiiamboris: Thank you for being thorough :+1:

qtxie:
Thanks! :+1:

Issue #3607, opened 2018-11-22
[Regression] draw matrix transformations became a mess


hiiamboris:
**Describe the bug**
Copying/transforming images with `draw .. [image ..]` is no longer possible.

Here's a *test* image **i** 4x4 pixels:
![](https://i.gyazo.com/89636aa59dcbd80020f98455159177ef.png)

Let's embed it into a bigger (8x8) image **i'**, shifting by 1x1 from the left top corner:
In the previous version:
![](https://i.gyazo.com/4f627dbce1012781ce2212b2dd57884b.png)

In the current version:
![](https://i.gyazo.com/039ec45eaf1bd0dba94c252afda48dbc.png)

Rotated clockwise it's even messier:
![](https://i.gyazo.com/b370698214646a79450d8e3e4f6f708b.png)

Obviously something has to do with miscalculation of the total size of the image (5x5 instead of 4x4)

**To Reproduce**
We'll need a helper function to be able to see pixel details:
```
inspect: function [i] [
	? (also i': make image! i/size * 20x20
	x: y: repeat y i/size/y [
		repeat x i/size/x [
			p: as-pair x y
			draw i' compose [
				line-width 1 pen (i/:p) fill-pen (i/:p)
				reset-matrix translate (20x20 * (p - 1x1))
				box 0x0 19x19
			]
		]
	])
]
```

To draw the original image
`inspect i: draw 4x4 [anti-alias off matrix [0 1 -1 0 3 0] pen red box 0x0 3x3 pen blue line 0x0 3x0]`

To make the embedded image
`inspect i': draw 8x8 [anti-alias off matrix [1 0 0 1 1 1] image i]` (as is)
`inspect i': draw 8x8 [anti-alias off matrix [0 1 -1 0 5 1] image i]` (rotated clockwise by 6 hours)

**Platform version(please complete the following information):**
The regression happened somewhere between Oct 16 and Oct 30 of 2018.
```
Red 0.6.4 for Windows built 22-Nov-2018/3:40:38+03:00 commit #755eb94
```

hiiamboris:
The regression was caused by commit https://github.com/red/red/commit/febc80259d991abbc50334a94775fb8b84f03c8f that tried to "fix" the issue https://github.com/red/red/issues/3504

But the issue itself obviously resulted from the lack of attention or understanding of what one's doing, so the "fix" is wrong and should be rolled back.

[...]

PR #3622, opened 2018-11-30
prevent #3607-like regressions

hiiamboris:
A test file for draw image. Let me know if I'm checking for View availability in a wrong manner.

### DIFF
--- a/tests/source/units/all-tests.txt
+++ b/tests/source/units/all-tests.txt
@@ -47,6 +47,7 @@ recycle-test.red
 debase-test.red
 enbase-test.red
 try-test.red
+draw-test.red
 auto-tests/lexer-auto-test.red
 power-test.red
 ../environment/scalars-test.red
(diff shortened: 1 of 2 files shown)


qtxie:
@hiiamboris A better way is to use macros. 
```
#if find config/modules 'view [
    ;-- tests
]
```

hiiamboris:
@qtxie unfortunately the testing framework won't allow this:
```
source/units/auto-tests/run-all-comp2.red - compiler error


-=== Red Compiler 0.6.4 ===-

Compiling D:\devel\red\red-src\red\tests\source\units\auto-tests\run-all-comp2.red ...
...using libRedRT built on 26-Nov-2018/22:23:13
*** Preprocessor Error in #if
*** Script Error : find expected series argument of type: series object port bitset
*** Where: [#if]
*** Near:  [find config/modules 'view]
```

greggirwin:
@qtxie in cases where one CI path fails (e.g. Travis), but the other succeeds, do we want to hold off merging? Appveyor almost always fails due to timing out, so I merge if that's the case.

qtxie:
@greggirwin I also think it's safe to merge the PR in this case.

hiiamboris:
But it was broken :( I'll try to find the time tomorrow to fix it.

PR #3625, opened 2018-12-01
proper ceiling in size-text

hiiamboris:
This formula
```
delta: either dpi-factor = 100 [0][1]
....: size/width * 100 / dpi-factor + delta
``` 
relies on 2 assumptions:
- (size/width * 100) / dpi-factor never becomes an integer
- dpi-factor is always <= 100

Both don't look valid to me. I suggest just round the result to the nearest higher integer.


greggirwin:
Looks like the build failures were just appveyor timing out.

@qtxie, please merge if this looks good to you.

Issue #3502, opened 2018-08-12
CI:AppVeyor testing not completing due to reaching maximum 1 hour execution time


x8x:
We have an issue where AppVeyor CI randomly errors out with this message:
`Build execution time has reached the maximum allowed time for your plan (60 minutes).`
It doesn't seam to be an error within the testing framework and there is no common pattern of failure.
Something gets stuck while tests are running which prevent completion.
List of such events happening in the last year:

https://ci.appveyor.com/project/red/red/build/1.0.2205/messages
https://ci.appveyor.com/project/red/red/build/1.0.2199/messages
https://ci.appveyor.com/project/red/red/build/1.0.2183/messages
https://ci.appveyor.com/project/red/red/build/1.0.2177/messages
https://ci.appveyor.com/project/red/red/build/1.0.2152/messages
https://ci.appveyor.com/project/red/red/build/1.0.2135/messages
https://ci.appveyor.com/project/red/red/build/1.0.2130/messages
https://ci.appveyor.com/project/red/red/build/1.0.2063/messages
https://ci.appveyor.com/project/red/red/build/1.0.2056/messages
https://ci.appveyor.com/project/red/red/build/1.0.2035/messages
https://ci.appveyor.com/project/red/red/build/1.0.2016/messages
https://ci.appveyor.com/project/red/red/build/1.0.2015/messages
https://ci.appveyor.com/project/red/red/build/1.0.2011/messages
https://ci.appveyor.com/project/red/red/build/1.0.2000/messages
https://ci.appveyor.com/project/red/red/build/1.0.1994/messages
https://ci.appveyor.com/project/red/red/build/1.0.1986/messages
https://ci.appveyor.com/project/red/red/build/1.0.1968/messages
https://ci.appveyor.com/project/red/red/build/1.0.1966/messages
https://ci.appveyor.com/project/red/red/build/1.0.1963/messages
https://ci.appveyor.com/project/red/red/build/1.0.1950/messages
https://ci.appveyor.com/project/red/red/build/1.0.1947/messages
https://ci.appveyor.com/project/red/red/build/1.0.1944/messages
https://ci.appveyor.com/project/red/red/build/1.0.1925/messages
https://ci.appveyor.com/project/red/red/build/1.0.1923/messages
https://ci.appveyor.com/project/red/red/build/1.0.1907/messages
https://ci.appveyor.com/project/red/red/build/1.0.1902/messages
https://ci.appveyor.com/project/red/red/build/1.0.1890/messages
[...]

PR #2191, opened 2016-08-31
FEAT: Regression tests

rebolek:

PR #3638, opened 2018-12-07
2191 part 1

hiiamboris:
The first part of #2191 - Red tests for both interpreter and compiler.
There are 2 more parts: Red compiler tests and R/S compiler tests. They contain about the same number of regression tests, but I won't be able to include them until #3502 is addressed, as Red part always hangs and R/S part hangs in 80% cases.

Also included is a fix for `parse-test`: redefining `split` was a bad idea as it affects the other tests.


hiiamboris:
https://github.com/red/red/issues/3640 is likely the cause of Travis failing.

Issue #3636, opened 2018-12-06
attempt to SAVE a block with LOADed image inside of it results in access violation


9214:
**Describe the bug**
Title.

**To Reproduce**
Create a Red script with content as specified below, and put any image file in the same directory:
```red
Red []

save %image.red reduce [load %<your-image-file>]
``` 

Then try to execute that script.

```
*** Runtime Error 1: access violation
*** at: 742BA8C9h
```

**Expected behavior**

An `image.red` file is created with `image!` value as its content.

**Platform version (please complete the following information):**
```
Red 0.6.4 for Windows built 28-Nov-2018/6:34:42+05:00 commit #d463f89
```

bitbegin:
how you execute the image.red?

endo64:
@bitbegin I think @9214 means executing the file above not the `image.red`

@9214 It worked for me on W10 x64: `Red 0.6.4 for Windows built 5-Dec-2018/12:31:35+03:00 commit #90c8e83`

9214:
It's strange that others can't reproduce it. Still the case for me with the latest build.

bitbegin:
you can put call-stack here, using `-r -d` argument

9214:
Interesting, I just tried to reproduce it with my own Github profile pic taken from this chat, no crash.

endo64:
@9214 Can you share the image file you try?

9214:
Sure, I got 3 of them.
[buggy.zip](https://github.com/red/red/files/2658859/buggy.zip)

hiiamboris:
Crashes on W7 x64 with all these 3 pics.

bitbegin:
seams GC bug:

```
>> recycle/off
>> to binary! trim mold/only reduce [load %3.jpg]
```

bitbegin:
```
to binary! trim mold/only reduce [append/dup a: "" "a" 1100000]
```

9214:
> The only known leak is the bitmaps object allocated from the OS by `image!` values.

endo64:
@9214 Crashes pictures in buggy.zip. Works if `recycle/off`.

bitbegin:
@9214 @endo64 can you check if https://github.com/red/red/pull/3644 work for you? it works for me both:
`save %image.red reduce [load %1.png]` and `to binary! append/dup a: "" "a" 1100000` are all run ok

qtxie:
[...]

PR #3644, opened 2018-12-10
FIX: #3636 (avoid system/call issue)

bitbegin:


greggirwin:
@bitbegin can you provide at least some description about this fix? Appveyor just timed out, so with that I can merge this PR.

hiiamboris:
No way this may work. R/S is unaffected by the GC. Is it not?

PR #3646, opened 2018-12-10
3502

hiiamboris:
Fix for the notorious https://github.com/red/red/issues/3502

As to the cause - why the input from the pipe is not immediately available after the process just terminated, figure it probably how the buffering in Windows works. It terminates then writes the buffer into pipe, and we are able to peek at it before the flush. 

I can't simply use `until [.... remain: remain - written   remain <= 0]` there, as what if the pipe holds exactly 10000 bytes (the buffer size)? Then the next ReadFile will try to wait for a write on that pipe which will never occur.

### DIFF
--- a/lexer.r
+++ b/lexer.r
@@ -838,6 +838,7 @@ lexer: context [
 		parse/all/case copy/part s e [
 			any [
 				escaped-char   (insert tail new value)
+				| #"^^"
 				| s: filter e: (insert/part tail new s e)
 			]										;-- exit on matching " or }
 		]
--- a/modules/view/backends/macOS/gui.reds
+++ b/modules/view/backends/macOS/gui.reds
@@ -179,8 +179,8 @@ get-text-size: func [
 	y: system/cpu/edx								;-- string height on screen
 	rc: as NSRect! :x
 
-	size/width: as-integer rc/x
-	size/height: as-integer rc/y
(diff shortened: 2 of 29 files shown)


greggirwin:
NICE! I know Nenad will want the unrelated caret/ceiling commits omitted if possible. Git causes me grief this way too, but if you can remove those from the PR, I will merge this ASAP.

hiiamboris:
@greggirwin these are already merged into the master branch. I have no idea why git lists these.

greggirwin:
Ah, git.

endo64:
Those commits looks in `3502` branch in `hiimboris/red` and merged into red from there. But they are not in master branch `hiiamboris/red`. I think you need to pull first.

endo64:
Couldn't reach @greggirwin 's speed :)
But anyway, it is always a good practice to pull from the upstream/master into your master when your PRs merged into it, before making other changes in your fork.

hiiamboris:
@endo64 that's what I did when it added the "Merge branch 'master' of https://github.com/red/red into 3502" commit.

qtxie:
I usually use `git pull --rebase` to avoid the extra `Merge branch 'master'` commit.

hiiamboris:
@qtxie thanks! I'll keep that in mind.

PR #3648, opened 2018-12-11
3607

hiiamboris:
Hopefully fixes Travis failing...
I'm also including @bitbegin's image-test to be run where View is available.


hiiamboris:
This will succeed once https://github.com/red/red/issues/3649 is resolved. Let it lie for now.

PR #3657, opened 2018-12-12
FIX: minor retouching of issue template

9214:
Small fixes here and there: missing spaces, inconsistent capitalization and ellipsis, etc.


greggirwin:
Good catches. Thanks.

PR #3688, opened 2018-12-29
FEAT: support input-stdin and read-stdin to read stdin directly

bitbegin:


qtxie:
Those two functions are needed by the vscode plugin. They are temporary. I prefer to add a caution in the docstring. e.g.
```
input-stdin: routine [
    "Temporary function, internal use only"
]
```

PR #3694, opened 2018-12-31
More trigonometric docstrings fixes

hiiamboris:


greggirwin:
@hiiamboris, is there a reason you removed the doc string for the argument? What about making it something like "The cosine of the angle, from -1.0 to 1.0"?

hiiamboris:
@greggirwin inverse trigonometric functions' domains are not angles, but real segments [-1,1] or [-inf,+inf].

hiiamboris:
Sure, as long as it is not misleading. But then again, just "a number" won't do... What about "a real number [-1,1]" or "a real number in range [-1,1]"?

greggirwin:
Is "The cosine of the angle, from -1.0 to 1.0" misleading or incorrect in any way?

greggirwin:
Arccosine is the inverse of cosine, so it doesn't matter where the number comes from. The behavior is the same, yes? We could say *an* angle, as a tweak: "The cosine of an angle, from -1.0 to 1.0"

To me "real number" is not as helpful, as it provides no context. In this case, there is context for the argument.

hiiamboris:
But there is no context. It's just a number. That's the whole point.

greggirwin:
The context is `cosine`. It's not just a number, but a bounded number with a specific meaning in that context. Here's what Excel uses: "The cosine of the angle you want and must be from -1 to 1."

greggirwin:
Mathematica uses this for the func description itself: "gives the arc cosine `cos-1(z)`  of the complex number `z`."

hiiamboris:
I can't believe we're arguing here on a matter as trivial as trigonometric docstrings. To be honest, I didn't expect you to find a single occurrence of the wording where `acos` argument will be called a "cosine". But to my surprise you did find one, from Excel ☻
For the record, I still find Excel wording fundamentally wrong. But I guess I can see the point: uneducated people are more likely to look up these functions descriptions, and thus for them we should be writing it.

greggirwin:
:^)

greggirwin:
Merged this, but missed that we still need a doc string for arccosine's `value` arg. TBD.

Issue #3655, opened 2018-12-12
to-pair with a single percent value is accepted, but with a block of two values they may not be percent


meijeru:
**Describe the bug**
See the title - this is an inconsistecy

**To Reproduce**
>> to-pair 200%
== 2x2
>> to-pair [100% 200%]
*** Syntax Error: invalid construction spec: [100% 200%]
*** Where: to
>> to-pair [1 2]
== 1x2

**Expected behavior**
>> to-pair [100% 200%]
== 1x2

Or, disallow `to-pair 200%`

**Platform version(please complete the following information):**
```
Presumably all
```

greggirwin:
I seem to recall old chat about percent values and their relative vs absolute nature. Deserving of deeper thought.

meijeru:
More consistent all around to disallow percent here. May be just a question of specifying number! when it was still only integer! + float! ...

endo64:
I'm for disallowing percent for `make pair!` (`to` is same with `make` for pair), unless we will have pairs with float values.

meijeru:
This can now also be closed?

greggirwin:
Thanks for weighing in guys.

PR #3697, opened 2019-01-01
FIX: issue #3655

meijeru:


hiiamboris:
@meijeru tuple elements can be assigned a `none` value:
```
>> color: 255.255.255.0
== 255.255.255.0
>> color/4: none
== none
>> color
== 255.255.255
```
My `base-test` relies on this.

meijeru:
Currently, setting any `t/(i)` to `none` removes all elements from `max(4,i)` and sets all elements from `i` to `max(i,3)` inclusive to `0`.
```
>> t: 1.2.3.4
== 1.2.3.4
>> t/2: none
== none
>> t
== 1.0.0
```
I have no problem with that. Will update the PR.

meijeru:
BTW how do I remove the unwanted commit from the pull request?

hiiamboris:
By working on top of it, or making a new PR.

greggirwin:
> Currently, setting any t/(i) to none removes all elements from max(4,i) and sets all elements from i to max(i,3) inclusive to 0.

Wow. I don't I ever tried that. Works the same in R2. It's a way to `clear`  tuples, but should be doc'd. My first instinct was that tuples should support `clear`, but there's no current index, as they aren't series. So no `at`, `tail`, etc. We can write a helper to make this clear, but won't modify the tuple arg in place, as when done directly. Never a shortage of design questions with Red. 

```
clear-tuple: func [
	"Removes tuple values from index forward; returns truncated tuple"
	t [tuple!] "Tuple to truncate (not modified)"
	index [integer!] "1..12; if less than 3, parts (index..3) become 0"
][
	t/:index: none
	t
]
```

Issue #3691, opened 2018-12-31
Color disabled in rich-text data


toomasv:
**Describe the bug**

Whenever `strike` or `underline` appear immediately before color-tuple in low-level data, color is not rendered.

**To reproduce**

```
>> view [rt: rich-text data [s [red "bbb"]] do [probe rt/data]]
[
    1x3 strike 255.0.0
]
```
![image](https://user-images.githubusercontent.com/11960635/50554894-c84e9b80-0ccb-11e9-8a3e-f2ee1ac1ebfc.png)

Color is rendered if tuple appears before `strike` / `underline` style-word. Also, this does not affect `italic` and `bold` styles.

**Expected behavior**

Color should be rendered with `strike` and `underline` as it is rendered now with `italic` and `bold`.

**Platform version (please complete the following information)**
```
Red 0.6.4 for Windows built 30-Dec-2018/4:42:09+02:00 commit #6968dad
```

qtxie:
It's strange that the code below works:
```
 view [rt: rich-text data [s [red "bbb"]] do [rt/data/2: 255.0.0 rt/data/3: 'strike probe rt/data]]
```

qtxie:
The text color should always follow the range. The color following the `strike` is used to specify the color of the strike line, but this is feature is not implemented yet.

toomasv:
OK, then my partial fix was OK. But this does not concern `italic` and `bold`, which show correctly if color follows to these?

Then the full fix needs to insert color just after address when uniting styles in optimization phase?

toomasv:
Thanks for the detailed spec!

PR #3698, opened 2019-01-02
FIX: Unite colors in right order

toomasv:
Color-tuple should stay just after range when uniting styles, as explained by @qtxie in https://github.com/red/red/issues/3691

Also: Count colors in path (may be multiple colors in path, last is rendered), fixed new-line for `close-colors`.

### DIFF
--- a/modules/view/RTD.red
+++ b/modules/view/RTD.red
@@ -13,7 +13,8 @@ Red [
 context [
 	stack: make block! 10
 	color-stk: make block! 5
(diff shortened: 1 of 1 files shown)


greggirwin:
Thanks for your recent fixes in this area @toomasv. Many are not simple.

qtxie:
Perfect! Kodus to you. @toomasv :+1:

toomasv:
Thanks @greggirwin and @qtxie !

PR #3699, opened 2019-01-02
 arctangent2 radians/degrees support 

hiiamboris:
So I was asleep when my mind brought it up to me that there was a typo in the previous docstrings. And then @greggirwin wished to also modify docstrings of corresponding natives. And then I noticed that arctangent2 was definitely meant to support radians, yet never implemented:
- otherwise it is meaningless to have atan2 and arctangent2 doing the same thing under 2 names
- https://github.com/red/red/blob/c87d85e8e7fbfb6b753538971e3ddfa8345cae9c/runtime/natives.reds#L1702 (presence of a check for nonexistent argument)
- https://github.com/red/red/blob/c87d85e8e7fbfb6b753538971e3ddfa8345cae9c/environment/functions.red#L984 (passing of `1` into nowhere as only a single arg is accepted)

Then I thought what's the better time than now... ;)

### DIFF
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -964,7 +964,7 @@ asin: func [
 ]
 
 atan: func [
-	"Returns the trigonometric arctangent (in radians in range [-inf,+inf])"
+	"Returns the trigonometric arctangent (in radians in range [-pi/2,+pi/2])"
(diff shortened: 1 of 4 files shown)


greggirwin:
:^)

Issue #3659, opened 2018-12-12
treatment of unset value by reduce and compose is different


meijeru:
**Describe the bug**
`reduce` accepts `any-type!` i.e. including `unset!` and `compose` accepts only `default!` i.e. without `unset!`

**To reproduce**
>> type? reduce ()
== unset!
>> type? compose ()
*** Script Error: compose does not allow unset! for its value argument

**Expected behavior**
either `reduce` refuses `unset!` or `compose` accepts `unset!`

**Platform version (please complete the following information)**
```
Presumably all
```

greggirwin:
I don't know why `reduce` was widened to accept `unset!`. It should be easy enough to change `compose`'s spec to `[any-type!]`, as it looks like the evaluation would be the same.

dockimbel:
I think `compose` should only accept `block!` as its first argument, as does the docstring implies. 

In Red, `reduce` currently acts like `do` for non-block arguments. Whether we'll keep or not this behavior in 1.0 is not decided yet.

endo64:
Compose accepts any-value! in both R2 & R3, would it better to be compatible with R2 & R3? Though I don't think it would lead much inconsistency if we change it to block!

greggirwin:
I vote for restricting `compose` to blocks, unless there's a compelling argument (which I can't think of).

9214:
> a compelling argument

Function composition (no pun intended) and consistency. `reduce` acts like `do` for non-blocks - a feature that I used a couple of times in function chaning; same goes for `compose` - it acts like `do` for non-blocks. The only difference is `unset` treatment. Another argument is compatability, which @endo64 already mentioned.

And, on the contrary, I don't find "restrict `compose` to `block!` because docstring says so" argument compelling at all.

greggirwin:
> And, on the contrary, I don't find "restrict compose to block! because docstring says so" argument compelling at all.

This is where we need to think in terms of intent, and humans. Let's go Rich Hickey for a moment. 
```
compose
v.	To make up the constituent parts of; constitute or form: an exhibit composed of French paintings; the many ethnic groups that compose our nation. See Usage Note at comprise.
v.	To make or create by putting together parts or elements.
```

[...]

PR #3701, opened 2019-01-02
FIX: issue #3659 COMPOSE should only accept block! argument

meijeru:


meijeru:
I don't know why I always get the merge request combining into my PR.

greggirwin:
Errors from runners in %red/tests/source/units/auto-tests, which will likely boil down to errors in %run-all/evaluation-test.red, since you changed `compose`. Looks like there are some non-block tests that need to go away. Look for "; non-block args (non-evaluating)"

hiiamboris:
@meijeru you can run these tests locally with:
`rebol --do "do/args %red.r {tests\source\units\auto-tests\run-all-interp.red} quit"` and
`rebol --do "do/args %red.r {-r tests\source\units\auto-tests\run-all-comp1.red} quit" & run-all-comp1`
Before you do, insert a `qt-verbose: yes` line after the `#include %......quick-test.red` so you'll see exactly on what test it tripped.

greggirwin:
@hiiamboris, thanks for reminding me about qt-verbose as well. I had forgotten already.

meijeru:
Why could it still be failing?

meijeru:
Found another test using compose!

hiiamboris:
@meijeru to make it regression proof you might want to add some tests like `error? try [compose 1]` or alike.

greggirwin:
Still failing it seems, but "Removed suprtflupis test" typo made me laugh out loud. :^)

meijeru:
Yes, the typo was hilarious. Now I found one more test, not marked as "non-block" but still one of those.
This time around it should work....

meijeru:
Now Travis has a time-out, apparently on fetching some Debian stuff. 😢

greggirwin:
Hmmm, usually it's Appveyor that times out. Since Appveyor passed, I think we're OK. Just be on the alert if someone posts a related issue.

PR #3704, opened 2019-01-03
Issue 3703

endo64:
This PR fixes the `replace` with a `tag!` value issue discussed [here](https://gitter.im/red/help?at=5c2e82bfbabbc178b2222348)

Other commits appear in this PR are already merged, so it is safe to merge. The reason explained here [GitHub pull request showing commits that are already in target branch
](https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch)

### DIFF
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -228,7 +228,7 @@ replace: function [
 		]
 		return series
 	]
-	if system/words/all [char? :pattern any-string? series] [
+	if system/words/all [any [char? :pattern tag? :pattern] any-string? series] [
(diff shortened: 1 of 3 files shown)


greggirwin:
This seems like a good change, because we don't want this:
```
>> replace "aaa <tag> ccc" <tag> <bbb>
== "aaa <<bbb>> ccc"
```

endo64:
@greggirwin Tests added to the same commit.

Issue #3384, opened 2018-05-10
Wrong error message when giving wrong accessor to time value


meijeru:
### Expected behavior
```
>> t: 0:00
== 0:00:00
>> t/day
*** Script Error: cannot access day in path t/day
*** Where: catch
*** Stack:  
```
### Actual behavior
```
>> t: 0:00
== 0:00:00
>> t/day
*** Script Error: cannot access day in path [script invalid-path]
*** Where: catch
*** Stack:  
```
### Steps to reproduce the problem
run the above code
### Red and platform version
```
any
```

meijeru:
I seem to remember having made a similar issue for another datatype (since then solved). For this one, I am not sure it has already been reported.

hiiamboris:
https://github.com/red/red/blob/e0a31a954b223ca9767c73d99ce26891d83440c1/runtime/datatypes/time.reds#L281
should be `fire [TO_ERROR(script invalid-path) path element` ?
same bug with pair!

meijeru:
Merged  and therefore solved.

PR #3712, opened 2019-01-05
FIX issue #3384 wrong error messages

meijeru:
This concerns path access in time and pair values.


greggirwin:
I'll merge this @meijeru, but if you could show before and after examples, that would be great, so it's clear what the change does.

meijeru:
The examples are in the issue.

greggirwin:
Ah yes, thanks.

PR #3715, opened 2019-01-07
new updated GTK tools for red 0.6.4

rcqls:
This is the  code of the red/GTK branch updated to be merged in red/master.


dockimbel:
We can't merge it in master until it's roughly on par with the other View backends.

You should repost the PR against the red/GTK branch.

rcqls:
Bad idea since lost of history

PR #3718, opened 2019-01-08
Gtk for red-0.6.4

rcqls:
Updated code to be used at least with red-0.6.4


build/includes.r
```diff
@@ -157,6 +157,7 @@ write %build/bin/sources.r set-cache [
 			%win32.reds
 			%COM.reds
 			%image-gdiplus.reds
+			%image-gdk.reds
```
qtxie: This file is included twice.

modules/view/backends/platform.red
```diff
@@ -790,18 +794,18 @@ system/view/platform: context [
 					progress:	21
 				]
 			]
+			; GTK Backend (similar to macOS from now)
 			Linux	[
 				extend system/view/metrics/def-heights [
```
qtxie: Is there a reason to change those def-heights to simliar to macOS? IIRC, those values are chosen specific for GTK3.

build/includes.r
rcqls: For the rules.red file I 'll only keep the Cancel-OK rule.

modules/view/backends/platform.red
rcqls: This is here as a reminder to fix the values later. You know that this branch is far from being stable.

qtxie: But those values are chosen on purpose, not arbitrary. Does the new values gives better results?

rcqls: Apparently, no significant effect. I then removed this setting

PR #3737, opened 2019-01-15
Fix color issue and multiple instances of GtkApplication

rcqls:
Solve a color issue and allow us to have multiple (compiled) Gtk applications running at the same time.


qtxie:
Great! Thanks! :+1:

PR #3742, opened 2019-01-16
fix data update

rcqls:
text-list, drop-down and drop-list can now be updated when changing data field


rcqls:
> Wouldn't be better to remove debugging lines?

This work is far from being stable. From now it helps me to see quickly where there is some work to improve. I added again some DEBUG print because I don't think there is a debugger to avoid this.

@qtxie Thanks to validate the PR very quickly!

PR #3744, opened 2019-01-18
Some preliminary resizing stuff added

rcqls:
GTK part of wallet is resizing properly.

### DIFF
--- a/modules/view/backends/gtk3/events.reds
+++ b/modules/view/backends/gtk3/events.reds
@@ -67,21 +67,35 @@ get-event-offset: func [
 	evt		[red-event!]
 	return: [red-value!]
 	/local
-		offset [red-pair!]
-		value  [integer!]
+		widget	[handle!]
+		sz 		[red-pair!]
+		offset	[red-pair!]
+		value	[integer!]
 ][
 	case [
 		any [
 			evt/type <= EVT_OVER
 			evt/type = EVT_MOVING
-			evt/type = EVT_SIZING
 			evt/type = EVT_MOVE
-			evt/type = EVT_SIZE
 		][
 			offset: as red-pair! stack/push*
 			offset/header: TYPE_PAIR
 			offset/x: motion/x_new
 			offset/y: motion/y_new
+			;;print ["event-offset: " offset/x "x" offset/y lf]
+			as red-value! offset
+		]
+		any [
+			evt/type = EVT_SIZING
+			evt/type = EVT_SIZE
+		][
+			widget: as handle! evt/msg
+			;;print ["event-offset type: " get-symbol-name get-widget-symbol widget lf]
+			sz: (as red-pair! get-face-values widget) + FACE_OBJ_SIZE
+			offset: as red-pair! stack/push*
+			offset/header: TYPE_PAIR
+			offset/x: sz/x
+			offset/y: sz/y
 			;;print ["event-offset: " offset/x "x" offset/y lf]
 			as red-value! offset
 		]
--- a/modules/view/backends/gtk3/gui.reds
+++ b/modules/view/backends/gtk3/gui.reds
@@ -527,48 +527,6 @@ change-rate: func [
 	]
 ]
 
-change-size: func [
-	hWnd [handle!]
-	size [red-pair!]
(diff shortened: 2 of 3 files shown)


modules/view/backends/gtk3/events.reds
```diff
@@ -85,6 +85,20 @@ get-event-offset: func [
 			;;print ["event-offset: " offset/x "x" offset/y lf]
 			as red-value! offset
 		]
+		any [
+			evt/type = EVT_SIZING
+			evt/type = EVT_SIZE
+		][
+			widget: as handle! evt/msg
+			;;print ["event-offset type: " get-symbol-name get-widget-symbol widget lf]
+			sz: (as red-pair! get-face-values widget) + FACE_OBJ_SIZE
```
qtxie: If it's the same as the face/size, why not just `stack/push as cell! sz`? Also this info is suppose to be got from a GTK object, not from a red face.

modules/view/backends/gtk3/handlers.reds
```diff
@@ -236,6 +236,23 @@ window-removed-event: func [
...
 
+window-size-allocate: func [
+	[cdecl]
+	widget	[handle!]
+	rect	[tagRECT]
+	/local
+		sz	 [red-pair!]
+][
+	;;DEBUG: print [ "window-size-allocate " rect/width "x" rect/height lf]
+	sz: (as red-pair! get-face-values widget) + FACE_OBJ_SIZE		;-- update face/size
```
qtxie: face/size should be changed after EVT_SIZING event.

modules/view/backends/gtk3/gui.reds
```diff
@@ -710,15 +668,40 @@ change-offset: func [
...
 		0
 	][
 		;OS-refresh-window as integer! main-window
 		container: either null? hWnd [null][g_object_get_qdata hWnd gtk-fixed-id]
-		unless null? container [gtk_fixed_move container hWnd pos/x pos/y]
+		;; DEBUG: print ["change-offset by" pos lf]
+		; _widget: either type = text [
+		; 	g_object_get_qdata hWnd _widget-id
+		; ][hWnd]
+		_widget: g_object_get_qdata hWnd _widget-id
```
qtxie:
```
_widget: g_object_get_qdata hWnd _widget-id
_widget: either null? _widget [hWnd][_widget]
```
This code pattern repeats many times, it's worth/better to wrap it in a function.
BTW, why we need it? `hWnd` should be the widget handle, no?

modules/view/backends/gtk3/events.reds
rcqls: Maybe it is the solution and I'll try to apply it.  But for that, could you give me some documentation about `stack/push` (I only mostly understand what it is but I really don't master it))? Moreover there is no documentation of how the GUI framework is designed,  I try then to learn everything from the code you provided on macOS and  less often windows.  Don't forget that my first preliminary goal is in the first time to make things (mostly working). In a second step, I hope it would be easier to optimize the code.

modules/view/backends/gtk3/handlers.reds
rcqls: I'll check that...

modules/view/backends/gtk3/gui.reds
rcqls: Sometimes widget/hWnd  comes with a parent (_widget) to work. I can put this 2 lines codes in a function indeed but I am now not sure if it is used so often.
Sometimes events in gtk are connected with the parent of widget.  A gtk_label (widget) has a parent gtk_event_box (_widget) reacting to the event. When I did that, it is maybe not the perfect solution but as I said before, someone (or I) could later find a better solution . BTW, since only `-r` mode for (cross)-compilation is working  to build a binary it is a nightmare to compile so many times code to fix it. It is already an improvement for me to realize that I can cross-compile on my macOS and then  test the newly created  binary in any docker container but it still is too long and time consuming . Fixing the  `-c` mode by reducing significantly the time of compilation has to be a priority IMHO since it helps to improve the code.

rcqls:
I close this PR since I prepared a new PR...

Issue #3739, opened 2019-01-16
[React] skips updates when graph is forking


hiiamboris:
**Describe the bug**

Consider the following map of reactions:
```
s/started => a/x ,  b/y
             a/x => b/x
                    b/x, b/y => detected by a single reaction (R)
```
React will:
- detect a change in `b/y` and trigger (R)
- detect a change in `a/x` and reflect it on `b/x`
- change f `b/x` will not be detected

This leads to very poor react experience, changes often failing to be detected. And it *may be* the cause of https://github.com/red/red/issues/3731.

**To reproduce**
```
Red []

reactor: func [spec] [make deep-reactor! spec]

s: reactor [started: no]
a: reactor [x: none]
b: reactor [x: none y: none]

react [
	if s/started [
		a/x: copy "NEW-VALUE!"	print ["NEW-VALUE! => a/x"]
		b/y: copy "junk"	print ["junk => b/y"]
	]
]

react [
	print ["A/x is" a/x]
	if b/x <> a/x [ b/x: copy a/x	print [b/x "=> b/x"] ]
]

react [
	print ["B is [x:" b/x "y:" b/y "]"]
	if b/x = "NEW-VALUE!" [print "^/!!!! IT'S WORKING !!!!^/"]
]

print "^/STARTED!"
s/started: yes
```
The output I see:
```
STARTED!
NEW-VALUE! => a/x
junk => b/y
B is [x: none y: junk ]
A/x is NEW-VALUE!
NEW-VALUE! => b/x            ; b/x is changed but never triggered the (R)
A/x is NEW-VALUE!
```

**Expected behavior**

The provided script should print "!!!! IT'S WORKING !!!!" line, but it does not.

**Platform version (please complete the following information)**
```
Red 0.6.4 for Windows built 9-Jan-2019/5:46:16+03:00 commit #acf3492
```

dockimbel:
Note: you can use `system/reactivity/debug?: yes` to activate traces of reactions evaluation.

dockimbel:
After a quick look, if you just remove `"y:" b/y` from last reactive formula, it will work as expected.

Speculative explanation: the reactive engine is detecting such case as a possible infinite loop of reactions, so does not trigger the last one. It could be a bug or a limitation of the current detection method.

I will do a deeper investigation asap.

Issue #3731, opened 2019-01-14
[View/React] Seemingly simple react structure isn't working


hiiamboris:
**Describe the bug**

It's just one of weird bugs of react I encountered, but this one was easy enough to isolate.
I have a layout:
```
w: window [
  p*: panel [
    p1: panel [field]
    p2: panel [field]
  ]
]
```
And a map of reactions to be triggered on window resize:
```
w/size/y => p*/size/y => p1/size/y + p2/size/y
                         p1/size  => p2/offset
```
And it *often* fails.

**To reproduce**

Here's the sample code:
```
Red [Needs: View]

system/view/VID/styles/panel2: [
	template: [
		type: 'panel  color: cyan  size: 200x400
		actors: [
			on-created: function [face] [
				face/pane: reduce [make-face 'field]
				; face/pane: reduce [make-face 'base]
				react/link function [fa pa] [
					fa/size/y: pa/size/y
					print ["panels size" pa/size "=> single panel size" fa/size]
				] [face face/parent]
			]
		]
	]
]

view/tight/flags [
	panels: panel 400x400 magenta [at 0x0 panel2 panel2] with [
		actors: object [
			on-created: func [face] [
				react/link func [pan wnd] [
					pan/size/y: wnd/size/y
					print ["window size" wnd/size "=> panels size" pan/size]
				] [face face/parent]
				react/link func [f1 f2] [
					f2/offset: f1/size * 1x0 + 10x0
					; f2/offset/x: f1/size/x + 10
					print ["left panel size" f1/size "=> right panel offset" f2/offset]
				] [face/pane/1 face/pane/2]
			]
		]
	]
] [resize]
```

**To see it in action:**
1. **Manually** resize the window - it will stretch the panels to the window height, albeit somewhat lagging
2. **Maximize** it with the MAX button, or with Win+Up, or Win+Right or Win+Left - the panels stay of the same size and no reactions are triggered!

Note that some seemingly absurd actions - _like replacing some lines with those commented below_ - will also make it work fine.
Note also that the same (2) sometimes (depending on some more absurd changes) will work fine if performed without (1), so it's advised to follow the instruction.

**Screenshots**

Tiny panel in full screen window:
![](https://i.gyazo.com/b4aa21a3e508f671ecad0f6e23ba1561.png)

**Expected behavior**

 After maximizing the window the panel should reach it's bottom immediately.
![](https://i.gyazo.com/933b60c54be615ff77c66ba9a0403b4f.png)

[...]

PR #3746, opened 2019-01-18
#3739 fix

hiiamboris:
A fix for the reactivity issue #3739 with minimum changes I could think of. I just moved the stack unwind code into the same function that builds it up. Let @dockimbel weigh in.
By the way, doesn't solve the other #3731 issue, which may be a View problem after all.


hiiamboris:
To hell with it, I've a better solution.

Issue #3738, opened 2019-01-15
cause-error incompatible with R2/R3 as it does not accept string! as an argument


Dobeash:
**Describe the bug**

In R2/R3 the following works:
```
cause-error 'user 'message "blah"
```

but Red requires:
```
cause-error 'user 'message ["blah"]
```

**To reproduce**

```
cause-error 'user 'message "blah"
```

**Suggested Fix**

Gregg Irwin proposed changing the spec and switching from `reduce` to `compose`:

```
cause-error: func [
    {Causes an immediate error throw, with the provided information} 
    err-type [word!] 
    err-id [word!] 
    args [block! string!] 
][
    args: compose [(:args )]
    do make error! [
        type: err-type 
        id: err-id 
        arg1: first args 
        arg2: second args 
        arg3: third args
    ]
]
```

endo64:
Related [conversation](https://gitter.im/red/bugs?at=5c3cfbca8318994524ed09f5)

greggirwin:
Yay! I get a coding assignment! :^)

9214:
@greggirwin I believe this one can be closed.

greggirwin:
Was there a PR other than mine (which needed fixes)?

9214:
@greggirwin nope, I thought yours fixed the issue.

greggirwin:
@hiiamboris found a problem, related to how the test system works with errors, so I need to revisit.

Issue #1136, opened 2015-04-29
Interpreter returns error about missing paren instead of bad lit-word


x8x:
In the following code there is no missing paren but a bad lit-word:
```rebol
red>> a: func [][set 'b: 1]
*** Syntax error: missing ] at [][set'b:1]
*** Where: do
```
also notice that in some cases like above error report trims all spaces from code

PR #3747, opened 2019-01-18
FEAT: #3738 make cause-error Rebol compatible for string arg

greggirwin:


greggirwin:
Regression test #1136 throwing an access violation, though it doesn't in the console.
```
	--test-- "#1136"
		e1136: try [load {a: func [][set 'b: 1]}]
		--assert not not all [
			equal? e1136/type 'syntax
			equal? e1136/id 'invalid
			equal? e1136/arg1 lit-word!
		]
```

greggirwin:
Backed out my changes and it still fails with those errors. Something else is amiss.

hiiamboris:
@greggirwin it won't work. Test 1396 on your branch makes a malformed error object because it's arguments (inside a block) weren't reduced:
```
root size: 2239, root max: 4184, cycles: 0
root size: 2239, root max: 4184, cycles: 1
root size: 2239, root max: 4184, cycles: 2
make error! [
    code: 200
    type: 'syntax
    id: 'invalid
    arg1: 'type
    arg2: "5+2)"
    arg3: none
    near: none
    where: 'do
    stack: 3539256
]
```
Then somewhere down the output code it is hardcoded that the error is well formed and it crashes otherwise:
```
*** in file: /D/devel/red/red-src/irwins/runtime/datatypes/object.reds
*** at line: 97
***
***   stack: red/object/rs-select 028AD604h 032C123Ch
***   stack: red/error/reduce 028AD624h 028AD604h
***   stack: red/error/form 028AD604h 028AD614h 028AD5F4h -18
***   stack: red/actions/form 028AD604h 028AD614h 028AD5F4h 0
***   stack: red/actions/form* -1
***   stack: red/natives/prin* false
***   stack: red/natives/print* false
***   stack: red/stack/throw-error 028AD634h
***   stack: red/fire 1 0018FF10h
***   stack: red/type-check 0289DB84h 0 028AD604h
***   stack: --assert
```
You should maybe make a special case for string `args`, instead of compose...
The cause of crash in 1750 is likely the same. 1136 is probably a false success, just a stroke of luck that pointer value occurred within the allocated frame.

greggirwin:
Thanks for digging in @hiiamboris. Hoping to make some time for this today.

PR #3750, opened 2019-01-19
Gtk: patch draw,  scale and resize

rcqls:
I closed the previous PR "Fix scale and image issues on draw".

This PR merges the previous PR updated  with changes that @qtxie asked and the new development on draw fixes.


qtxie:
Great work! Thanks! :-)

PR #3752, opened 2019-01-20
Gtk patch make font

rcqls:
Main improvement: make-font now returns pango_font_description and get-text-size has now the same signature on Windows, macOS and Linux (see platform.red).


rcqls:
This PR was not complete because get-text-size was not working... I close it and submit a new PR

Issue #24, opened 2011-05-01
Bug in set-width (?)


meijeru:
I was trying to write my own (primitive) version of the 'form function, and hit on the following bug, when compiling 

```
red/system []
i: 3
s: "          "
s/1: as byte! ((i // 10) + 48)
print s
```

Result:

```
** Script Error: size-of? expected type argument of type: word
** Where: set-width
** Near: width: emitter/size-of? operand-type? operand
```

PR #3758, opened 2019-01-27
FIX: improve func's spec parser to support local details

bitbegin:
>> probe help-ctx/func-spec-ctx/parse-func-spec v: func [a [integer!] b [integer!] /local x [integer!] "x" y [block!] "y"][]
make object! [
    desc: none
    attr: none
    params: [[name a type [integer!] desc none] [name b type [integer!] desc none]]
    refinements: []
    locals: [[name x type [integer!] desc "x"] [name y type [block!] desc "y"]]
    returns: []
]

### DIFF
--- a/environment/console/help.red
+++ b/environment/console/help.red
@@ -252,6 +252,7 @@ help-ctx: context [
 		parse-func-spec: function [
 			"Parses a function spec and returns an object model of it."
 			spec [block! any-function!]
+			/noprint		"don't print wangings"
 			/local =val		; set with parse, so function won't collect it
 		][
 			clear stack
@@ -272,7 +273,9 @@ help-ctx: context [
 						parse reduce =val [some [datatype! | typeset!]]
 						parse =val ['function! block!]
 					][
-						print ["Looks like we have a bad type spec:" mold =val]
+						unless noprint [
+							print ["Looks like we have a bad type spec:" mold =val]
+						]
 					]
 				)
 			]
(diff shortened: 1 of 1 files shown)


meijeru:
Important! See also red/REP issue #24

environment/console/help.red
```diff
@@ -252,6 +252,7 @@ help-ctx: context [
 		parse-func-spec: function [
 			"Parses a function spec and returns an object model of it."
 			spec [block! any-function!]
+			/noprint		"don't print wangings"
```
greggirwin: Suggest `/quiet` as the refinement name, and "don't print wangings" should be "don't print warnings" I think. :^) Even better might be to call it `/warn` so the check then reads `if warn ...`.

greggirwin:
The extra info can be useful, as documentation, but we need to know (to support Rudolf's spec work), what is meant to be supported. @dockimbel will have to say if he wants type enforcement on locals, which may be a lot more overhead. If not, do we need to allow type and docstring elements for locals. Could be used by tooling, but certainly goes beyond what `help` would show.

bitbegin:
> just want to use this internal function for some other tool, and i have copied it to my codes(not necessary to merge this). 

I used /local like this way `/local a [integer!]` for a long time, and just realized that `/local a` would be enough.
So, `/local a [integer!]` or `/local a [integer!] "desc"` are right?

greggirwin:
> So, /local a [integer!] or /local a [integer!] "desc" are right?

They don't throw an error, but they don't do anything either. That is, there is no runtime checking support for local type enforcement, and showing the doc strings for internals may be exposing too much in `help`, which is for usage info. `Source` will still show them. `Help`'s `parse-func-spec` could be improved, and externalized from `help`, for more general use. Sounds like you're starting on that. :+1:

dockimbel:
Optional locals typing and docstring is welcome, though:

* the current runtime code might not account for that everywhere, so could result in crashes or corruptions. This would need an important testing effort to determine if current Red can support it flawlessly, or need changes.

* it was meant to be used by current compiler for better code generation, though my latest research on that shows that it might not be doable without also modifying some of the semantics, resulting in a subset of Red (codename: "Red/Strict"). So until the work on Red/Strict starts (not this year), there will be no semantics attached to those extra spec information.

greggirwin:
@dockimbel or @qtxie would you get with @bitbegin on this, make a decision, and merge or close this PR? Thanks.

dockimbel:
Red does not support typed locals for now (and unlikely will do in the future, though a future dialect of Red might do), so allowing them in the spec block would cause confusion to users.

Issue #2182, opened 2016-08-27
Runtime crash on using forall with word! that value isn't series.


ifgem:
```
sym: 10 ; 10.0, ...
forall sym []
*** Runtime Error 1: access violation
*** at: 004485F5h
```

dockimbel:
Fixed for both compiler and interpreter.

greggirwin:
Is this the expected error message?

red>> n: 10 forall n []
*** Script Error: forall does not allow series! for its 'word argument
*** Where: forall

Seems like it should say "requires a series!"

dockimbel:
Good catch!

hiiamboris:
There's a recent regression in the compiler (between Dec 07 and Dec 18). Should be reopened.

`1.red`:
```
Red []
sym: 10 forall sym []
```

`red -r -d 1.red`:
```
*** Runtime Error 1: access violation                        
*** in file: /D/devel/red/red-src/red/datatypes/object.reds  
*** at line: 97                                              
***                                                          
***   stack: red/object/rs-select 027DD254h 031EC58Ch        
***   stack: red/error/reduce 027DD274h 027DD254h            
***   stack: red/error/form 027DD254h 027DD264h 027DD244h -18
***   stack: red/actions/form 027DD254h 027DD264h 027DD244h 0
***   stack: red/actions/form* -1                            
***   stack: red/natives/prin* false                         
***   stack: red/natives/print* false                        
***   stack: red/stack/throw-error 027DD294h                 
***   stack: red/fire 1 0018FF38h                            
***   stack: red/type-check 027CD964h 0 027DD254h            
```

greggirwin:
Interpreter OK.

@hiiamboris, let's make a list of all these regressions, so we can give it to @dockimbel and @qtxie for easy referencing.

hiiamboris:
@greggirwin OK just added the list to the https://github.com/red/red/pull/3781

Issue #1974, opened 2016-06-05
Crash on passing word! as argument instead of a string!


Zamlox:
Following sequence will crash console:

```
f1: func [p [string!]][print p]
reflect f1 'spec
```

Call stack:
```
*** Runtime Error 1: access violation
*** in file: /D/work/red/source/red/runtime/datatypes/context.reds
*** at line: 270
***
***   stack: red/_context/get 028E69E4h
***   stack: red/error/get-call-argument 0
***   stack: red/interpreter/eval-arguments 028F3BD8h 029311ACh 029311ACh 00000000h 00000000h
***   stack: red/interpreter/eval-code 028F3BD8h 0293119Ch 029311ACh true 00000000h 00000000h 028F3BD8h
***   stack: red/interpreter/eval-expression 0293119Ch 029311ACh false true
***   stack: red/interpreter/eval-arguments 028EEE78h 0293118Ch 029311ACh 00000000h 00000000h
***   stack: red/interpreter/eval-code 028EEE78h 0293118Ch 029311ACh false 00000000h 00000000h 028EEE78h
***   stack: red/interpreter/eval-expression 0293118Ch 029311ACh false false
***   stack: red/interpreter/eval 02863CB0h true
***   stack: red/natives/catch* true 1
***   stack: ctx270~try-do 0028C5DCh
***   stack: ctx270~do-command 0028C5DCh
***   stack: ctx270~eval-command 0028C5DCh
***   stack: ctx270~run 0028C5DCh
***   stack: ctx270~launch 0028C5DCh
***   stack: ctx289~launch 0028C0E0h
```

meijeru:
Curiously, if the `spec` is empty, there is no crash but an `unset` error.

dockimbel:
The issue is that you call `f1` instead of getting a reference to the function (using `:f1`). So, it's not related to `reflect`:
```
f1: func [p [string!]][print p]
reflect :f1 'spec
== [p [string!]]

f1: func [p [string!]][print p]
f1 'spec
== <crash>
```

hiiamboris:
There's a recent regression in the compiler (between Dec 07 and Dec 18). Should be reopened.

`1.red`:
```
Red []
do [
	f1: func [p [string!]][print p]
	reflect f1 'spec
]
```

`red -r -d 1.red`:
```
*** Runtime Error 1: access violation                                                                   
*** in file: /D/devel/red/red-src/red/datatypes/object.reds                                             
*** at line: 97                                                                                         
[...]

Issue #1935, opened 2016-05-26
Invalid path crashes console


philipyi:
Type the following into the console and it will crash:
test/:

hiiamboris:
There's a recent regression in the compiler (between Dec 07 and Dec 18). Should be reopened.

`1.red`:
```
Red []
do load {test/:}
```

`red -r -d 1.red`:
```
*** Runtime Error 1: access violation                              
*** in file: /D/devel/red/red-src/red/runtime/datatypes/object.reds
*** at line: 97                                                    
***                                                                
***   stack: red/object/rs-select 0284D274h 0325C548h              
***   stack: red/error/reduce 0284D294h 0284D274h                  
***   stack: red/error/form 0284D274h 0284D284h 0284D264h -18      
***   stack: red/actions/form 0284D274h 0284D284h 0284D264h 0      
***   stack: red/actions/form* -1                                  
***   stack: red/natives/prin* false                               
***   stack: red/natives/print* false                              
***   stack: red/stack/throw-error 0284DBB4h                       
***   stack: red/natives/do* true -1 -1 -1                         
***   stack: ctx264~transcode 0030E3E4h                            
***   stack: load                                                  
```

greggirwin:
This should be a new ticket I think, as it's not the console crashing, but a compiler issue now.

greggirwin:
We can reference this ticket there though.

Issue #1598, opened 2016-02-05
pair // float


mikeparr:
;pair remainder float  on Windows 7, red-054.exe
red>> 3x4 // 1.1
**\* Math Error: unsupported right operand for pair operation
**\* Runtime Error 13: integer divide by zero
**\* at: 0042D497h
;then red-459 crashes.
Note that pair // integer is OK, but Red help says that float is OK as well.
The Math Error message is not the problem, but Red crashes immediately after displaying it.
The same problem happens with the remainder function.

hiiamboris:
There's a recent (1-2 months) regression. Should be reopened.

`1.red`:
```
Red []
3x4 // 1.1
```

`red -r -d 1.red`:
```
*** Runtime Error 1: access violation                              
*** in file: /D/devel/red/red-src/red/runtime/datatypes/object.reds
*** at line: 97                                                    
***                                                                
***   stack: red/object/rs-select 027AD264h 031BBCC8h              
***   stack: red/error/reduce 027AD284h 027AD264h                  
***   stack: red/error/form 027AD264h 027AD274h 027AD254h -18      
***   stack: red/actions/form 027AD264h 027AD274h 027AD254h 0      
***   stack: red/actions/form* -1                                  
***   stack: red/natives/prin* false                               
***   stack: red/natives/print* false                              
***   stack: red/stack/throw-error 027AD324h                       
***   stack: red/fire 0 0018FE88h                                  
***   stack: red/actions/compare 027AD2F4h 027AD304h 3             
***   stack: red/actions/compare* 3                                
***   stack: red/natives/lesser?* true                             
***   stack: mod                                                   
***   stack: modulo                                                
```

endo64:
Confirmed on Win10:

```
C:\Users\endo\Documents\Red\red\build\bin>test.exe
root size: 2153, root max: 4127, cycles: 0
root size: 2153, root max: 4127, cycles: 1
root size: 2153, root max: 4127, cycles: 2

*** Runtime Error 1: access violation
*** in file: /C/Users/endo/Documents/Red/red/build/bin/datatypes/object.reds
*** at line: 97
***
***   stack: red/object/rs-select 026FD374h 029AC31Ch
***   stack: red/error/reduce 026FD394h 026FD374h
***   stack: red/error/form 026FD374h 026FD384h 026FD364h -18
***   stack: red/actions/form 026FD374h 026FD384h 026FD364h 0
***   stack: red/actions/form* -1
***   stack: red/natives/prin* false
[...]

PR #3781, opened 2019-02-14
Compiler regression tests

hiiamboris:
These are the missing parts of PR #2191:
1. R/S compiler regression tests
2. Red compiler regression tests

I was hesitant for a long time to include these into the master branch. But after it caught 10 regressions at once I realized the impact it has.

Part 2 is only included into the `tests/run-all.r` (run only by the CI/CD setup), for it's pretty slow (about 5 minutes). It is split into five parts for the same reason. When your test №2000+ fails you don't want to start the whole run over. It's just more convenient to have it in parts in my experience.

Tests for this part should fail for now until we fix the december regression. Regressions are caused by https://github.com/red/red/commit/4414d6473cfea80aecf32c6b8aa5ee99b161715d : #2182 #1207 #1171 #1120 #1035 #1598 #837 #1974 #1935 #916

Once this PR is merged, we can close 2191 and make @rebolek happy :)


greggirwin:
I'll let @dockimbel  or @qtxie make the call on this one. Given the results, I think it's good to include it, but we also need to look at testing speed overall, to see what we can do there.

hiiamboris:
Yes. It's up to @dockimbel to first fix the stack thing that caused 10 tests to fail. Don't merge until that or no new automated builds :)

dockimbel:
@hiiamboris Thanks for catching those regressions!

Indeed, we need to make the tests pass before including them in %run-all.r. As we run the whole test suite often dozens of times a day, the running speed is critical for our productivity. Keeping the running time as short as possible is high priority. Splitting the tests in several parts is a good option.

hiiamboris:
Some of those old tickets were reopened. Isn't it enough?

dockimbel:
@hiiamboris Given the number of regressions that all caused by the same commit, an umbrella "epic" ticket would be welcome.

hiiamboris:
Interesting. Tests fail but CI/CD gives OK.

dockimbel:
After the latest fixes, almost all the new regression tests are passing except for #3733 and #3773. See the following comments for details.

tests/source/compiler/regression-test-redc-5.r
```diff
@@ -0,0 +1,209 @@
...
+		--assert not crashed?
+		--assert script-error?
+
+		;; FIXME: this should not error out when compiled:
+		--compile-and-run-this-red {
+			do [f: func [/q] bind [1] context []]
+			f/q
+		}
+		--assert not crashed?
+		; --assert not script-error?
```
dockimbel: This test is not a regression, it's a new issue and needs its own ticket.

tests/source/compiler/regression-test-redc-5.r
```diff
@@ -0,0 +1,209 @@
...
+		}
+		--assert compiler-error?
+
+		;; :mc = func [x][x], so `mc :mc` executing `x` applies it to an empty arg list => error
+		--compile-and-run-this-red {
+			#macro mc: func [x] [x]
+			probe quote (mc :mc)
+		}
+		--assert compiler-error?
+
```
dockimbel: Those tests are invalid, compiled macros are run by **Rebol2**, not by Red. ;-)

hiiamboris: Haha :) Right! I'll remove it then

tests/source/compiler/regression-test-redc-5.r
hiiamboris: Okay: https://github.com/red/red/issues/3945

hiiamboris: It should pass the `not crashed?` test. The `not script-error?` test should be added by those who fix the new #3945 issue, to completely cover it.

hiiamboris:
@dockimbel I've fixed the tests, however there's a recent (likely, July) R/S compiler regression that makes tests 273 and 275 of %regression-test-rsc.r fail on // operator:
```
*** Compilation Error: argument type mismatch on calling: //
*** expected: [number!], found: [float!]
```

dockimbel:
@hiiamboris Support for `//` and `%` on floats has been removed, those operations are no more supported in R/S (see red/docs/red-system-specs-light.html#section-8.1). The reason is the lack of clear definition for such operations and difficulty to get consistent results on all platforms.

dockimbel:
@hiiamboris Are we good now? Is there anything remaining to fix or change, or can we merge this PR now?

hiiamboris:
@dockimbel yeah, all set.
The failure in on ARM only, unrelated to the PR.

Issue #3717, opened 2019-01-08
[Red lexer] when parse "]", `system/lexer/transcode/one` should give an error!


bitbegin:
**Describe the bug**

```
>> pos: "]"
== "]"
>> out: []
== []
>> w: system/lexer/transcode/one pos clear out false
== "]"
>> w = pos
== true
>> 
>> out
== []
>> 
```

`w` place should get an error!

**Expected behavior**

```
>> pos: "]"
== "]"
>> out: []
== []
>> w: system/lexer/transcode/one pos clear out false
; here should have an error!
```

**Platform version (please complete the following information)**
```
Red 0.6.4 for Windows built 4-Jan-2019/2:08:48+08:00 commit #af59285
```

bitbegin:
another test case:

```
>> load/next "]" 'a
== []
>> a
== "]"
>> 
```

PR #3785, opened 2019-02-20
FIX: #3717 (make error when transcode right bracket)

bitbegin:


greggirwin:
Nice set of added tests for this. Since all checks passed, I'll merge this, and @dockimbel  or @qtxie can review if needed.

Issue #3786, opened 2019-02-20
WISH: add --quiet flag to red.r to prevent consoles from launching after compiling


gltewalt:
Can be useful for admin scripts written in Red that keep Red binaries current, or for a future `upgrade` feature.
I have made the changes to red.r in my repo.

```
C:\Users\Old Man\Desktop\red\build\bin>red --cli --quiet
Compiling compression library...
Compiling Red console...

C:\Users\Old Man\Desktop\red\build\bin>
```

gltewalt:
Used in a script. 

```
C:\Users\Old Man>rbu
Updating Red...
Compiling compression library...
Compiling Red console...
Compiling Red GUI console...
Done.
C:\Users\Old Man>
```

endo64:
Didn't your PR work?

9214:
@endo64 Greg forgot to add the final touch.

gltewalt:
It works but the PR was an experiment with GitHubs new Draft feature. 
Do I make a regular PR and tie it to this Wish somehow?

9214:
@gltewalt you can reference it from a ticket as a proof-of-concept.

gltewalt:
#3787

dockimbel:
@gltewalt What about this ticket? Is the PR for it ready?

endo64:
@dockimbel @gltewalt It looks that PR (already merged #3787) doesn't work as expected:

```
D:\Projects\Red\red\build\bin>..\build.r

D:\Projects\Red\red\build\bin>red --cli --quiet
Compiling compression library...
Compiling Red console...
*** Access Error: cannot open: %--quiet
*** Where: read
*** Stack:
--== Red 0.6.4 ==--
Type HELP for starting information.

>>
```

9214:
@endo64 there's no indication that PR was merged, @gltewalt closed it for later resubmission.

endo64:
@9214 Thanks, I thought it was merged.

PR #3787, opened 2019-02-20
Proof of concept for WISH #3786

gltewalt:


red.r
```diff
@@ -525,11 +526,13 @@ redc: context [
 		]
 		exe: safe-to-local-file exe
 
-		either gui? [
-			gui-sys-call exe any [all [file form-args file] ""]
-		][
-			if with [repend exe [" " form-args file]]
-			sys-call exe								;-- replace the buggy CALL native
+    unless quiet? [
```
greggirwin: Needs leading tabs instead of spaces.

red.r
```diff
@@ -436,8 +436,9 @@ redc: context [
 	]
 
 	run-console: func [
-		gui?	[logic!]
-		debug?	[logic!]
+		gui?	   [logic!]
+		debug?	 [logic!]
+		nol?     [logic!] 								;-- no launch
```
greggirwin: `nol?` is...not a great name.

greggirwin:
Seems like it would be better to make the option an affirmative, rather than a negative. i.e. `launch-console`.

red.r
gltewalt: Hmm, thought it was tabs...

red.r
gltewalt: Yeah, I changed to `quiet?` not too long after that version.

gltewalt:
> 
> 
> Seems like it would be better to make the option an affirmative, rather than a negative. i.e. `launch-console`.

Do you mean change the default behavior for the consoles to compiling only? Launch only if a flag is supplied?

greggirwin:
No, I mean the default is still to launch the console, but *internally* that default is true. If a "--no-console" flag is given, `launch-console` gets set to false.

gltewalt:
I had to re fork my red repo, so I can't make changes to the wish PR anymore.

For comment, these are the changes in line with your suggestion:
https://github.com/gltewalt/red/commit/ba2c63c2accaf8bbca1fb6d2af841f4fef5b4af0

gltewalt:
Will resubmit at a later time

PR #3791, opened 2019-02-23
FEAT: Added last? function

endo64:
This PR adds `last?` function.
Test added.

```
>> last? []
== false
>> last? [1]
== true
>> last? make hash! [1]
== true
>> last? make hash! [1 2]
== false
>> last? make vector! [1 2]
== false
>> last? make vector! [1]
== true
>> last? ""
== false
>> last? "1"
== true
```

### DIFF
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -1044,7 +1044,7 @@ average: func [
 	divide sum block to float! length? block
 ]
 
-last?: func [
+single?: last?: func [
 	"Returns TRUE if the series length is 1"
 	series [series!]
 ] [
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -1044,6 +1044,13 @@ average: func [
 	divide sum block to float! length? block
 ]
 
+last?: func [
+	"Returns TRUE if the series length is 1"
+	series [series!]
+] [
+	1 = length? series
+]
+
 ;------------------------------------------
 ;-				Aliases					  -
 ;------------------------------------------
--- a/tests/source/environment/functions-test.red
+++ b/tests/source/environment/functions-test.red
@@ -493,4 +493,10 @@ Red [
 		--assert 0.5 = average [1 30% 20%]
 ===end-group===
 
+===start-group=== "last? tests"
+	--test-- "last? test"
+		--assert false = last? []
+		--assert true = last? [1]
+===end-group===
+
 ~~~end-file~~~
--- a/tests/source/environment/functions-test.red
+++ b/tests/source/environment/functions-test.red
@@ -494,9 +494,24 @@ Red [
 ===end-group===
 
 ===start-group=== "last? tests"
-	--test-- "last? test"
-		--assert false = last? []
+	--test-- "last? true tests"
+		--assert error? try [last? 2]
 		--assert true = last? [1]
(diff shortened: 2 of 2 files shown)


PeterWAWood:
1. The function accepts map! values. This raises the question of how to define last? for a map. Is it when `2 = length? a-map-value`. The function should either raise an error when receiving a map! or use an agreed definition of an empty? map.

2. The tests should include, at least all of the examples above, and really should include all valid datatypes (e.g. binary) and error tests such as
```
    --assert error? try [last? 2]
```

3. @dockimbel should be consulted as to whether this function should be included in the standard Red distribution.

gltewalt:
`length?` appears to count each key: value pair as one value with `map!`, so `1 = length? a-map-value` would be consistent with that.

 However, as I read the "Files changed" here at 12:40AM, `last?` accepts `series!`, and `map!` isn't part of the series typeset.

PeterWAWood:
You are correct about map! Greg

endo64:
@PeterWAWood `last?` accepts `series!` but not `map!` values:

```
>> m: #(a: 1)
>> last? m
*** Script Error: last? does not allow map! for its series argument
>> series? m
== false
```

About tests, you're right I'm going to update the PR with more tests.

And final decision belongs to @dockimbel or @greggirwin for sure.

meijeru:
From the tests I get the impression this is a test for single component, it could as well be called `single?`. I seem to recall that Carl added such a thing to R3.

endo64:
`single?` and `last?` are same in R3. But I prefer `last?` as I usually use it in that meaning (in a `forall` loop to check is it the last value)

My first `last?` function was: `last?: func [series] [tail? next series]` but this one is slower and `last? []` returns `true`.

greggirwin:
While `single?` and `last?` may use the same logic in some cases, that may not always hold true. And their meaning, when expressing intent, is very different. `Last?` indicates that you are iterating, as @endo64 mentions. `Single?` says "Is there only one?" Both are useful words. 

As I noted on @gltewalt's `last` proposal, to include date/time support, `last` itself implies ordering, which means a series of some kind in Red. In this case `single?` denotes multiplicity, or a container of some kind. Objects and maps are containers, but aren't ordered. 

The hard part of language design is...all of it. ;^) We could easily say that a few generic functions work on all types, so access is universal, but that has a big downside. Constraints, while sometimes inconvenient, are powerful tools that help us manage complexity (one of Red's key goals). To that end, we have to find a balance between specificity and generality. 

`Last?` is a good name, clear, and complementary to `last`. `Single?`, on the other hand, has never struck me as a great word for this; but it's the best *single* word we have. We can't use `one?`, because `zero?` makes that confusing. Compound words are the only other choice. After word choice, then we have to decide whether to leverage minimal code or make them specific in their interfaces.

greggirwin:
I do think these are worth including in the standard runtime though, because they are small and fit well with Red's data-centric model.

endo64:
@greggirwin So you say `single?: last?: func [` in the PR?

greggirwin:
@endo64, yes, I think that's the way to go right now.

gltewalt:
`last?`, Sems

endo64:
Done.

Issue #2083, opened 2016-06-30
broken chain reactor


WiseGenius:
```
red>> a: make reactor! [x: 1 y: is [x + 1] z: is [y + 1]]
== make object! [
    x: 1
    y: 2
    z: 3
]
red>> a/x: 4
== 4
red>> a
== make object! [
    x: 4
    y: 5
    z: 3
]
red>> a/z
== 3
```
Expected `6`.

dockimbel:
I forgot to implement events triggering in the SET native when applied to object. IS function is relying on it, though REACT is not affected by that. Will implement it today.

PR #3792, opened 2019-02-25
FEAT: support `empty? map!`

bitbegin:


endo64:
Using `switch` looks faster than `either`, and would be better when we add other value types like `port!`

```
empty1: func [
	"Returns true if a series is at its tail or a map! is empty"
	series	[series! none! map!]
	return:	[logic!]
][
	either map? series [
		series = #()
	][
		either series = none [true][tail? series]
	]
]

empty2: func [
	"Returns true if a series is at its tail or a map! is empty"
	series	[series! none! map!]
	return:	[logic!]
][
	switch/default type?/word series [
		series! [tail? series]
		map! [series = #()]
	] [none! [true]]
]
```

Here are the test results:

```
>> profile/show/accuracy [ [empty1 []] [empty2 []] ] 0:0:0.5
Time                 | Memory      | Code
1.0x (856ns)         | 168         | [empty2 []]
1.82x (2μs)          | 320         | [empty1 []]
>> profile/show/accuracy [ [empty1 #()] [empty2 #()] ] 0:0:0.5
Time                 | Memory      | Code
1.0x (963ns)         | 320         | [empty1 #()]
1.02x (983ns)        | 168         | [empty2 #()]
>> profile/show/accuracy [ [empty1 none] [empty2 none] ] 0:0:0.5
Time                 | Memory      | Code
1.0x (898ns)         | 168         | [empty2 none]
1.52x (1μs)          | 320         | [empty1 none]
>> profile/show/accuracy [ [empty1 [1 2]] [empty2 [1 2]] ] 0:0:0.5
Time                 | Memory      | Code
1.0x (875ns)         | 168         | [empty2 [1 2]]
1.75x (2μs)          | 320         | [empty1 [1 2]]
>> profile/show/accuracy [ [empty1 #(a: 1 b: 2)] [empty2 #(a: 1 b: 2)] ] 0:0:0.5
Time                 | Memory      | Code
1.0x (966ns)         | 168         | [empty2 #(a: 1 b: 2)]
1.01x (978ns)        | 320         | [empty1 #(a: 1 b: 2)]
```

hiiamboris:
My thoughts exactly, @endo64 
Although I think the switch should be:
```
switch/default type?/word series [
	none! [true]
	map! [#() = series]
] [tail? series]
```
Otherwise, won't work.

endo64:
Thanks @hiiamboris I've fixed my `empty2` function's default case, it was a copy/paste mistake.
However I prefer to put the most used case in the beginning of the `switch` block, which is `block!` (hence  `series!`) and then `map!` and the last one `none!`.

hiiamboris:
@endo64 you're forgetting `series!` is a typeset ☺

bitbegin:
@endo64 can you create a PR?

endo64:
@hiiamboris Sure, I meant:

```
switch/default type?/word series [
		series! [tail? series]
		map! [series = #()]
	] [true]
```

@bitbegin I can make a separate PR but wouldn't be better if you update yours? If you are agree with me of course.

bitbegin:
this can't work

bitbegin:
```
	case [
		series? series [tail? series]
		map? series [series = #()]
		true [true]
	]

**Comment by endo64** [comment]
Why `switch` doesn't work? Any error or failed tests, in what case it doesn't work?

**Comment by bitbegin** [comment]
```
	--test-- "#2083"
		a2083: make reactor! [x: 1 y: is [x + 1] z: is [y + 1]]
		a2083/x: 4
		--assert equal? 6 a2083/z
		unset 'a2083

bitbegin:
`type?/word series` = `block!`, can't match `series!`

endo64:
But there is no `block!` in my function,

```
switch/default type?/word series [
	series! [tail? series]
	map!    [series = #()]
]               [true]
```

I meant that `series! [tail? series]` line should be the first because it is the most common case, then `map!` and then `none!`.

`switch` should be faster than `case` in this scenario, right?

bitbegin:
i mean `'series! <> 'block!`

9214:
@endo64 `series!` is a typeset, not a type. `type? ...` simply can't return `series!`.

greggirwin:
We can profile `case` vs `switch`, though I don't think there's a clear winner either way here. I will caution against using a default `true` case, and vote for being explicit in the type checks. If we make the series check the default, as @hiiamboris did, it gives us a little more protection long term.

greggirwin:
But I also agree with @endo64 that having the most common case first is nice.

endo64:
@bitbegin @hiiamboris @9214  Ah, how stupid I am :) Sure it won't work. 

@greggirwin about default `true`, it doesn't harm as the function spec restricts the parameter type with `series!`, `map!` and `none!` (and later `port!` will be)

greggirwin:
@endo64, things like that hit all of us at times. :^)

The spec is currently restricted, and not likely to expand much in this case, so it's more of a general best practice than anything. Not required by any means here.

bitbegin:
changed to this,
```
	case [
		series? series [tail? series]
		map? series [series = #()]
		none? series [true]
	]
```
is this will be better?

PR #3816, opened 2019-03-13
FIX: lit-path!  or get-path! with ":" be treated as set-path!

bitbegin:
For example: 'a/b: 1 for :a/b: 1 should not be compiled ok

### DIFF
--- a/environment/lexer.red
+++ b/environment/lexer.red
@@ -327,7 +327,7 @@ system/lexer: context [
 			length [integer! string!]
 		return: [block!]
 		/local
-			new s e c pos value cnt type process path
+			new s e c pos value cnt type wtype process path
 			digit hexa-upper hexa-lower hexa hexa-char not-word-char not-word-1st
 			not-file-char not-str-char not-mstr-char caret-char
 			non-printable-char integer-end ws-ASCII ws-U+2k control-char
@@ -624,9 +624,10 @@ system/lexer: context [
 					  reject
 				]
 			]
-			opt [#":" (type: set-path! set-path back tail stack)][
-				ahead [path-end | ws-no-count | end] | (throw-error [type path])
+			opt [#":" (type: set-path! set-path back tail stack
+					if (wtype <> word!) [throw-error [type path]])
 			]
+			[ahead [path-end | ws-no-count | end] | (throw-error [type path])]
 			(pop stack)
 		]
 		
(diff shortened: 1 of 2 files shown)


bitbegin:
before: 
```
a: context [b: 0]
probe 'a/b
probe 'a/b: 1

probe :a/b
probe :a/b: 2

```

will be allowed.

After merge this PR:

```
Red []

a: context [b: 0]
probe 'a/b
probe a/b: 1

probe :a/b
probe a/b: 2

```

only this way be compiled ok.

greggirwin:
We certainly want tests to go along with this, as it's an important, syntactic decision. A `lit-set-path!` has no effect, and a `get-set-path!` is already handled by `set-path!` (just talking about the interpreter and design, not compiler issues). That is, if the final selector results in an active value, like a func, it's not evaluated if a `set-path!` is used.

dockimbel:
Those syntactic forms should not be `load`-able (even though they could be created programmatically), as they have confusing meaning.

The fact that they were allowed so far should be considered a bug and this PR is a bugfix, restricting the lexer to only valid syntaxes.

@greggirwin If you have no objections, we can accept this PR.

greggirwin:
No objections here.

greggirwin:
Though appveyor issue doesn't look like a timeout.

bitbegin:
it passed on Rebol 2.7.8 , failed on 2.7.6 (but this version not response on my PC). 

Seems run-all-comp1.exe crashed, when used "-d" flag, this can be reproduced (because of `--assert error? try [1 + if false [2]]`). (but strange with this PR)

greggirwin:
I'll let you and @dockimbel make the call on merging it then, just in case.

environment/lexer.red
```diff
@@ -624,9 +624,10 @@ system/lexer: context [
 					  reject
 				]
 			]
-			opt [#":" (type: set-path! set-path back tail stack)][
-				ahead [path-end | ws-no-count | end] | (throw-error [type path])
+			opt [#":" (type: set-path! set-path back tail stack)
+				(if wtype <> word! [throw-error [type path]])
```
bitbegin: if change to: 
```
(type: set-path! set-path back tail stack if wtype <> word! [throw-error [type path]])
```
test case will be failed on Rebol 2.7.6

greggirwin:
Ping. OK to merge?

bitbegin:
@greggirwin can't merge, as i still can't solve the ci issue

greggirwin:
Guess we'll wait for @qtxie or @dockimbel to assist. Thanks.

dockimbel:
Closing this PR, as it has been replaced by another fix (referenced above).

PR #3828, opened 2019-03-29
FIX: get on object! doesn't return values-of object (regression)

endo64:
This PR fixes the regression of using `get` on `object!` values that should return the values of the given object.

```
og1: make object! [a: 1 b: 2 c: "x"]
>> get :og1  ;before the PR
*** Script Error: get does not allow object! for its word argument

>> get :og1  ;after the PR
== [1 2 "x"]
```

### DIFF
--- a/environment/natives.red
+++ b/environment/natives.red
@@ -198,7 +198,7 @@ compose: make native! [[
 
 get: make native! [[
 		"Returns the value a word refers to"
-		word	[word! path!]
+		word	[word! path! object!]
 		/any  "If word has no value, return UNSET rather than causing an error"
 		/case "Use case-sensitive comparison (path only)"
 		return: [any-type!]
--- a/tests/source/units/object-test.red
+++ b/tests/source/units/object-test.red
@@ -2179,6 +2179,20 @@ Red [
 				
 ===end-group===
 
+===start-group=== "get"
+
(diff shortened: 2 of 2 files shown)


greggirwin:
Thanks Sems.

meijeru:
The semantics is now identical to R3, but there are still the following differences with R2: there the value `none` is allowed (result: `none`) and `path!` is not allowed as argument type. Something to note in the Wiki under Differences between Red and Rebol.

greggirwin:
Thanks @meijeru. @gltewalt, would you please update the wiki? Thanks.

endo64:
We should consider adding `none` support.

But R3's behavior is weird:

```
>> get 1
== 1
>> get "x"
== "x"
>> probe get :print
make native! ...
```

Agreed that should be noted on Red & Rebol Diff wiki page. I'll do it.

meijeru:
When I said the semantics is identical to that of R3, I was referring to the "published"semantics, i.e. the docstring. The fact that in R3 `get` also accepts a `get-word!` argument is in contradiction with that docstring.

endo64:
Added to wiki [Differences-between-Red-and-Rebol](https://github.com/red/red/wiki/%5BDOC%5D-Differences-between-Red-and-Rebol#get)

Issue #3832, opened 2019-03-31
[VID] Cannot give multiple flags in VID dialect


endo64:
**Describe the bug**
When using multiple `flags` for a face, first one takes priority and the rest doesn't work.

**To reproduce**

```
>> view [fld: field all-over no-border do [? fld/flags]]
== FLD/FLAGS is a word! value.  all-over
```

**Expected behavior**
`flags` should be `[all-over no-border]`

**Screenshots**

![image](https://user-images.githubusercontent.com/840604/55290425-f1222200-53db-11e9-8b42-eb8e3bec4113.png)

`no-border` didn't get into account.

**Platform version**
```
RED: [ branch: "master" tag: #v0.6.3 ahead: 1317 date: 14-Mar-2019/17:49:29 commit: #43b26838b4a39cd20e06a9b6c2f5c2b873e058b7 ]
PLATFORM: [ name: "Windows 10" OS: 'Windows arch: 'x86-64 version: 10.0.0 build: 17134 ]
```

PR #3833, opened 2019-03-31
FIX: Cannot give multiple flags in VID dialect

endo64:
This PR fixes issue #3832 

```
>> view [fld: field all-over no-border do [? fld/flags]]
== FLD/FLAGS is a word! value.  all-over  ; << should be a block! [all-over no-border]
```

### DIFF
--- a/modules/view/view.red
+++ b/modules/view/view.red
@@ -118,7 +118,7 @@ set-flag: function [
 	value [any-type!]
 ][
 	either flags: face/:facet [
-		if word? flags [flags: reduce [flags]]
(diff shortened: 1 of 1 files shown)


endo64:
I couldn't find a place to put test, below test could be added:

```
--assert equal? (lay: layout [field all-over no-border] lay/pane/1/flags) [all-over no-border]
```

greggirwin:
Per your other question on `layout` tests, that would be great to start, and add the above test as a primer. Thanks!

PR #3834, opened 2019-03-31
FEAT: Added password flag to VID

endo64:
This PR adds `password` flag to VID:

```
view [field password]
```

image

### DIFF
--- a/modules/view/VID.red
+++ b/modules/view/VID.red
@@ -304,6 +304,7 @@ system/view/VID: context [
 				| 'options	  (add-option opts fetch-argument block! spec)
 				| 'loose	  (add-option opts [drag-on: 'down])
 				| 'all-over   (set-flag opts 'flags 'all-over)
+				| 'password   (set-flag opts 'flags 'password)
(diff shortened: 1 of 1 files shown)


greggirwin:
Nice! Does this work on MacOS as well? R2 used `hide` as the word, but I think `password` is less confusing, given that we have `hidden` already. I know it can be used for other purposes, but that's the most common, and `obfuscated` or `redacted` aren't very friendly. ;^)

endo64:
Please note that I didn't add the support for `password` field, the code was already there for Windows and MacOS, I just added the `password` word to VID.

So it should work on MacOS (I didn't test yet):

```
;modules/view/backends/macOS/gui.reds
sym = field [
	class: either bits and FACET_FLAGS_PASSWORD = 0 ["RedTextField"][
		"RedSecureField"
	]
]
```

`hidden` makes the face invisible and very similar word to `hide` may lead to confusion. 
Another solution could be adding `password` as a face, a shortcut just like `h1`.

greggirwin:
Sounds good. Prepare to merge.

PR #3837, opened 2019-04-04
FIX: help error info will be printed to buffer

bitbegin:


greggirwin:
@bitbegin is there an associated ticket that was opened for this? What is the reason for the change?

The changes to `show-function-help` should be fine, but changing `parse-func-spec` means you can't use it directly. While we want a better solution in the future, it can be useful now, but won't show when it encounters a bad spec with this change.

bitbegin:
i want use it some other place, but dont want the `print`.

you still can use `parse-func-spec` directly, as it just print error info to buffer.

bitbegin:
@greggirwin 
>> but won't show when it encounters a bad spec with this change.
it still show the error info, because `print help-string :word`.

greggirwin:
So I haven't gotten back to this. What I mean is, before the change you get:
```
>> help-ctx/func-spec-ctx/parse-func-spec [a [1]]
Looks like we have a bad type spec: [1]
== make object! [
    desc: none
...
```
After the change you get:
```
>> help-ctx/func-spec-ctx/parse-func-spec [a [1]]
== make object! [
    desc: none
...
```
`help-string` won't do any good here, because it doesn't operate on func specs, but funcs themselves. 

It's a bit of a hack as it stands, because the only information you get is the printed output, not an error in the result. If @dockimbel and @qtxie are OK with your change, we can merge this, and deal with it later.

greggirwin:
@qtxie @dockimbel any thoughts on this?

qtxie:
@greggirwin If I undertand the code correctly, you still get the error message printed after the change. @bitbegin Is it used in the VSCode plugin?

bitbegin:
No longer used, `merge` will be good

Issue #3855, opened 2019-04-21
Syntax Error: invalid binary!


jefurry:
**Describe the bug**
Syntax Error: invalid binary! at "2#{0000 1111}".

**Expected behavior**
for Rebol:
```
>> 16#{12 34}
== #{1234}

>> 64#{abcd efgh}
== #{69B71D79F821}

>> 2#{0000 1111}
== #{0F}
```

for Red:
```
>> 16#{12 34}
== #{1234}

>> 64#{abcd efgh}
== #{69B71D79F821}

>> 2#{0000 1111}
*** Syntax Error: invalid binary! at "2#{0000 1111}"
*** Where: do
*** Stack: load
>>
```

**Platform version (please complete the following information)**
```
Red 0.6.4 for macOS built 22-Nov-2018/8:40:38+08:00 commit #755eb94
```

qtxie:
The lexer only allows separate each byte by spaces. That why `2#{0000 1111}` throw an error.

jefurry:
@qtxie For binary, it is intuitive to separate by half-byte.

9214:
Both R2 and R3 support this, so it makes sense to relax lexing rules a bit.

meijeru:
Not trivial, but doable.

Issue #3764, opened 2019-02-01
Empty binary with prefix notation throws an error


rebolek:
**Describe the bug**

When trying to use empty binary with prefix notation - `2#{}`, `64#{}`, ..., Red throws an error:

```
>> 2#{}
*** Syntax Error: invalid integer! at "2#{}"
*** Where: do
*** Stack: load 
>> 64#{}
*** Syntax Error: invalid integer! at "64#{}"
*** Where: do
*** Stack: load
```

@greggirwin reports slightly different error:

```
>> 2#{}
*** Syntax Error: invalid type at "2#{}"
*** Where: do
*** Stack: load
```

According to @hiiamboris `2#{ }` (added space) works.
**To reproduce**

See above.

**Expected behavior**

It's not very useful, but it's valid and should work.

**Platform version (please complete the following information)**
```
>> about
Red 0.6.4 for Linux built 25-Jan-2019/12:19:17+01:00
```

PR #3856, opened 2019-04-21
FIX: issue #3855 (Syntax Error: invalid binary!)

9214:
Red now can handle base-2 `binary!` notation with spaced-out half-bytes:
```red
>> 2#{1101 1110 1010 1101 1011 1110 1110 1111}
== #{DEADBEEF}
```
This makes sense in general (easier to read, allows for more free-form formatting suitable for a problem at hand), and also improves compatibility with R2 and R3.

### DIFF
--- a/environment/lexer.red
+++ b/environment/lexer.red
@@ -552,7 +552,7 @@ system/lexer: context [
 
 		base-2-rule: [
 			"2#{" (type: binary!) [
-				s: any [counted-newline | 8 [#"0" | #"1" ] | ws-no-count | comment-rule] e: #"}"
+				s: any [counted-newline | 4 8 [#"0" | #"1" ] | ws-no-count | comment-rule] e: #"}"
 				| (throw-error [binary! skip s -3])
 			] (base: 2)
 		]
--- a/lexer.r
+++ b/lexer.r
@@ -527,7 +527,7 @@ lexer: context [
 
 	base-2-rule: [
 		"2#{" (type: binary!) [
-			s: any [counted-newline | 8 [#"0" | #"1" ] | ws-no-count | comment-rule]
+			s: any [counted-newline | 4 8 [#"0" | #"1" ] | ws-no-count | comment-rule]
 			e: #"}" (base: 2)
 			| (pos: skip s -3 throw-error)
 		]


greggirwin:
@9214, would you add some tests please? It look like your change will allow not just groups of 4 or 8, but anywhere *from* 4 to 8, which may not be desirable, and we also need to know it won't do something bad if it gets *only* a nibble.

9214:
@greggirwin maybe we should start with desiderata.  Both R2 and R3 follow relaxed rules:
```red
>> 2#{111100 001111 000011 110000}
== #{F0F0F0}
>> 2#{111100001111000011110000}
== #{F0F0F0}
>> 2#{1111 0000 1111 0000 1111 0000}
== #{F0F0F0}
```

Should we restrict size of groups to 4 **and** 8:
```red
>> 2#{1111 00001111 00001111 0000}
== #{F0F0F0}
``` 
or to 4 **or** 8?
 ```red
>> 2#{1111 0000 1111 0000 1111 0000} ; either group all by 4
== #{F0F0F0}
>> 2#{11110000 11110000 11110000} ; or group all by 8, but not 4 and 8 at the same time
== #{F0F0F0}
```

With one nibble both R2 and R2 report invalid `binary!`:
```red
>> 2#{1111}
** Syntax Error: Invalid binary -- 2#{1111}
** Near: (line 1) 2#{1111}
```

In my case error reporting is not strict enough:
```red
>> 2#{1111}
*** Syntax Error: invalid integer! at "2#{1111}"
*** Where: do
*** Stack: load 
```

There's also related ticket https://github.com/red/red/issues/3764

greggirwin:
#3764 seems like a simple lexer bug, and can safely be fixed.

No dangling nibbles. All data must be specified.

Groups of 4 or 8, as long as the total is modulo 8 (again, complete data).

My reasoning is this: The rare cases where there may be value in being completely relaxed are less important than the risk of getting the count wrong in a group, and having to squint to debug.  It's not easy to see how many bits are in a group and be sure. This also provides some protection if data were to be line wrapped during processing (assuming breaks on spaces, and not at any random point). The use case for literal binary data is a bit different than for base-64, but I don't feel the above rules would be too restrictive. It still lets you break things up into "words" for humans to check, but also provides enough constraints to be helpful.

9214:
Alrighty then, I close this PR for the time being.

PR #3870, opened 2019-05-12
FIX: issues #3581, #2772, #1773

hiiamboris:
Fixes #3581, fixes #2772, fixes #1773

This PR implements simple but correct CLI argument quoting/unquoting mechanics:
- on Windows following the CMD.EXE logic (only double quotes)
- on other platforms using single quotes
- **update**: also the "--" command line switch that marks the end of switch processing

**References:**

https://unix.stackexchange.com/a/357932
https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/
https://docs.microsoft.com/en-us/previous-versions//17w5ykft%28v=vs.85%29
https://github.com/red/red/issues/2772#issuecomment-440485682

**Tests**

I have only manually tested this using stuff I uploaded here: https://gitlab.com/hiiamboris/tests-files-3870
I don't have Rebol SDK (of course) so can't rebuild the `red` executable to see if it even builds. `red.r` compiler works for me, as well as compiled executables.
I'd like some Linux power user to skim thru the changed code, just in case.

*And what do we do about `red.exe`? Push it and if any issues arise patch them? Or will somebody with the SDK access try to rebuild it first?*

**Overview**

Argument flow structure according to my findings (maybe deserves a wiki page idk):
1) `red.r` is invoked from (encapped) Rebol with a script name and arguments (OS is determined at runtime)
2) `red.r` receives parsed `options/args` and (invalid - see below `*`) `script/args` from Rebol
3) `red.r` (on Windows only) in [`parse-options`](https://github.com/red/red/blob/8d712e4ba1b16e1b7a339006d1154aebc9cb7ce4/red.r#L683) replaces `script/args` from a WinAPI call `GetCommandLineW` and then replaces also `options/args` by parsing `script/args` with [`split-tokens`](https://github.com/red/red/blob/8d712e4ba1b16e1b7a339006d1154aebc9cb7ce4/red.r#L265)
4) `red.r` joins `options/args` with [`form-args`](https://github.com/red/red/blob/8d712e4ba1b16e1b7a339006d1154aebc9cb7ce4/red.r#L322) into a new command line string. Also [`form-args`](https://github.com/red/red/blob/8d712e4ba1b16e1b7a339006d1154aebc9cb7ce4/red.r#L322) inserts `--catch` into the command line head if it finds this flag before the script name
5) `red.r` passes this string as an argument list to the console executable
[...]

PR #3877, opened 2019-05-15
FIX: issue #3869

### LINKED ISSUE #3869: Internal usage of `call/shell` does not conform to `cmd.exe` specs on Windows
**Describe the bug**

Having a `prog ram.exe` (with a space in it's name) I tried to `call` it with some arguments:
```
>> call/wait/console {"prog ram" "" 1 "2 3"}          ;) this is totally OK
[prog ram] ["" 1 "2 3"]
    -> [] [1] [2 3]
== 0

>> call/wait/console {prog ram "" 1 "2 3"}            ;) NO idea why this works - it should run `prog`
[prog ram] ["" 1 "2 3"]
    -> [] [1] [2 3]
== 0

>> call/wait/shell/console {"prog ram" "" 1 "2 3"}    ;) adding /shell breaks everything
"prog" is not recognized as an internal or external command, operable program or batch file.
== 1
```

**To reproduce**

The problem with the 3rd call is that `/shell` simply [prepends the command line with `cmd /c`](https://github.com/red/red/blob/8d712e4ba1b16e1b7a339006d1154aebc9cb7ce4/runtime/call.reds#L331). Which is wrong. Look:
```
>cmd /c "prog ram.exe" "" 1 "2 3"
"prog" is not recognized as an internal or external command, operable program or batch file.

>cmd /s /c ""prog ram.exe" "" 1 "2 3""
[prog ram.exe] ["" 1 "2 3"]
    -> [] [1] [2 3]
```
The latter example is a proper way to provide a command line to the shell - see https://superuser.com/questions/1213094/how-to-escape-in-cmd-exe-c-parameters
TL;DR (with /S flag):
> first character is
    a quote character and if so, strip the leading character and
    remove the last quote character on the command line, preserving
    any text after the last quote character.

I didn't look into the 2nd call issue though.

**Expected behavior**

2nd call should fail, 3rd should not

```
Red 0.6.4 for Windows built 22-Apr-2019/19:40:30+03:00 commit #9ac4430
```

**`prog ram` source:**
```
Red []
print rejoin ["[" system/options/boot "] [" system/script/args "]"]
prin "    ->" foreach arg system/options/args [prin rejoin [" [" arg "]"]]  prin lf
```

hiiamboris:
Fixes #3869 

Adds `cmd /s /c` to a `/shell` call and wraps the command line in double quotes, fixing this scenario:
```
>> call/wait/shell/console {"prog ram" "" 1 "2 3"}    ;) adding /shell breaks everything
"prog" is not recognized as an internal or external command, operable program or batch file.
== 1
```

Another scenario
```
>> call/wait/console {prog ram "" 1 "2 3"}     ;) NO idea why this works - it should run `prog`
[prog ram] ["" 1 "2 3"]
    -> [] [1] [2 3]
== 0
```
is not Red's fault, but a result of MS [CreateProcessW being smart](https://docs.microsoft.com/en-us/windows/desktop/api/processthreadsapi/nf-processthreadsapi-createprocessw) (how uncommon for those guys):
>  If you are using a long file name that contains a space, use quoted strings to indicate where the file name ends and the arguments begin; otherwise, the file name is ambiguous. For example, consider the string "c:\program files\sub dir\program name". This string can be interpreted in a number of ways. The system tries to interpret the possibilities in the following order:
c:\program.exe c:\program files\sub.exe c:\program files\sub dir\program.exe c:\program files\sub dir\program name.exe

### DIFF
--- a/runtime/call.reds
+++ b/runtime/call.reds
@@ -327,18 +327,20 @@ ext-process: context [
 			
 			either shell? [
 				len: (1 + platform/lstrlen as byte-ptr! cmd) * 2
-				cmdstr: make-c-string (14 + len)
-				copy-memory as byte-ptr! cmdstr as byte-ptr! #u16 "cmd /c " 14
-				copy-memory as byte-ptr! cmdstr + 14 as byte-ptr! cmd len
+				cmdstr: make-c-string (24 + len)
+				copy-memory as byte-ptr! cmdstr as byte-ptr! #u16 {cmd /s /c "} 22
+				copy-memory as byte-ptr! cmdstr + 22 as byte-ptr! cmd len
+				copy-memory as byte-ptr! cmdstr + len - 4 as byte-ptr! #u16 {"} 2
 			][
 				cmdstr: cmd
 			]
 			unless platform/CreateProcessW null cmdstr null null inherit 0 null null :s-inf :p-inf [
 				either 2 = platform/GetLastError [		;-- ERROR_FILE_NOT_FOUND
 					len: (1 + platform/lstrlen as byte-ptr! cmd) * 2
-					cmdstr: make-c-string (14 + len)
-					copy-memory as byte-ptr! cmdstr as byte-ptr! #u16 "cmd /c " 14
-					copy-memory as byte-ptr! cmdstr + 14 as byte-ptr! cmd len
+					cmdstr: make-c-string (24 + len)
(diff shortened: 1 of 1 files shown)


greggirwin:
See also https://github.com/red/red/pull/3870 that has lots of CLI examples. This is something to think about for future test frameworks and CI.

PR #3878, opened 2019-05-15
FIX: patch for commit #8e99d131 (PR #3877)

hiiamboris:
I made a little bit of a mess with the indices in the PR before that only now occurred to me, sorry ☺


greggirwin:
Thanks for the quick fix!

PR #3887, opened 2019-05-21
FIX: issue #3472 (Help text sometimes overflows to the next line)

### LINKED ISSUE #3472: Help text sometimes overflows to the next line
### Expected behavior
There should be a single line for each entry in the help text.
### Actual behavior
The descriptions that are too long to fit are trimmed with ellipsis, but not trimmed quite short enough.
image
The Gui console seems to wrap a little more aggressively because it wraps at word boundaries.
image

I spent some time looking at the code, and I thought the reason was because the width calculated for the descriptive text is based on the other column widths, but doesn't seem to take into account extra spaces between columns (the `_print` method using `form reduce`). I had trouble verifying that, however.
Run `? "series"` in gui or cli console. I experience the issue in both, but several people do not have it. See the discussion from here: https://gitter.im/red/bugs?at=5b4c36611c0f906b145a253a

### Red and platform version
```
RED: [ branch: "master" tag: #v0.6.3 ahead: 791 date: 15-Jul-2018/2:03:36 commit: #8b2db6bb9417d65f5fea3a70c5acdd1fcea3f8dd ]
PLATFORM: [ name: "Windows 10" OS: 'Windows arch: 'x86-64 version: 10.0.0 build: 17134 ]
```

hiiamboris:
Fixes #3472 

Tested on W7 (GUI+CLI) & Xubuntu (more test results are welcome)

Summary:
- `char-width` in gui console is changed to a float! for better arithmetic
- scrollbar is accounted for, albeit using a hardcoded 20 virtual pixels width
- `form-value` has no way to know at what column the value will be output (after 1 or 2 columns of text, or whether columns are of their minimum width or bigger, what separators were used) so I moved the ellipsizing code into `_print` which has a flag now that commands it to truncate the output
- basically every occurrence of `_print [... form-value ...]` became `_print/fit [... form-value ...]`
- improved `view` subsystem presence check (it wasn't working)

Yeah, I got pictures ☻

CLI:

GUI:

### DIFF
--- a/environment/console/GUI/core.red
+++ b/environment/console/GUI/core.red
@@ -235,9 +235,9 @@ object [
 	update-cfg: func [font [object!] cfg [block!] /local sz][
 		box/font: font
 		max-lines: cfg/buffer-lines
-		box/text: "XX"
+		box/text: "XXXXXXXXXX"
 		sz: size-text box
-		char-width: 1 + sz/x / 2
+		char-width: sz/x + 1 * 0.1 						;-- +1 compensates for size-text rounding
 		box/tabs: tab-size * char-width
 		line-h: rich-text/line-height? box 1
 		box/line-spacing: line-h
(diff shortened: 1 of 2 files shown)


greggirwin:
Thanks @hiiamboris. @qtxie if the %core.red change is OK by you, I'll sign off on the %help.red change.

qtxie:
@greggirwin It's OK.👌

PR #3895, opened 2019-05-23
FEAT: adds remove/key for series.

qtxie:


qtxie:
Seems travis timeout.

greggirwin:
I merged this, but should `/skip` also be implied. Right now it isn't. That is, if the `key` part implies a key-value model, it should use `/skip 2` as well.
```
>> blk: [a b c d e f]
== [a b c d e f]
>> remove/key blk 'b  ; I think of 'b as a value, not a key here
== [a d e f]
```

hiiamboris:
`remove/key/part blk 'b 2` ?

greggirwin:
Don't need `/part` there, but then you're also adding `/skip` to `remove`, and I'd end up writing a wrapper for that line as I have in the past. The redundancy of having to say `2` twice is ugly and error prone.

Issue #2279, opened 2016-10-20
WISH foreach should be allowed on maps


meijeru:
It is quite common to have to step through all elements of a map. Currently, one has to write:

```
foreach key keys-of map-value [
    element: select map-value key ; or element: map-value/(key)
    <some processing using key and element>
]
```

Wouldn't it be nice to be able to write:

```
foreach [key element] map-value [<some processing using key and element>]
```

The first argument of foreach should then always be a block of two words, but this is easy to enforce. Blocks of words are already allowed in that position.

hiiamboris:
Would've been also great if foreach implemented that iteration on R/S level, without having to look up every item.

greggirwin:
There are always tradeoffs and constraints. You may have seen a new map PR show up, which will change the interface in some important ways, but making maps act like series is not one of them. During the design process, that came up a lot, and there's no perfect answer. It was also heavily discussed about how much a user should know or care about the implementation, and how best to treat blocks as if they were maps. That is, `map` is a symbolic construct and interface, which requires extra work when using blocks as key-value structures. You have to use `/part/skip 2 2` consistently, which is best done in wrapper funcs.

@dockimbel @qtxie @giesse and myself spent a lot of time talking about this, and the next version of the `map!` interface may not be the final one, related to wishes like this. In the meantime, how painful (and slow) is it to do something like this:
```
for-map: function [
	"Iterate over a map, making each `key` and `value` available to the body."
	data [map! series!]
	body [block!]
][
	either series? data [
		foreach [key value] data [do body]
	][
		foreach key keys-of data [
			value: data/:key
			do bind body 'key
		]
	]
]
blk: [a: 1 b: 2 c: 3]
for-map m [print [key value]]
m: #(a: 11 b: 22 c: 33)
for-map m [print [key value]]
```
I think we still need function attributes to making writing control funcs easy, WRT `break/continue`, which is also important for HOFs. We also can't profile the above versus and R/S version without building that, but we can see how the mezz version(s) perform against a clock. They may not work for everything, but my guess is that they'll be fast enough for all but the largest datasets.

greggirwin:
@meijeru if https://github.com/red/red/pull/3896 looks good to you, please close this ticket.

greggirwin:
Or not. Seems it was closed, not merged.

greggirwin:
#3899 is the new PR for this.

9214:
@meijeru if you are happy with this change, we'd like to close the issue.

[...]

PR #3899, opened 2019-05-24
Foreach map

qtxie:
```
m: #(a 1 b 2 c 3)
foreach [k v] m [print [k v]]
```
Will output: 
```
a 1
b 2
c 3
```

### DIFF
--- a/environment/natives.red
+++ b/environment/natives.red
@@ -91,7 +91,7 @@ forever: make native! [[
 foreach: make native! [[
 		"Evaluates body for each value in a series"
 		'word  [word! block!]   "Word, or words, to set on each iteration"
-		series [series!]
+		series [series! map!]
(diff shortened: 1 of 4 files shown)


greggirwin:
Allowing mod 2 sizes for iterator words is cool, but also makes me a little uneasy. It comes back to whether we want to guarantee key ordering. I suppose there are algorithms that could use it, and be safe because they're independent of order (commutative). @dockimbel, thoughts?

dockimbel:
We will stick with an argument block of strictly 2 words for maps.

greggirwin:
Will merge later today, if nobody says otherwise. And I'll find the ticket related to this.

greggirwin:
https://github.com/red/red/issues/2279 is the old wish ticket.

Issue #847, opened 2014-06-07
NaN comparision doesn't follow IEEE754 spec.


qtxie:
Using the code below to do the test.

```
Red []

foo-test: routine [][
    inf: 1e308 + 1e308
    nan: 0.0 * inf
    print-line inf
    print-line nan
    print-line ["inf > nan: " inf > nan]
    print-line ["inf < nan: " inf < nan]
    print-line ["inf <> nan: " inf <> nan]
    print-line ["inf = nan: " inf = nan]
]

foo-test
```

Results on Windows

```
1.#INF
-1.#IND
inf > nan: false
inf < nan: true                 ;@@ should be false
inf <> nan: false               ;@@ should be true
inf = nan: true                 ;@@ should be false
```

Results on ARM
On ARM (QEMU VM), we find that it doesn't generate INF and NaN, don't know why.

```
1.060479830104e-314             ;@@ should be INF ?
1.0607388627933e-314            ;@@ should be NaN ?
inf > nan: true
inf < nan: false
inf <> nan: true
inf = nan: false
```

```
```

qtxie:
@dockimbel We already supported plan 2 now.

hiiamboris:
look, more bugs came up (not `#inf`, normal numbers):
```
>> 1.0 < 1.#nan
== true
>> 1.#nan > 1.0
== false
>> 1.0 > 1.#nan
== false
>> 1.#nan < 1.0
== true
```
(all should yield `false`)

qtxie:
> look, more bugs came up (not `#inf`, normal numbers):
> 
> ```
> >> 1.0 < 1.#nan
> == true
> >> 1.#nan > 1.0
> == false
> >> 1.0 > 1.#nan
> == false
> >> 1.#nan < 1.0
> == true
> ```
> 
> (all should yield `false`)
> 
> ```
> -----------RED & PLATFORM VERSION----------- 
> RED: [ branch: "master" tag: #v0.6.3 ahead: 703 date: 10-May-2018/3:19:55 commit: #7c5f402062ac67fbd6078592ca477bbe91a56f92 ]
> PLATFORM: [ name: "Windows 7 Service Pack 1" OS: 'Windows arch: 'x86-64 version: 6.1.1 build: 7601 ]
> --------------------------------------------
> ```

No more bugs. This issue has never been fixed.

dockimbel:
```
>> 1.#nan >= 1.#nan
== true
>> 1.#nan <= 1.#nan
== true
[...]

Issue #810, opened 2014-05-08
CASE return value improperly


qtxie:
```
Red/System []
res: case [
    true [1 = 0]
]
print-line res              ;-- true, should be false.
```

but the code behaviour is correct as below.

```
Red/System []

res: case [
    true [0 = 1]
]
print-line res              ;-- false, it's correct.
```

rebolek:
Checked with latest version and I get `true` as output.

hiiamboris:
Still unfixed in 0.6.4

endo64:
Works correct if there is a assignment inside the case block:

```
Red/System []
r: false  ; or r: true doesn't matter
res: case [
    true [r: 0 = 1]  ; if remove "r:" then res will be true incorrectly
]
print-line r
print-line res

; >> false
; >> false
```

PR #3902, opened 2019-05-28
IEEE-754 compliance in float comparison operators

hiiamboris:
Fixes issues #847 and #2567 for IA-32 emitter only. Probably should be left open (with a reminder comment) until ARM emitter is also fixed.

It's a fundamental change that affects virtually everything. @dockimbel should definitely review this.


hiiamboris:
I have also an optimization developed:
```
			;-- optimization: JNx jumps fail on NaNs anyways, Jx - succeed; no need for parity tests
			if all [
				parity									;-- with NaN: CF=PF=ZF=1
				either unord=true [
					;-- JP can be left off if Jcc always succeeds on P=1: JC(<), JZ(=), JBE(<=)
					find [< = <=]  op
				][
					;-- JP can be left off if Jcc always fails on P=1: JNC(>=), JNZ(<>), JA(>)
					;-- FIXME: `>=` should also be in this list, but it doesn't work in case return tests - why??
					find [> <>] op
				]
			] [parity: no]
```
It is to be inserted [here](https://github.com/red/red/blob/694d183ac1af0fcdb5c33cf8dbb8d07cf23c0d5b/system/targets/IA-32.r#L1248) but for some reason if I'm inserting a `>=` into the second op list these tests start to fail:
```
b: case [true [one >= nan]]      if b [print-line "FAILED: case return one >= nan"]
b: case [true [nan >= one]]      if b [print-line "FAILED: case return nan >= one"]
```
Considering this, I'm not sure about merging this fragment until the the problem source is revealed. Currently, I haven't the slightest clue...

hiiamboris:
OK. The source of `case` problem is https://github.com/red/red/issues/810. I'm just going to comment the `case` tests for now as they're not doing any good.

system/targets/IA-32.r
```diff
@@ -1311,30 +1311,129 @@ make-profilable make target-class [
 		code [binary!]
 		op [word! block! logic! none!]
 		offset [integer! none!]
+		parity [none! logic!] "yes = also emit parity check for unordered (NaN) comparison"
 		/back?
-		/local size imm8? opcode jmp
+		/local size jump jump+ jmpofs jxx unord=true
```
dockimbel, line 1316: `unord=true`: I give you a point for the creativity, but it's really a confusing literal word as it looks like a conditional expression. Better avoid that. Just use `unord?` or `unordered?`.

system/targets/IA-32.r
```diff
@@ -1311,30 +1311,129 @@ make-profilable make target-class [
...
+
+		jump: copy #{}									;-- resulting binary
+		jump+: func [bin] [append jump bin]
+
+		set-unord-branch: func [op] [					;-- unordered jump destination according to IEEE-754
+			unord=true: op = first [<>]					;-- NaN <> x = true, otherwise NaN op x = false
+			op
+		]
+		jmpofs: does [size * pick [-1 1] yes = back?]	;-- size to signed offset converter
+		jxx: func [
```
dockimbel, line 1329: The added code inlines many new functions inside `emit-branch`. This means that those functions will be reconstructed on every call to `emit-branch` (thousands of calls during a typical compilation job), which will slow down the compilation process (slower `emit-branch` execution and higher pressure on the GC). There is really no need for such slowdown, just avoid inlining functions into functions in any piece of critical code (some of them could be inlined as lazily called blocks).

dockimbel:
The changes made into the IA-32 backend are missing for the ARM backend. In the current state, it's not mergeable until ARM is also covered. Given how specialized is this PR, I guess only @hiiamboris can do that job properly.

hiiamboris:
I don't have an ARM Red setup to test on ;)

rebolek:
@hiiamboris I can setup a spare rPi for you and let you SSH to it :)

hiiamboris:
@rebolek Thanks, that'll be welcome ;) If you can do that somewhere in Oct-Nov, I'll finish the PR.

rebolek:
@hiiamboris Ok, I'll set up the machine and let you know. I leave it running so you can access it at any time.

greggirwin:
@hiiamboris we can provide ARM hardware for you to test against locally. Confirm with @dockimbel and @ameridroid what they think would be best.

hiiamboris:
I would prefer a remote access. ;)

rebolek:
@hiiamboris sorry, I almost forgot about it, I got the machine ready, I will create you an account and let you know. I just need your public key for SSH access.

hiiamboris:
@rebolek will this do?
```
---- BEGIN SSH2 PUBLIC KEY ----
AAAAB3NzaC1yc2EAAAADAQABAAABgQDZOlYdJG+oqDx4HZl7gdEYddBLhWHSyOzW1nm8C6i8
+v/Yw8lNEBzGGS6mUwvSYSA/ICr1KclxK4MEwPVXlFQQY2NkhqSVnwlfvECJVn7Yexqsor63
JgRC1p3Bhd0fPnmubUC2k0olnBUtlhUp1sRDuhpwncufnagKJbgMnZE1CGE5qRqkxAItCite
20w85aUg45///0ik6a5M1AKfFdMd5nvOHB56QZ4mszMk+yltuTGu9rEG+3KzqEVVKYfBmNB6
TDHHJ3N+4Z8UY+ZdqDvGMLikpGdHFnA4QIjMzGVno/qloAjNPIgB8dEllrtU1aXtO6Qu2nCu
Rvz5mxNFrP+XqbXLBsQ4VwioUMQ1XKJR6H5oFOU/Iy1gzJt95GYmQ9W0JMMsXu1hA5EOpEKr
0dFVmyJSv8YEHyYDUIE05va9C23Y8CL+n9lPD3ytstf9YYCqsxT8MwPzKHjdzWspmqV6pk7j
ytlKFu+5KXNFwyAvKizIyGmZlhMOxyjOvzVD1cE=
---- END SSH2 PUBLIC KEY ----
```

system/targets/IA-32.r
hiiamboris: After refreshing my memory a bit, I can't agree ;)

The meaning of `unord=true` name is this: "unordered comparison (read: one of operands is a NaN) should jump into a TRUE branch" (which naturally happens only when operator is `<>`, but since emitter supports negation of the conditional expression, this flag often gets negated as well and then everything but `<>` jumps to TRUE).

So, renaming it to `unord?` will be misleading IMO. It does not reflect the "unordered or not" state, because it does not and can not know the values operands will take at run time. Ideal name would be `unordered-jumps-to-true?`, but that's way too long :D

system/targets/IA-32.r
hiiamboris: Will be fixed in the new PR.

hiiamboris:
Resurrected as https://github.com/red/red/pull/4717

system/targets/IA-32.r
greggirwin: `unordered-jumps-to-true?` may be long, but it's clear. You already abbreviate so why not `unord-jumps-to-true?`, or `unord-jmp-true?`

greggirwin: If it only happens for NaN (not also INF), maybe something like `nan-jump-true?`

greggirwin: `jump-true?` with a comment, per Doc's suggestion.

hiiamboris: Renamed to `unord-jumps-to-true?`. Would this be OK? ;)

Issue #3004, opened 2017-08-28
WISH: /SKIP refinement for REVERSE


9214:
I wish `reverse` could treat series as a bunch of fixed-size records:
```Red
>> reverse/skip [a 1 b 2] 2
== [b 2 a 1]
```
perhaps `/by` is a better name?

meijeru:
`/skip`  is appropriate: it tells what it does, in the same way as `sort` and all the others that have such a refinement

greggirwin:
@toomasv posted a mezz version in gitter chat I think. Though having it as standard would be better. I don't have one in my old toolbox, but think I needed this a couple times through the years.

dockimbel:
Agreed. We should also add the missing `/part` option, which is present in Rebol.

toomasv:
Here is my version:
```
reverse: func [
    series /skip size /part length /local i tuple
][
	switch type?/word series [
		pair! [return to-pair reduce [series/2 series/1]]
		tuple! [tuple: clear [] repeat i length? series [append tuple pick series i] series: copy tuple tuple: yes]
	]
    size: any [size 1]
    either 0 = ((length? series) % size) [
        length: any [length length? series]
        i: 1
        loop (length / size - 1) [
            i: i + size
            series: at head series i
            insert head series take/part series size
        ] 
        series: head series
    ][cause-error 'user 'message ["Wrong length of series or skip!"]]
	either tuple [to-tuple series][series]
]
```

9214:
I took the role of self-serving wish fairy and granted myself what I asked for 3 years ago 🧚‍♂️ Hope that's OK.

PR #3907, opened 2019-05-30
Issue 2586 fix

### LINKED ISSUE #2586: Exclude percent! from trig functions
Today the trig funcs use `number!`, which includes `percent!`. Change to `integer! float!` in their type signatures. Percent support is not by design.

hiiamboris:
Fixes #2586


greggirwin:
Hmmm. Appveyor failed, but not due to a timeout.

greggirwin:
Interpreter test failure.

hiiamboris:
I cannot reproduce the test failure on W7 PC & W10 VM. Can't help it ):

greggirwin:
Post to Gitter, and see if someone on Linux is willing to test.

hiiamboris:
But it's the AppVeyor failing. On Windows Server 2012 I think.

greggirwin:
Rats. I thought that one was Linux.

dockimbel:
@hiiamboris You have excluded `float!` from the type spec of arguments to many action's refinements, but that is a regression. Both Rebol and Red allow decimal (float) values for `/part` or `/seek` refinements. Such values are internally rounded (floored) to integer.

greggirwin:
@dockimbel though it's technically a regression, is `float!` support for `/part` a useful feature? I don't recall ever using it (anecdotal), but it also seems like it could lead to gaps in processing, unless you also round and track separately. e.g. some chunking loop ends up exactly on a whole number after chunks that aren't.

hiiamboris:
Can I get a single example of an action (in Red or Rebol) that successfully accepts (no error thrown) a float! part argument and returns an expected result?

9214:
@hiiamboris would you like to follow through with this one? I propose to implement trig changes only (as was requested in the original ticket), and factor-out `action!` changes into a separate PR, for design discussions (e.g. in the light of recent changes to iterators that can now accept `float!` counters).

hiiamboris:
I'll see what I can do.

hiiamboris:
I also think actions changes will require a bit R/S work. And that should be done *after* the #4097 merge. `/part`, `/dup`, `/skip`, `/seek` and `sort/compare <number>` are possible targets.

9214:
@hiiamboris does your PR contains anything related to `/skip` semantics? I'm in the process of granting #3004 and bottomed down on `sort/skip` behavior as a reasonable default.

hiiamboris:
No, I just *removed* percent and float. But if we're *adding* float, it should be done everywhere, not in random places.

9214:
@hiiamboris sorry, I meant `/skip` in #4097 and proposed series model, not this PR.

Issue #3905, opened 2019-05-29
Restrict `compose` to block arg


greggirwin:
**Describe the bug**

`Compose` allows any type in the spec, but the body only supports blocks. This was initially done to match R3, which acts as a pass-thru for non-block args, but Red behaves differently.

**To reproduce**
>> compose ""
*** Script Error: expected block! not string!
*** Where: compose
*** Stack:  

>> compose 1
*** Script Error: expected block! not integer!
*** Where: compose
*** Stack:  

**Expected behavior**

Should enforce block arg in spec.

**Platform version (please complete the following information)**

All

endo64:
@greggirwin I think `FIX: ` in the title should go to a commit message :)

greggirwin:
Fixed.

dockimbel:
@greggirwin Is there anything else to do for this ticket or can it be closed now?

greggirwin:
Should be good to go.

PR #3908, opened 2019-05-30
FIX: #3905 Restrict `compose` to block arg

greggirwin:


dockimbel:
@greggirwin If all tests are passing, please merge this PR.

PR #3922, opened 2019-06-22
FEAT: adds support for releasing IMAGE! in GC.

qtxie:


qtxie:
I'll close it. There are better ways to do it.

PR #3933, opened 2019-07-05
Gtk: fix cairo_font_options_destroy issue

rcqls:
* Fix cairo_font_options_destroy issue 
* a first attempt on `on-resize` event for window widget using gtk `configure-event` event.

### DIFF
--- a/modules/view/backends/gtk3/events.reds
+++ b/modules/view/backends/gtk3/events.reds
@@ -685,7 +685,8 @@ respond-window-add: func [
 	on-type: 0
 	if respond-event?  actors "on-close" [on-type: on-type or ON_CLOSE]
 	if respond-event?  actors "on-move" [on-type: on-type or ON_MOVE] if respond-event?  actors "on-moving" [on-type: on-type or ON_MOVING]
-	if respond-event?  actors "on-size" [on-type: on-type or ON_SIZE] if respond-event?  actors "on-sizing" [on-type: on-type or ON_SIZING]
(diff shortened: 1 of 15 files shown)


modules/view/backends/gtk3/events.reds
```diff
@@ -933,8 +934,9 @@ connect-widget-events: function [
 		sym = window [
 			;; DEBUG: if debug-connect? DEBUG_CONNECT_WIDGET [print ["Add window delete-event " lf]]
 			gobj_signal_connect(widget "delete-event" :window-delete-event null)
+			;BUG (make `vid.red` failing): gtk_widget_add_events widget GDK_STRUCTURE_MASK
+			gobj_signal_connect(widget "configure-event" :window-configure-event null)
 			;BUG (make `vid.red` failing):gtk_widget_add_events widget GDK_STRUCTURE_MASK
```
dumblob: Nitpick: this line seems superfluous then :wink: (it's the same as the one 2 lines above).

Btw. thanks for developing and maintaining the GTK backend - it's one of the biggest and most complex backends for Red. Great work!

rcqls: You're right since it is as you guess a bad copy-paste that I'll remove in a next PR! These pieces of code are a very early development on `on-resize` actor for window used in particular by @loziniak. At the moment, there is plenty of comments that need to be cleaned  😀. BTW thanks for your kind words!

qtxie:
Thanks for the fixing. 👍

rcqls:
Your welcome! Now we need to fix image that does not show anymore!

PR #3941, opened 2019-07-22
Gtk: Fix input of gui-console for characters & and < by escaping them since rich-text is XML based

rcqls:
Add: read-clipboard and write-clipboard completed for Linux based on gtk_clipboard.

### DIFF
--- a/modules/view/backends/gtk3/gtk.reds
+++ b/modules/view/backends/gtk3/gtk.reds
@@ -617,6 +617,11 @@ GPtrArray!: alias struct! [
 			data		[int-ptr!]
 			return: 	[integer!]
 		]
+		g_markup_escape_text: "g_markup_escape_text" [
+			text 		[c-string!]
+			len 		[integer!]
+			return:		[c-string!]
+		]
 	;; ]
 	;; LIBGDK-file cdecl [
 		gdk_screen_width: "gdk_screen_width" [
--- a/modules/view/backends/gtk3/gtk.reds
+++ b/modules/view/backends/gtk3/gtk.reds
@@ -667,6 +667,37 @@ GPtrArray!: alias struct! [
 			display 	[handle!]
 			return: 	[handle!]
 		]
+		gtk_clipboard_get: "gtk_clipboard_get" [
+			atom 		[handle!]
+			return: 	[handle!]
+		]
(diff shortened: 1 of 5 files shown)


modules/view/backends/gtk3/gtk.reds
```diff
@@ -662,6 +667,37 @@ GPtrArray!: alias struct! [
 			display 	[handle!]
 			return: 	[handle!]
 		]
+		gtk_clipboard_get: "gtk_clipboard_get" [
+			atom 		[handle!]
+			return: 	[handle!]
+		]
+		gtk_clipboard_set_text: "gtk_clipboard_set_text" [
```
qtxie: Is it used in GTK backend? It's imported in clipboard.reds too.

rcqls: It is put in both gtk.reds and clipboard.reds

qtxie: Why? if it's only needed in clipboard.reds, import it once is enough.

rcqls: I put it first only in gtk.reds (which is IIMHO, the right file) but it did not compile. I added then in clipboard.reds. Also I used the gtk.reds as is as a gtk-plugins for red/system.

rcqls:
Added to the PR some fix about visible issue for tips.

bitbegin:
https://www.uninformativ.de/blog/postings/2017-04-02/0/POSTING-en.html
we can use x windows clipboard

rcqls:
@bitbegin Tha's was also an option. I checked `libxcb.so` but since we already use the GTK backend, I thought it was easier to directly use it.

bitbegin:
https://gist.github.com/bluecube/9835979 
yes, this's very complicated。

modules/view/backends/gtk3/gtk.reds
qtxie: After those functions are imported in clipboard.reds, you can use them in the GTK backend. Just add the context prefix when calling it. e.g. `clipboard/gtk_clipboard_set_text ...`.

rcqls: Thanks for this explanation!

PR #3943, opened 2019-07-24
Gtk: fix visible? management

rcqls:
hide-invisible inside do-events was not enough to deal with `visible?` facet. In gui-console opening about or settings weirdly makes tips visible again. This PR fix this issue.

### DIFF
--- a/environment/console/GUI/core.red
+++ b/environment/console/GUI/core.red
@@ -803,7 +803,7 @@ object [
 			debug-print ["char: -"  char "- "  
 			switch/default char [
 				#"^M"	["M"] #"^H"	["H"] #"^-"	["-"]
-				left	["left"] right	["right"] up		["up"] down	["down"]
+				left	["left"] right	["right"] up ["up"] down ["down"]
 				insert	["insert"] delete	["delete"]
 				#"^A" home	["home"] #"^E" end	["end"] #"^C"	["C"] #"^V"	["V"]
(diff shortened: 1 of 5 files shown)


qtxie:
Current implementation shows all the faces, then hides the invisible ones. Is there a way to make the face invisible before showing it?

rcqls:
I tried to find a solution longtime ago  when testing view-test.red? The only solution I have found is to save the invisible widgets at the creation in OS-make-view. Of course if you find a better solution it is welcome but I spent a very longtime without finding surprisingly a better solution.

rcqls:
This fix solves the issue about run-file not working at all since request-file returned none in `gui-console` but not inside `console`.

rcqls:
After trying to reimplement gtk_widget_show_all by taking into account `visible?` facets, I found that the best solution is to deal with `visible?` facets after but by using the view tree of widgets. `gtk_widget_show_all` is more difficult to reimplement in red since one has to deal with a lot of widgets not directly accessible from the view tree.

rcqls:
Tab key was not activated! Now gui-console has completion

qtxie:
@bitbegin Do you have any idea about the `visible?` management?

bitbegin:
agree with "show all, then hide widget", but maybe we can use `gtk_widget_hide` to hide the widget.  Anyway, current solution is good enough.

PR #3946, opened 2019-07-26
Gtk: Workaround for caret issue when first shown in gui-console

rcqls:
Fix caret issue and also offset for window used in console-cfg.red

### DIFF
--- a/environment/console/GUI/core.red
+++ b/environment/console/GUI/core.red
@@ -247,6 +247,7 @@ object [
 		if font/color [change theme/foreground font/color]
 		#if config/OS = 'Linux [
 			; since in Linux/GTK, no on-resizing when the window is first drawn
+			gui-console-ctx/console/size: cfg/win-size
 			resize gui-console-ctx/console/size
 		]
 		adjust-console-size gui-console-ctx/console/size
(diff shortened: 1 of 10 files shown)


rcqls:
Hope the second attempt is a better solution for the caret issue. As @qtxie mentioned it, Linux/GTK does not fire a resize event when the window is first shown.

rcqls:
Added as a comment the solution consisting in firing a `configure-event` (to resize window when shown). The commented code is in `OS-show-window` and it is supposed to avoid the Linux code added in `update-cfg` inside `core.red` (related gui-console). But currently, this solution with `update-cfg` is far from being more stable compared with the resizing one.

rcqls:
Rich-text event delegation was broken (see makedoc cooked by @toomasv) after introduction of scrolling for rich-text. Introduction of real-widget fixes this issue.

rcqls:
Important change to have settings (font and colors) working  in gui-console working .

rcqls:
This last commit is a rewriting of g_markup_escape_text in red/system to only escape "<" and "&" characters that are required to have gui-console working for these characters.
As a direct consequence, selection is useable whenever these characters are not in the text to select. This would be fix when updating `caret-to-offset` to take into account these characters. NB: there is a small bug in the echoing of selection but it is useable.

rcqls:
With this commit no more bug on the echoing when selecting. I have also activated syntax highlighting for fun....

rcqls:
@qtxie Thanks!

Issue #3662, opened 2018-12-14
[Lexer] Hex literals inconsistencies and bugs


hiiamboris:
**Describe the bug**
R/S and Red lexers treat these differently, and Red compiler is buggy.

**To reproduce**

**In the interpreter:**
```
>> print [10h 100h 1000h 10000h 100000h 1000000h]
16 256 4096 65536 1048576 16777216
```
That was totally OK!

**Compiled:**
```
Red []
print [10h 100h 1000h 10000h 100000h 1000000h]
```
Prints:
```
16 0 4096 0 1048576 0
```
what???

**R/S:**
```
Red/System []
print-line 100h
```
Prints:
```
Compiling to native code...
*** Loading Error: invalid hex literal: 100h
*** in file: %/D/devel/red/red-src/tests/1.reds
*** at line: 3
```

**Red routine:**
```
Red []
r: routine [][print-line 100h]
r
```
Prints:
```
0
```

**Expected behavior**
All scenarios should follow the interpreted behavior.

**Platform version (please complete the following information)**
```
Red 0.6.4 for Windows built 7-Dec-2018/8:41:53+03:00 commit #1bdbdc2
```

endo64:
Confirmed on Win10, `Red 0.6.4 for Windows built 6-Dec-2018/4:52:41+03:00 commit #08e9d17`

hiiamboris:
One more bit: R/S accepts 2-, 4- and 8-digit hex literals. Nothing in between, not even FFFFFFh (6-digit) and such.

9214:
```red
>> 1h
== 1
>> 10000000h
== 268435456
>> 100000000h
== 0
```

PR #3947, opened 2019-07-26
FIX: issue #3662 (allow to load 1-8 hexa to integer!)

bitbegin:
treated `hexa` same for RS and Red

### DIFF
--- a/environment/lexer.red
+++ b/environment/lexer.red
@@ -694,7 +694,7 @@ system/lexer: context [
 		sticky-word-rule: [								;-- protect from sticky words typos
 			ahead [integer-end | ws-no-count | end | (throw-error [type s])]
 		]
-		hexa-rule: [2 8 hexa e: #"h" ahead [integer-end | ws-no-count | end]]
+		hexa-rule: [1 8 hexa e: #"h" ahead [integer-end | ws-no-count | end]]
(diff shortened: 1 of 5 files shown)


hiiamboris:
Hexadecimals of length 1 were likely forbidden by design, to let `Ah` and `Eh` be words :)

bitbegin:
should not be a reason, as `AEh` also can be treated as words

greggirwin:
Hex literals are an interesting issue, and problematic for me. I completely understand the desire for them, but they create a lexical ambiguity I don't like at all, and a case sensitivity aspect as well. Even if we never talk about DEAD BEEF in a CAFE, or a FEE the FED might charge, we shouldn't be DEAF to ADA accessibility. Not to mention possibly security exploits, which are BAD.

Do we *really* want these in Red? R/S makes more sense, and doesn't have word types.

dockimbel:
The current hex syntax is ambiguous. Restricting it to 2, 4 and 8 characters only was done on purpose to limit ambiguous cases. As discussed in other tickets, this syntax was meant to be temporary and be replaced by a proper one. So, given that #3662 was fixed by another PR and that we don't want to extend the current hex syntax, I'll close this PR.

PR #3952, opened 2019-07-29
FEAT: process `make/to/mold` handle! more better

bitbegin:


runtime/datatypes/handle.reds
```diff
@@ -53,6 +53,36 @@ handle: context [
...
+
+	to: func [
+		proto 	[red-value!]							;-- overwrite this slot with result
+		spec	[red-value!]
+		type	[integer!]
+		return: [red-value!]
+		/local
+			h	[red-value!]
+	][
+		#if debug? = yes [if verbose > 0 [print-line "to/make"]]
```
greggirwin: Should this line print `handle/to` rather than `to/make`?

qtxie:
The handle! value is hidden on purpose. The limitation of `make` and `to` are on purpose too.

dockimbel:
`handle!` type exists to make Red able to pass around an opaque pointer for low-level needs, without letting the user modify it at Red level. Those changes are breaking the purpose of `handle!` type, so they will be rejected.

bitbegin:
this PR just for test, as it failed the ci, related to #3956

PR #3960, opened 2019-08-02
FIX: gui-console crashed on macos

bitbegin:


bitbegin:
sometime on Macos:

`as red-object! (as cell! view + ivar_getOffset ivar)` <> `as red-object! view + ivar_getOffset ivar`

PR #3969, opened 2019-08-06
FIX: support resize image! for Macos

bitbegin:


bitbegin:
test case:

```
Red [
	Needs:	'View
]

resize: routine [
	img		[image!]
	pair	[pair!]
][
	stack/set-last as cell! image/resize img pair/x pair/y
]

img: load %button.gif
size: img/size
l: size / 2
h: size * 2

limg: resize img l
himg: resize img h

view [
	image img return
	image limg return
	image himg return
]
```

PR #3973, opened 2019-08-07
Gtk: Save image for jpeg png and bmp

rcqls:


qtxie:
Nice feature! Thanks!

PR #3981, opened 2019-08-10
FEAT: CSV codec added

rebolek:
CSV codec with `load-csv` and `to-csv` functions.
See https://github.com/red/red/wiki/CSV-codec for detailed documentation.

### DIFF
--- a/boot.red
+++ b/boot.red
@@ -30,6 +30,7 @@ Red [
 	#include %environment/codecs/bmp.red
 	#include %environment/codecs/gif.red
 	#include %environment/codecs/json.red
+	#include %environment/codecs/csv.red
 
 	#include %environment/reactivity.red				;-- requires SET intrinsic
 	#include %utils/preprocessor.r
--- a/build/includes.r
+++ b/build/includes.r
@@ -37,6 +37,7 @@ write %build/bin/sources.r set-cache [
 			%gif.red
 			%jpeg.red
 			%png.red
+			%csv.red
 			%json.red
 			%json/ [
 				%common.red
--- a/environment/codecs/csv.red
+++ b/environment/codecs/csv.red
@@ -309,7 +309,7 @@ context [
 					repeat index length [
 						value/(header/:index): line/:index
 					]
-					append output copy value
+					append output value
 				][
 					either flat [
 						append output copy line
--- a/environment/codecs/csv.red
+++ b/environment/codecs/csv.red
@@ -1,0 +1,413 @@
+Red [
+	Title: "CSV codec"
+	Author: "Boleslav Březovský"
+	Rights:  "Copyright (C) 2015-2019 Red Foundation. All rights reserved."
+	License: {
+		Distributed under the Boost Software License, Version 1.0.
+		See https://github.com/red/red/blob/master/BSL-License.txt
+	}
+	Resources: [
+		https://tools.ietf.org/html/rfc4180
+		https://www.python.org/dev/peps/pep-0305/
+	]
+	Documentation: https://github.com/red/red/wiki/CSV-codec 
+]
+
+put system/codecs 'csv context [
+	Title:     "CSV codec"
+	Name:      'CSV
+	Mime-Type: [text/csv]
+	Suffixes:  [%.csv]
+	encode: func [data [any-type!] where [file! url! none!]] [
+		to-csv data
+	]
+	decode: func [text [string! binary! file!]] [
+		if file? text [text: read text]
+		if binary? text [text: to string! text]
+		load-csv text
+	]
+]
+
+context [
+	; -- state variables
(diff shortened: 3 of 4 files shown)


dockimbel:
It seems that there is a misunderstanding about codecs. They exist to provide a standard API that is easy to remember for the user: `load/as` and `save/as` (with auto-detection of the format when possible). If each codec provides its own specific API, then the goal is missed (and global context gets polluted with many new functions instead of the basic two ones.).

Moreover, as they are part of the Red runtime, codecs should provide just the basic feature expected from them, basically, load/save a given external data format. Each codec should strive to implement the required features with as less code as possible, in order to prevent a significant increase of the final Red binary.

I haven't look in details into the CSV specification, but I suspect it should be possible to make a good implementation in about a page of code.

greggirwin:
The standard API is there, but the inner funcs used by it, with advanced options, are also available.

@rebolek can you extract the bare-bones load/save part into a new branch or test area, and see how much difference in makes in the binary size?

Yes, we want to avoid excess baggage, but also one-up the competition. After JSON (and maybe before because of spreadsheet use, which is why I supported the auto column name feature), CSV will be the most widely used data codec, so I'd like to have a complete, official implementation. It is also one of the most flexible formats, hence the options which I think are useful. 

What other codecs do you want included as standard @dockimbel? I only have Zip that I think should be standard. Others like YAML, INI, TAR, EDN, TOML, msgpack, etc. I see as optional modules.

My plan was also to note that it's experimental, for comment and feedback, where we could also include a note that the advanced features may move to an optional module in future releases to keep the binary size down. If we keep the same names, it shouldn't break any code when that happens.

greggirwin:
The first option I'd choose to give up is `/flat`, which will make data harder to work with, requiring helpers for clarity, but could make a difference if you have a lot of rows. But it's also dependent on your access patterns, and if you use `/flat` but remove data, it could hurt performance.

greggirwin:
Chatting with @rebolek about `/flat`. He points out that it lets you do ` foreach [col1 col2 col3] series [...]` processing. Removing flat support will save about 20 lines, mostly in `encode-flat`, and will remove a refinement from each main func.

I've never used a column model, but that's because I never had a loader for it I think. I know of at least one project where I could have, maybe should have, used it.

So let me ask it this way @dockimbel: To make it really simple, what single data model do you want for loaded data? Rows, Records, or Columns? If we have only one, is *has to be* `rows`, right? Because if you have data without a header, you can't do records, and using columns will be backwards from how I think most people view CSV data. Data scientists and fin-tech people may think in columns of course.

If you agree, what does using that data look like now? e.g., given a common case of a CSV with headers `[first-name last-name street city state zip]`

greggirwin:
An alternative, @dockimbel, to exposing inner funcs, would be to make the `/as type` are allow blocks, which could provide more options. e.g. `load/as data [csv columns record-sep #";" field-sep #"|"]`, which means each codec has its own dialect. Not much value unless we expect to have a lot of codecs that need such flexible support. CSV is an outlier in that regard, because it's really DSV (delimiter separated), without one true spec. 

I'm fine with `load/save` being kept fairly pure, which is just a tradeoff. Either the codec system is designed to be extended, where variants of a type each have a unique name, or people will write `load*/save*` functions to parallel the standard funcs.

dockimbel:
`/flat` is very useful, don't remove it. That's the option that requires the less memory (think about CSV files with dozens/hundreds of thousands of rows) and the most flexible way to access the data the way you want (using calculated X and Y coordinates).

I would rather drop all the column-oriented and the record-oriented code, that is too specialized (it should go into a library and not in this codec). Once again, the codec needs **just** to "encode" and "decode". The natural format for decoded CSV is a block of rows (with an option for a flat block).

Codecs are not libraries, they should provided minimal encoders/decoders, because such code is costly in terms of final binary size and compilation time (this impacts every single Red script compilation, and we do dozens if not hundreds of compilation each day...). Once we get modules and/or modular compilation, you are then free to add as much features and libraries as you want.

Last, the codec model is to provide a *single* API (`load/as` and `save/as`), not one API per codec (that makes the codec model quite useless then).

GiuseppeChillemi:
Please note that lines in **CVS dataset is not constrained to having data with the same row size** on all the lines, so rows loaded could, and will, differ in length.
  Block of block [[data data data] [data data data]] is the only way to address such scenario.
  Ashley of MUNGE fame switched from flat to block of block during subsequent revisions. Also block of blocks is the standard data format. Some list-view implementations in REBOL and lot of code relies on this data structure  and it could be easly ported.
  I would suggest to adopt either /flat as option and block of block as mainstream.

9214:
Contrary to the above claim, codec can ditch ragged tables in favor of rectangular ones; in such case, missing cells are typically pre-filled with some reasonable default value (e.g. `0` or `none`); this is a standard practice in data science and numerical applications.

Issue #3759, opened 2019-01-27
`round/to .. integer!` is not always correct with `/floor` and `/ceiling` refinements


hiiamboris:
**Describe the bug**

```
>> foreach x [-2.4 -1.6 -1.4 -0.6 -0.4 -0.0 0.0 0.4 0.6 1.4 1.6 2.4] [
  print collect [keep copy [x]
    foreach s [1.0 2.0 1 2] [
      keep reduce ["^-" round/to/floor x s round/to/ceiling x s]
]]]
; x     /to 1.0         /to 2.0         /to 1   /to 2
-2.4 	-3.0 -2.0 	-4.0 -2.0 	-2 -2 	-2 -2
-1.6 	-2.0 -1.0 	-2.0 0.0 	-2 -2 	-2 -2
-1.4 	-2.0 -1.0 	-2.0 0.0 	-1 -1 	-2 0
-0.6 	-1.0 0.0 	-2.0 0.0 	-1 -1 	-2 0
-0.4 	-1.0 0.0 	-2.0 0.0 	0 0 	0 0
0.0 	0.0 0.0 	0.0 0.0 	0 0 	0 0
0.0 	0.0 0.0 	0.0 0.0 	0 0 	0 0
0.4 	0.0 1.0 	0.0 2.0 	0 0 	0 0
0.6 	0.0 1.0 	0.0 2.0 	1 1 	0 2
1.4 	1.0 2.0 	0.0 2.0 	1 1 	0 2
1.6 	1.0 2.0 	0.0 2.0 	2 2 	2 2
2.4 	2.0 3.0 	2.0 4.0 	2 2 	2 2
```

Results in the columns 2-3 are correct, but those in 4-5 aren't (they should be the same as in 2-3 but of integer type)

**Platform version (please complete the following information)**
```
Red 0.6.4 for Windows built 25-Jan-2019/6:49:24+03:00 commit #702ec38
```

greggirwin:
Good catch. R2 and R3 both appear to be correct.

PR #3988, opened 2019-08-15
FIX: issue #3759 (`round/to integer!` isn't right)

bitbegin:


greggirwin:
"isn't right" is not very descriptive. ;^)

greggirwin:
If it wasn't failing tests before, new tests should be added to check for regressions as well as correct behavior based on the problem you found.

bitbegin:
@greggirwin ok

Issue #2679, opened 2017-05-12
`prin` does not work correctly with `ask` command


Oldes:
Using code like:
```
prin "name: " ask ""
```
will put cursor on invalid position in gui-console and does not show `prin` content in the terminal console.

Oldes:
Problem is in these lines: https://github.com/red/red/blob/master/environment/console/input.red#L333-L338

Simply `ask` clears screen bellow and always resets cursor'a col position (because of its use as a main prompt in console). So now it is not possible to have input in the middle of the screen using ANSI sequences like is possible in REBOL:
```
prin "^[[2J" ;clears screen and resets cursor to 0x0
prin "^[[5;10H"  ;moves cursor position on row 5 and col 10
ask "Name: " ;do some input there
```

dockimbel:
Same issue in 0.6.4 branch.

hiiamboris:
`prin "something" input` doesn't work too
:point_up: ["help" room  June 21, 2018 10:37 AM](https://gitter.im/red/red/welcome?at=5b2b55adce3b0f268d346455)

PR #3996, opened 2019-08-19
FIX: issue #2679 (`prin` does not work correctly with `ask` command)

bitbegin:


qtxie:
This is a fix for gui-console only.

dockimbel:
@qtxie What's your opinion about this PR now? Should we close it or ask @bitbegin for extra work on it?

qtxie:
I think we have to do big changes in CLI console to fix it. If this PR fixes the issue in gui-console, then it's good to be merged. Just to be sure it won't cause any regression. It's tricky to handle `prin`, need to retest all the `prin & print` issues.

bitbegin:
this fix still not good

Issue #2573, opened 2017-04-13
unmatched multiline end character in block doesn't rise error


toomasv:
On W10 gui-console
```
>> [}]
[
```
As well
```
>> [{}}]
[
```
Also
```
>> ["}]
[
```
Should rise error.

dockimbel:
@qtxie Closing the GUI console in debug mode after inputting `[}]`, results in the Red process continuing in the DOS window, and eating up the memory.

Issue #2145, opened 2016-08-02
More brackets don't match properly


WiseGenius:
Some brackets still don't match since #2104 was fixed.

```
red>> {{}"}
{     
```

The following are all affected:

```
{{}"}
{{}a"}
{{"}"}
{{{}"}}
{"{}"}
{{"{}"}}
{"{"{}"}"}
{{{"{}"}}}
{{"{"{}"}"}}
{"{"{}"}"}
{"{""}"}
```

etc.

dockimbel:
Good catch!

Issue #2109, opened 2016-07-12
Can't paste ML string into console with semi before closing }


greggirwin:
```
red>> s: {
{    ;}
{    
```

Issue #2104, opened 2016-07-11
Some brackets don't match properly


WiseGenius:
In some situations, the console behaves as though brackets/parenthesis are still open, not letting the line execute:

Examples:
```
red>> [{}]
== [""]
red>> [{{}}]
[
```
```
red>> ({})
== ""
red>> ({{}})
(    
```
```
red>> [{[{}]}]
== ["[{}]"]
red>> [{({})}]
[    
```
```
red>> ({({})})
== "({})"
red>> ({[{}]})
(    
```

x8x:
This one too:
```
  [}
[    ]
[    }
[    
```

dockimbel:
@x8x Such issue is already logged in #1995.

PR #4004, opened 2019-08-22
FIX: issue #2573 (unmatched multiline end character in block doesn't rise error)

bitbegin:
test relative issues:
#1907 pass
#1991 pass
#1775 pass
#1995 pass
#2104 pass
#2050 pass
#2109 pass
#1147 pass
#2145 pass


dockimbel:
Great work! :+1:

PR #4018, opened 2019-08-30
merge similiar event callback

bitbegin:


bitbegin:
I tested focus event for other widget, but can't work, seems only keyboard input widget can get/set focus, and it's same with macOS (it confused me for a long time), so we will can't have the two event for other widgets

Issue #3536, opened 2018-09-18
make! hash loops when spec block contains a large number of items.


PeterWAWood:
### Expected behavior
This is from a Rebol Console session:
```text
>> blk: copy []                            
== []
>> loop 1000000 [ append blk copy "abcde" ]
== ["abcde" "abcde" "abcde" "abcde" "abcde" "abcde" "abcde" "abcde" "abcde" "abcde" "abcde" "abcde" "abcde" "abcde" "abcde" "abcde"...
>> do [
[    st: now/precise                         
[    hsh: make hash! blk                     
[    et: now/precise                         
[    difference et st                        
[    ]
== 0:00:00.060786
```

### Actual behavior
When the same expressions are evaluated in a Red Console, cpu usage rises to ~100% and remains at that level. I aborted the following red console session after it had consumed more than 19 minutes of cpu time.
```text
--== Red 0.6.3 ==-- 
Type HELP for starting information. 

>> blk: copy [] 
== []
>> loop 1000000 [ append blk copy "abcde" ]
== ["abcde" "abcde" "abcde" "abcde" "abcde" "abcde" "abcde" "abcde" "...
>> do [
[    st: now/precise
[     hsh: make hash! blk
[    et: now/precise
[    difference et st  
[    ]
^C
```
### Steps to reproduce the problem
See above.
### Red and platform version
commit 6115b00130e64436f0127e442a4fdd8316182371 (HEAD -> master, origin/master, origin/HEAD)
macOS 10.12.6

qtxie:
We need to optimize the hashtable for this case.

PR #4026, opened 2019-09-07
FIX: issue #3536 (make! hash loops when spec block contains a large n…

qtxie:
…umber of items).

FEAT: adds some functions to manipulate integer and pointer array (needed in many places)

### DIFF
--- a/runtime/hashtable.reds
+++ b/runtime/hashtable.reds
@@ -13,6 +13,141 @@ Red/System [
 	}
 ]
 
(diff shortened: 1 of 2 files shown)


qtxie:
I'm not satisfied with the implementation. It makes the hashtable code even more complicated just for fixing a edge case. I'm not sure in real use case we'll have this issue. So I prefer to wait a real use case to merge it.

PR #4027, opened 2019-09-09
fix some small issues

bitbegin:


dockimbel:
@bitbegin Please insert a [GTK] tag after FEAT/FIX in commit messages for commits that are only for GTK backend.

dockimbel:
@bitbegin Maybe no need to insert the [GTK] header for now, until the GTK branch will be merged in master.

PR #4028, opened 2019-09-10
FIX: remove not needed 'either' branching

avitkauskas:
While studying the Red code, I found this place where both `either` branches produce absolutely the same result so this branching is not needed and removal of it simplifies the code.


dockimbel:
Looks like a nice catch!

avitkauskas:
Yes, this duplication could be with the idea of speed optimization, but with simplification we are also removing the `unit = unit2` test in the loop, so the loss is less that one `switch` check ;)
I am glad you liked and accepted the change!

Issue #4033, opened 2019-09-15
RECYCLE does not have a docstring


meijeru:
It is the only native! function with this omission.

meijeru:
I am pleased with the quick reaction and will close this now. The more issues closed, the better!

PR #4034, opened 2019-09-15
FIX: RECYCLE does not have a docstring #4033

endo64:
Added docstring to `recycle`. Also fixed a very small typo in `translit` function.


greggirwin:
Thanks Endo!

Issue #3993, opened 2019-08-16
Float numbers loading is not identical across platforms


dockimbel:
On Windows:
```
>> to-binary 2.022794416824729e-308
== #{000E8BA2E8BA2E8C}
>> to-binary 2.022794416824728e-308
== #{000E8BA2E8BA2E8A}
>> 
```
On non-Windows:
```
>> to-binary 2.022794416824729e-308
== #{000E8BA2E8BA2E8D}
>> to-binary 2.022794416824728e-308
== #{000E8BA2E8BA2E8B}
```
This results in accuracy issues leading to failing unit tests like in #2184.

Issue #3243, opened 2018-03-02
Loss of precision when serialising and deserialising a float!


PeterWAWood:
### Expected behavior
A float! when molded and loaded should not lose precision i.e.
    float-number == load mold float-number

### Actual behavior
Precision is lost when molding and loading a float! value:
```text
>> 
>> f: 11.651178950846456
== 11.65117895084646
>> f == load mold f
== false
>> f - load mold f
== -3.5527136788005e-15
>> 
```

### Steps to reproduce the problem
See above

### Red version and build date, operating system with version.
>> system/build
== make object! [
    date: 5-Feb-2018/7:28:22+08:00
    config: make object! 
>> 

macOS 10.12.6

qtxie:
Using [dtoa](www.netlib.org/fp/dtoa.c) will solve it.

dockimbel:
This is a known limitation of the current string-to-float parser. It's a notoriously complex task to handle, and the `dtoa` lib is the best option, though, it's a huge and complex dependency that will add ~40KB to the runtime payload. We plan to integrate it at some point, though, it requires extra work and lacks unit testing.

hiiamboris:
Let's reopen! :)
Same issue with percents exists:
```
>> f: 11.651178950846456%
== 11.65117895085%           ;) cut down in the output
>> to float! f
== 0.1165117895084646
>> f = load mold/all f
== false
>> f = load mold/all to float! f
== true
```

qtxie:
It's a percent! issue. I prefer to open a new issue for it.

PR #4036, opened 2019-09-17
FEAT: preliminary use dtoa library to load and form float.

qtxie:


qtxie:
This PR fixes issue #3993 and #3243.

PR #4049, opened 2019-09-24
FEAT: switch pair! from integer! to float32!

bitbegin:


qtxie:
We still need to support float pair! in the compiler lexer.

hiiamboris:
Just curious here :) 
Have you considered using a pair of 32-bit integers instead of float-32?
IMO, range from -2147483.648 to 2147483.647 should cover every possible graphics need, even with all the intermediary computation results.
It's just that using floats you open a pandora's box of NaNs, INFs, rounding, and inexact arithmetic.
Is there a reason why float should be preferred to integer in pair's case?

qtxie:
> Have you considered using a pair of 32-bit integers instead of float-32?

It is 32-bit integers in current implementation.

> It's just that using floats you open a pandora's box of NaNs, INFs, rounding, and inexact arithmetic.

You have those problems too using integers. Even worse, using integers you lost precision in many places. e.g DRAW dialect, font size.

hiiamboris:
@qtxie in the above I'm implying that the last 3 digits of the integer can be used as a fractional part. To put it another way, 1/1000 of a pixel minimum increment size. The same tactic that's used for the money datatype, really.

qtxie:
@hiiamboris Ah, Ok. You mean a custom decimal format. That needs more work. Also most of the modern OSes use float32! in their Graphic APIs. Using float32! make it easier to integrated with the OS APIs.

hiiamboris:
I see. Thanks for the info :)

qtxie:
Some notes about this change:
Pros:
1. More accurate drawing in DRAW.
2. More accurate offset and size. E.g. face/offset, font/size.
Cons:
1. Integer pair works better for images.
2. No bitwise operations. (and, or, xor)

Using float32 pair we can work around those Cons by converting `pair/x` and `pair/y` to integer first.
But we cannot have the Pros using integer pair.

dockimbel:
Closing this PR as the feature has been implemented as a different `point2D!` datatype.

Issue #3778, opened 2019-02-09
Clipboard doesn't work consistently under Windows when compiled


Mennohexo:
write-clipboard "ABC" only works in interpreted mode and not compiled .exe Windows
```write-clipboard "ABC"```

**Comment by hiiamboris** [comment]
works fine for me compiled on W7

**Comment by 9214** [comment]
Any time you submit an issue, please make an effort to fill a more detailed bug report and to use the [standard issue template](https://github.com/red/red/blob/master/.github/ISSUE_TEMPLATE/bug_report.md) (which was automatically suggested to you on submission).

Moreso, I cannot reproduce your problem neither with MSDOS nor with Windows (and not even with WindowsXP) as compilation targets, neither in release nor in development mode.

In light of the above (and taking @hiiamboris' message into account), the issue can be dismissed, and OP is recommended to confirm his findings in [/bugs](https://gitter.im/red/bugs) room (preferrably by multiple members and on multiple platforms) before opening tickets - that would save everyone's time and effort.

**Comment by Mennohexo** [comment]
again.
write-clipboard "ABC" is not working in compied .exe Windows.
Description 
Ingredients - Windows XP SP3 - red-064.exe
If you have a red programming script with the command ```write-clipboard "ABC"``` that is
activated in your script with a button then this is already done in INTERPRETED mode.
When you compile the program with  - c flag or with -r -t Windows then the windows .exe is ready compiled and working correct.
Only the clipboard command is NOT working.
You can check it out with the windows text editor.
Type some text and COPY it with the mouse context menue. Then this text is alredy in clipboard
everyone knows. Pasting it in editor more times.
Now , if you press the button of your red program that intention is to OVERWRITE the clipboard with the letters ABC , it is NOT working.
I have checked this out on two different computers.

**Comment by endo64** [comment]
Worked for me on Win10:

My `test.red` script:
```
Red [Needs: View]
write-clipboard "ABC"
```

Compiled and tested:
```
red -t msdos -c test.red
test.exe
```

`ABC` was set into clipboard.

**Comment by greggirwin** [comment]
I confirmed the bug in chat and asked @Mennohexo to file a ticket. I agree that following the submission format and providing more info would help.

This doesn't show the problem:
```
write-clipboard "ABC"
print read-clipboard
```

But this does:
```
write-clipboard ""
print read-clipboard
write-clipboard "ABC"
print read-clipboard
```
On Win10 here. Similarly, under View you can show the problem:

No problem:
```
Red [needs: view]
[...]

PR #4050, opened 2019-09-25
FIX: more robust fix for issue #3778

hiiamboris:
It is very annoying when I press `Ctrl+C` in GUI console, then `Ctrl+V` to paste the snippet into a different place - only to see that it pasted something else entirely and hopelessly messed the code...

**Summary**: This fix reduces clipboard failure rate from 10-15% to around 0.0001% (which I find acceptable) by trying OpenClipboard up to three times in a row, yielding the thread's time slice between attempts.

**More info** about #3778 problem:

The MS docs say in multiple places: 
> Before calling EmptyClipboard, an application must open the clipboard by using the OpenClipboard function. If the application specifies a NULL window handle when opening the clipboard, EmptyClipboard succeeds but sets the clipboard owner to NULL. Note that this causes SetClipboardData to fail.

I do not observe this. Although we're calling OpenClipboard with a NULL handle, SetClipboard never failed in all my tests. Instead, OpenClipboard *sometimes* fails after a *recent* EmptyClipboard operation (with an error code 5 = "Access Denied"). No call to EmptyClipboard => OpenClipboard never fails. Providing a non-NULL hwnd does not fix the issue.

So, it looks like an MS bug. The best solution I could come up with is make 3 attempts.

Here are my stats before the fix:
```
>> n: 0 loop 100000 [write-clipboard "A" if false = read-clipboard [n: n + 1]] 100% * n / 100'000
== 14.243%
>> n: 0 loop 100000 [write-clipboard "A" if false = read-clipboard [n: n + 1]] 100% * n / 100'000
== 10.072%
>> n: 0 loop 100000 [write-clipboard "A" if false = read-clipboard [n: n + 1]] 100% * n / 100'000
== 9.626%

>> n: 0 loop 100000 [unless write-clipboard "A" [n: n + 1]] 100% * n / 100'000
== 2.177%
>> n: 0 loop 100000 [unless write-clipboard "A" [n: n + 1]] 100% * n / 100'000
== 2.685%
>> n: 0 loop 100000 [unless write-clipboard "A" [n: n + 1]] 100% * n / 100'000
== 3.596%
```

And this is after the fix:
```
>> n: 0 loop 1000000 [write-clipboard "A" if false = read-clipboard [n: n + 1]] 100% * n / 1'000'000
== 0%
>> n: 0 loop 1000000 [write-clipboard "A" if false = read-clipboard [n: n + 1]] 100% * n / 1'000'000
== 0.0001%
>> n: 0 loop 1000000 [write-clipboard "A" if false = read-clipboard [n: n + 1]] 100% * n / 1'000'000
== 0%

>> n: 0 loop 1000000 [unless write-clipboard "A" [n: n + 1]] 100% * n / 1'000'000
== 0.0001%
>> n: 0 loop 1000000 [unless write-clipboard "A" [n: n + 1]] 100% * n / 1'000'000
== 0.0001%
>> n: 0 loop 1000000 [unless write-clipboard "A" [n: n + 1]] 100% * n / 1'000'000
== 0.0001%
```

### DIFF
--- a/runtime/clipboard.reds
+++ b/runtime/clipboard.reds
@@ -64,6 +64,9 @@ clipboard: context [
 					str			[byte-ptr!]
 					return:		[integer!]
 				]
+				Sleep: "Sleep" [
+					dwMilliseconds	[integer!]
+				]
 			]
 		]
 
@@ -81,13 +84,20 @@ clipboard: context [
 			/local
 				hMem	[integer!]
 				p		[byte-ptr!]
-				cp		[integer!]
+				ok		[logic!]
 				val		[red-value!]
 		][
 			p: null
 			val: as red-value! none-value
 
-			unless OpenClipboard main-hWnd [return as red-value! false-value]
+			;-- OpenClipboard often fails on W7 after an EmptyClipboard call, so be persistent
+			ok: yes
+			loop 3 [
+				unless ok [Sleep 1]
+				ok: OpenClipboard main-hWnd
+				if ok [break]
+			]
+			unless ok [return as red-value! false-value]
 
 			hMem: GetClipboardData CF_UNICODETEXT
 			if hMem <> 0 [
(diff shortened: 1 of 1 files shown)


qtxie:
Thanks a lot for fixing this nasty issue. :+1:

greggirwin:
@hiiamboris I always appreciate your attention to detail. I remember having clipboard issues (outside of Red) in the past, and this is a nice workaround.

PR #4057, opened 2019-09-30
FEAT: clipboard extension to support images and file lists, Windows only

### LINKED ISSUE #4055: Read-clipboard doubles newline chars
I have already fixed this issue. Just reporting it for historical reference.

**Describe the bug**

Text in the clipboard:
```
1
2
3
```
Let's try to read it:
```
>> print probe read-clipboard
{1^M^/2^M^/3^M^/}
1

2

3

```
As you can see every new-line appears twice.

This also has a very harmful effect:
```
>> length? read-clipboard
== 7
>> write-clipboard read-clipboard
== true
>> length? read-clipboard
== 9
```
It grows! ☻ 

Then `loop 10000 [write-clipboard read-clipboard]` will just crash it since memory footprint grows exponentially.

**Expected behavior**
```
>> print probe read-clipboard
{1^/2^/3^/}
1
2
3
```

```
Red 0.6.4 for Windows built 25-Sep-2019/22:36:30+03:00 commit #f753e25
```

hiiamboris:
Also: Fixes #4055 

Allows to read/write from/to the clipboard:

**1. Images**

This is a bloody mess. See https://stackoverflow.com/a/6565158 , https://stackoverflow.com/a/46400011 , https://stackoverflow.com/a/15691001 for more info.

**TL;DR** Windows' clipboard only supports opaque images reliably. We can read/write alpha-enabled images in Red, but most programs will ignore the alpha. To avoid the troubles of GDI/GDI+/DIB/clipboard interaction, some programs also copy/paste raw image data into the clipboard (as PNG/JPEG/etc), so I made it to support this too: it writes the images in both PNG and DIB format, reads in any of PNG/JPG/GIF/DIB (in the order of preference).

`read-clipboard` in this case returns an `image!`
`write-clipboard` accepts an `image!`
`read-clipboard write-clipboard <image>` preserves the alpha channel data

Ctrl+V in the GUI console does not paste the images (as those can be quite huge)

**2. File lists**

This is the format that Explorer uses when one selects and copies some files to the clipboard.
This is also how Firefox (and some other programs I've seen) copies an image: as a block with a single file path.

`read-clipboard` in this case returns a `block!` of `file!`s converted with `to-red-file`
`write-clipboard` accepts a `block!` with 0+ `string!`s or `file!`s

Ctrl+V in the GUI console inserts a molded block.

**3. Allows** one to **empty the clipboard** with `write-clipboard none` command

----

~~I have the `read-clipboard` and `write-clipboard` moved from `routines.reds` into `functions.reds` as I had to convert these into functions in order to use `to-red-file` and `to-local-file`. Calling these from R/S looked troublesome to me. Is there an easy way I'm missing?~~

Also introduced a set of tests for the clipboard.

### DIFF
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -1069,47 +1069,6 @@ last?: func [
 	1 = length? series
 ]
 
-clipboard-internal: context [
-	read-clipboard*: routine ["Return the contents of the system clipboard"][
-		stack/set-last clipboard/read
-	]
-
-	write-clipboard*: routine ["Write content to the system clipboard" data [any-type!]][
-		logic/box clipboard/write as red-value! data
-	]
-]
-
-#either config/OS = 'Windows [
-	read-clipboard: func ["Return the contents of the system clipboard" return: [string! block! image! none! logic!] /local r] [
-		also r: clipboard-internal/read-clipboard*
-		if block? r [forall r [change r to-red-file r/1]]
-	]
-
-	write-clipboard: func ["Write content to the system clipboard" data [string! block! image! none!] return: [none! logic!]] [
-		either block? data [
-			data: copy data
-			forall data [change data to-local-file :data/1]
-		][
-			all [
-				image? data
-				empty? data
-				data: none
-			]
-		]
-		clipboard-internal/write-clipboard* data
-	]
-][
-	read-clipboard: func ["Return the contents of the system clipboard" return: [string! none! logic!] /local r] [
-		clipboard-internal/read-clipboard*
-	]
-
-	write-clipboard: func ["Write content to the system clipboard" data [string!] return: [none! logic!]] [
-		clipboard-internal/write-clipboard* data
-	]
-]
-
-unset 'clipboard-internal
-
 ;------------------------------------------
 ;-				Aliases					  -
 ;------------------------------------------
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -1069,6 +1069,37 @@ last?: func [
 	1 = length? series
 ]
 
+clipboard-internal: context [
+	read-clipboard*: routine ["Return the contents of the system clipboard"][
+		stack/set-last clipboard/read
+	]
+
+	write-clipboard*: routine ["Write content to the system clipboard" data [any-type!]][
+		logic/box clipboard/write as red-value! data
+	]
+]
+
+read-clipboard: func ["Return the contents of the system clipboard" return: [string! block! image!] /local r] [
+	also r: clipboard-internal/read-clipboard*
+	if block? r [forall r [change r to-red-file r/1]]
+]
+
+write-clipboard: func ["Write content to the system clipboard" data [string! block! image! none!]] [
+	either block? data [
+		data: copy data
+		forall data [change data to-local-file :data/1]
+	][
+		all [
+			image? data
+			empty? data
+			data: none
+		]
+	]
+	clipboard-internal/write-clipboard* data
+]
+
+unset 'clipboard-internal
+
 ;------------------------------------------
 ;-				Aliases					  -
 ;------------------------------------------
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -1079,23 +1079,33 @@ clipboard-internal: context [
 	]
 ]
 
-read-clipboard: func ["Return the contents of the system clipboard" return: [string! block! image!] /local r] [
-	also r: clipboard-internal/read-clipboard*
-	if block? r [forall r [change r to-red-file r/1]]
-]
-
-write-clipboard: func ["Write content to the system clipboard" data [string! block! image! none!]] [
-	either block? data [
-		data: copy data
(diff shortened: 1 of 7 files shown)


environment/functions.red
```diff
@@ -1069,6 +1069,47 @@ last?: func [
...
+	]
+
+	write-clipboard*: routine ["Write content to the system clipboard" data [any-type!]][
+		logic/box clipboard/write as red-value! data
+	]
+]
+
+#either config/OS = 'Windows [
+	read-clipboard: func ["Return the contents of the system clipboard" return: [string! block! image! none! logic!] /local r] [
+		also r: clipboard-internal/read-clipboard*
```
greggirwin: Splitting the `also` args over lines here makes things less clear IMO. A third line of just `r`, for the return, seems better to me. Not critical, and I see what you're doing in how you use it.

tests/source/units/clipboard-test.red
```diff
@@ -0,0 +1,95 @@
...
+#include  %../../../quick-test/quick-test.red
+
+~~~start-file~~~ "clipboard"
+
+===start-group=== "text IO"
+
+	do [if system/platform <> 'Linux [
+
+		--test-- "text-io-1"
+			--assert false <> write-clipboard ""
```
greggirwin: Why test in the negative with `false <> ...`?

environment/functions.red
```diff
@@ -1069,6 +1069,47 @@ last?: func [
...
+		if block? r [forall r [change r to-red-file r/1]]
+	]
+
+	write-clipboard: func ["Write content to the system clipboard" data [string! block! image! none!] return: [none! logic!]] [
```
greggirwin: TBD, add a doc string for `data` to note the required block criteria (i.e. files only).

environment/functions.red
```diff
@@ -1069,6 +1069,47 @@ last?: func [
...
+		either block? data [
+			data: copy data
+			forall data [change data to-local-file :data/1]
+		][
+			all [
+				image? data
```
greggirwin: Maybe add a comment that images pass thru, and empty strings are converted to none to clear the clipboard. Took me a minute to see that.

greggirwin:
This is a huge amount of work, and really exciting @hiiamboris. Thanks! I'll give @dockimbel and @qtxie a chance to review, but will merge shortly if they don't have time. Thanks, too, for including tests.

tests/source/units/clipboard-test.red
hiiamboris: It returns `false` when some WinAPI calls fail. Something we don't want to happen.

hiiamboris:
☺ You're welcome.
Let's wait for their reviews, and when you're ready to merge let me know, I'll apply all the fixes.

qtxie:
Thanks for the clipboard extension. 👍 
@hiiamboris You can use `#call` in R/S to call the Red functions. e.g.
https://github.com/red/red/blob/master/modules/view/backends/windows/comdlgs.reds#L134

Also do you have any idea about whether those features possible on macOS and Linux?

hiiamboris:
@qtxie Thanks, `#call` looks promising :) I'll try to turn these funcs into routines again.

MacOS seems promising: https://developer.apple.com/documentation/appkit/nspasteboard/pasteboardtype
I can see all the formats supported by Windows (as well as audio data which we might implement once Red has a type to hold it), plus a lot more.
[A pioneer says that it's the easiest on MacOS](https://jtanx.github.io/2016/08/19/a-cross-platform-clipboard-library/#os-x).

On Linux the situation with clipboard was always grim. One should arm oneself with [this heavy tome](https://tronche.com/gui/x/icccm/) to banish all possible kinds of evil into the clipboard. On a bright side, I can see text and images [in this list](https://tronche.com/gui/x/icccm/sec-2.html#s-2.6.2). At a glance it looks like it may be possible to transfer lists of files as well. Unsure about audio data.

environment/functions.red
hiiamboris: I'm going to make these routines again, not enough space to put any comments in there.

Technically `write-clipboard ""` and `write-clipboard none` are different. The former does actually put a string there, albeit an empty one. While the latter does not put anything into it (likely the state you're in after the OS just booted up). It's not distinguishable by means of Ctrl+V, but only programmatically ☺

An empty image (0xH or Wx0) on the other hand, works the same as `write-clipboard none`. Too many points of failure appear otherwise, so I decided to take the easy way.

environment/functions.red
hiiamboris: Okay.

environment/functions.red
hiiamboris: Removed this.

hiiamboris:
I've had some adventures with the `#call` thing. It destroys the stack contents, so I've wrapped it into a `mark-native`+`unwind` pair.

qtxie:
I'll merge it. We'll move the code the `clipboard://` port later.

PR #4059, opened 2019-10-03
FEAT: JSON codec is now encapped and merged into one file

rebolek:


greggirwin:
The reason for this change is that encapping saves space in the Red binary, and can also save small amounts of build time.

PR #4065, opened 2019-10-05
some code improvements

bitbegin:


bitbegin:
@rcqls ok, i will try to fix it

Issue #3270, opened 2018-03-22
Panel face with set text facet fills with black


DVL333:
### Expected behavior
Panel face with set text facet is transparent.
### Actual behavior
Panel face with set text facet fills with black.
### Steps to reproduce the problem
```
pnl: make face! [type: 'panel size: 80x80 offset: 10x10 text: "Panel"]
repend p: [] [pnl]
view make face! [type: 'window size: 100x100 pane: p]
```
### Red version and build date, operating system with version.
Windows7 Ultimate 64bit, Red for Windows version 0.6.3 built 19-Mar-2018/3:02:17+03:00

9214:
Can't reproduce on W10.

DVL333:
2 9214
Yes, on W10 it works well. Problem is on W7 - background of the panel becomes black regardless of its font color.

hiiamboris:
On my W7 it's not black: it just displays what was underneath the window at the time of it's creation.
Interestingly, this works well: `g: view/no-wait [at 10x10 panel 80x80 "Panel"]`
But this does not: `view/no-wait f: make face! [type: 'window size: 100x100 pane: p parent: system/view/screens/1]`
Although I can't find any difference between `? f` and `? g` outputs. So it must be related to how View constructs the window face in the first place.

hiiamboris:
FYI. The cause of the issue is that [`layout` func at L721](https://github.com/red/red/blob/1d4f76601ae1ebfc459e79292820088205008c6f/modules/view/VID.red#L721) assigns a default color to panel when none is given. 
`make-face` and `make face!` are more low-level and do not do that, so the panel color remains `none` and it's background is not drawn.

hiiamboris:
We have removed "text" support from panel in one of the commits (panel is not meant for text display). But I'm not sure if current behavior is the desired one.

Expectation `Panel face with set text facet is transparent` will unlikely be fulfilled, due to GDI limitations. But should we make panels of OS default color, or derive that color it's parent face (other panel or window)?

Currently it behaves like this, at least on W7:
`view [backdrop white size 100x100 panel 80x80]`
![](https://i.gyazo.com/f7cdcfbcf27707048aca6e29f65bbd4c.png)

If any transparency is present in the panel's color, it does not draw itself over the window's background, but rather over the underlying window:
`view [backdrop white size 100x100 panel #FF00FFE0 80x80]`
![](https://i.gyazo.com/f8949a3e5dfa336e454d40afa35a0cf3.png)
[...]

PR #4073, opened 2019-10-11
FIX: issue #3270 (Panel face with set text facet fills with black)

hiiamboris:
I believe `panel` face was not designed to display any "text" on it directly. It kinda worked using same code that does so for a `base`, but `base` with `color: none` does not paint it's background so it never did for the panel. Thus I disabled it. As a consequence the background is now painted by DefWindowProc even when it has a "text" facet assigned.

An alternative would be to redraw the panel manually using `base` code, but specifying a default window color, and *then* draw a text on it.

Let me know if I judged correctly.

### DIFF
--- a/modules/view/backends/windows/base.reds
+++ b/modules/view/backends/windows/base.reds
@@ -132,6 +132,7 @@ render-base: func [
 	if all [
 		group-box <> type
 		window <> type
+		panel <> type
 		render-text values hWnd hDC :rc null null
 	][
 		res: true


greggirwin:
Thanks @hiiamboris, for the analysis and fix. Now we have a design question to answer. _Should_ `panel`s support a `text` facet. If so, we need to go with your alternative, correct? If not, we need to doc that and decide if failing silently when someone tries to use `text` is OK.

PR #4076, opened 2019-10-13
Issue 3362 fix

### LINKED ISSUE #3362: Strange behaviour of CONTEXT and OBJECT with exit in spec-block
### Expected behavior

`make object! [x: 1 exit]` when used outside a function gives an error saying `return or exit not in function`
`context` and `object` are (equivalent) functions that contain a call of `make object!`; therefore a `return` or `exit`  in the spec block is not caught this way; in fact `return` is processed "normally" yielding the value following it as return value of the function `context` resp. `object`; however, `exit` in the spec block should yield the `unset!` value as a result of the `context` or `object` function, and it does not seem to do so;
Compare:
```
>> f: does [exit]
== func [][exit]
>> type? f
== unset!
>> 
```

### Actual behavior

```
>> o: context [exit]
== [exit]
>> type? o
== console
```

Execute the above code
### Red and platform version
```
-----------RED & PLATFORM VERSION----------- 
RED: [ branch: "master" tag: #v0.6.3 ahead: 669 date: 27-Apr-2018/10:07:23 commit: #4c36108655f3410124f8539543cbd4bc4525fed1 ]
PLATFORM: [ name: "Windows 10" OS: 'Windows arch: 'x86-64 version: 10.0.0 build: 16299 ]
--------------------------------------------

```

hiiamboris:
Fixes #3362 (interpreter only, compiler doesn't work with that pattern at all anyway)

### DIFF
--- a/runtime/datatypes/function.reds
+++ b/runtime/datatypes/function.reds
@@ -249,6 +249,7 @@ _function: context [
 				RED_THROWN_BREAK
 				RED_THROWN_CONTINUE
 				RED_THROWN_THROW	[re-throw]			;-- let exception pass through
+				RED_THROWN_EXIT
 				RED_THROWN_RETURN	[stack/unwind-last]
 				default [0]								;-- else, do nothing
 			]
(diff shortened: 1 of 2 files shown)


greggirwin:
Thanks!

dockimbel:
@greggirwin Just as a reminder: all PR's to Red's core need to be validated by me or at least Qingtian.

PR #4090, opened 2019-10-17
FIX: libRedRT export string/load-at

loziniak:
with some cleanup

### DIFF
--- a/system/utils/libRedRT-exports.r
+++ b/system/utils/libRedRT-exports.r
@@ -376,7 +376,45 @@
 	red/natives/recycle*
 
 	;-- for view backend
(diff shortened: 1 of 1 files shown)


bitbegin:
great!

dockimbel:
Thanks for the cleanup. ;-)

qtxie:
Just as a reminder, those export functions are only for accelerating the compilation of the GTK view module. End user should not need them. (use `-u` should be the right way for end user). 
So export functions are Linux backend specific. We need to remove those functions before merge GTK branch into master. e.g
```
red/OS-image/to-pixbuf
red/OS-image/load-pixbuf
red/OS-image/buffer-argb-to-abgr
```

PR #4093, opened 2019-10-19
Gc bug

qtxie:
The tail of a series is the same as the beginning of the next series it adjacent to. If a direct pointer point to the tail of a series, the GC may think it's the beginning of a series and updated it to wrong value.

This fix add 4 bytes gap between two adjacent series!s. It fix the bug occurs in `run-all-interp.red`.


greggirwin:
Nice find!

PR #4098, opened 2019-10-21
remove object! from find's docstring; see commit 7b1eeb76

loziniak:
as discussed at gitter: https://gitter.im/red/docs?at=5dadc2739825bd6baca0d166


9214:
Note that `any-object!` also includes `error!` and `port!`, and the latter supports `find` action.

loziniak:
Great, thanks for pointing that out.

PR #4120, opened 2019-11-04
fix some issues

bitbegin:


greggirwin:
Could the subject for this PR be more specific? :^)

PR #4122, opened 2019-11-05
support all keys

bitbegin:


bitbegin:
still need some improve, just wait before merge, thanks!

PR #4143, opened 2019-11-21
time formatting tests

hiiamboris:
what was fixed so far by @qtxie


hiiamboris:
Failure on MacOS due to:
`curl: (7) Failed to connect to static.red-lang.org port 443: Operation timed out`

PR #4144, opened 2019-11-22
FIX: issues 497, 3561

### LINKED ISSUE #497: DO doesn't evaluate path!
red>> b: [1]
== [1]
red>> p: 'b/1
== b/1
red>> do p
== b/1

### LINKED ISSUE #3561: DO ANY-PATH! misbehaves
### Description
`do` applied to some of `any-path!` values yields unexpected results.

* Here I expected `c`:
```red
>> a: [b c]
== [b c]
>> a/b
== c
>> do 'a/b
== false
```
* And here too:
```red
>> do quote a/b
== a/b
```
* Here `lit-path!` should have decayed to `path!`:
```red
>> do quote 'a/b
== 'a/b
```
* And this one seems to be correct:
```red
>> do quote :a/b
== c
```
### Red version and OS
```
Red 0.6.3 for Windows built 6-Oct-2018/7:28:28+05:00 commit #dafc828
```

hiiamboris:
Fixes #3561 , fixes #497 

As usual, stack trickery is the culprit ;)
Nothing a few magick spells can't fix.

`stack/set-last arg + 1` line was a remnant of [2013 version of Red](https://github.com/red/red/blob/778dfb31a57810a60eabec171d110b47f607ed18/red/runtime/interpreter.reds#L489)


dockimbel:
This PR breaks regression test for #3385.

hiiamboris:
The test was valid so it's a regression. I'll look into it.

hiiamboris:
@dockimbel https://github.com/red/red/pull/4663 should fix this PR

Issue #2431, opened 2017-02-08
DO and LOAD do not skip the Red header


meijeru:
In Rebol a well formed Rebol header (`Rebol [...]`) is not evaluated to obtain the result of DO or LOAD. I am not sure what the policy for Red is going to be, but so far it seems the Red header is not skipped.

geekyi:
I don't understand by `do` skipping the header, but `load` definitely behaves differently:
Rebol2:
```rebol
>> load {Rebol}
== Rebol
>> load {Rebol []}
== [
]
>> load {Rebol [green] blue}
== blue
```
In red:
```red
red>> load {Red [green] blue}
== [Red [green] blue]
```
All my tests with `do` so far, indicates header is `load`ed but not evaluated (similar to a `block!`)
Any examples?

dockimbel:
LOAD does load the Red header, but DO should not evaluate it (as long as the argument is a `file!`). 

When testing, it seems DO does not behave as intended:
```
red>> load %test.red
== [Red [] 
    print "ok" 
    halt
]
red>> red: does [print "header"]
== func [][print "header"]
red>> do %test.red
header
ok
(halted)
```

meijeru:
See also new issue #2439

x8x:
It is my understanding that a file loaded by `load` doesn't need a Red header, but doing a file with `do` does.
Actually doing a file doesn't skip the Red header, it works as a side effect of the `red` color being defined in `%red/environment/colors.red` here is the proof:
```
unset 'red
  do %test.red
*** Script Error: Red has no value
*** Where: do
  do load %test.red
*** Script Error: Red has no value
*** Where: do
```
More a bug then a wish imho.

geekyi:
@x8x I may be wrong, but the side effect may have been intentional, as it allows the following convention:
```rebol
Rebol [
  file: %r2red.r
  ..blah..blah..
]
Red[]
```
For creating Red and Rebol inter-operable files. Still speculation, but I feel doc had chosen a name that already exists as a Rebol `word!` for this. (and Red having "Re.." in it)

All just what it looks to me on the outside

That said, do you still think it would cause problems? (Rebol could also have been unset, also, it could contain malicious code, etc. But I don't immediately imagine that would be out of control of the developer..)

PR #4147, opened 2019-11-23
FIX: Red [] headers consistency and some

### LINKED ISSUE #3588: Red program assumed too eagerly
**Describe the bug**

When running a file (in GUI console) containing text before Red program header, interpreter is started at the first occurrence of the word "Red" in text before header, not at the proper header.

**To Reproduce**

1. Create file `try.txt`
```
Hello Red probe "try"
Red [] probe "Hoi!"
```

2. `do %try.txt`
Result
```
"try"
"Hoi!"
```

3. Change "Red" in file to e.g. "Reb":
```
Hello Reb probe "try"
Red [] probe "Hoi!"
```

4. Result of `do %try.txt`:
```
"Hoi!"
```

**Expected behavior**

Interpretation of Red program should start on proper Red header.

```
Red 0.6.3 for Windows built 8-Nov-2018/9:26:02+02:00 commit #06ddce9
```

### LINKED ISSUE #3670: [Compiler] Does not detect the script header properly
**Describe the bug**
A headerless script should trigger a compilation error.
What really happens:
```
*** Red Compiler Internal Error: Script Error : copy expected value argument of type: series port bitset
*** Where: rejoin
*** Near:  [mold copy/part pos 40]
```

**To reproduce**
Try to compile any script without `Red []` or `Red/System []` prefix

```
Red 0.6.4 for Windows built 7-Dec-2018/8:41:53+03:00 commit #1bdbdc2
```

hiiamboris:
Fixes #3670, fixes #3588 
fixes `do` part of #2431 (`load` part still needs `/header` implemented or something)

Summary:

**header is always searched for using first match of the rule `["Red" opt "/System" any [" " | "^-" | "^/" | "^M"] "["]`.**
I'm not sure if we want to accept "Red" and "[]" on different lines but perhaps this can be useful for auto-generated scripts. That was the prevailing behavior anyway.

So, "Red [] Red/System []" is a Red file, "Red/System [] Red []" is an R/S file. We can't have both in one script as we do with "REBOL[] Red []", because we rely on the header to determine the dialect used (Red or R/S), rather than file extension.

**headerless files are properly reported by the compiler**

**`do %file` skips both `Red` and header**, so:
- you can `unset 'red` at will and still be able to `do %files`
- using `#macro ['Red] [..]` or pre-load to do esoteric stuff like inserting things between "Red" and "[..]" after the header was found, won't work anymore

***qt-tmp-file* (file! value) can be used in tests** to write and read temporary junk. Uncommented tests that are using it.


red.r
```diff
@@ -249,14 +249,13 @@ redc: context [
 	]
 
 	red-system?: func [file [file!] /local ws rs?][
-		ws: charset " ^-^/^M"
+		ws: charset " ^-^M^/"
```
dockimbel, line 252: :-D

environment/console/engine.red
```diff
@@ -283,7 +283,7 @@ system/console: context [
 
 	launch: function [/local result][
 		either script: src: read-argument [
-			parse script [some [[to "Red" pos: 3 skip any ws #"[" to end] | skip]]
+			parse/case script [to [pos: "Red" opt "/System" any ws #"[" to end]]
```
dockimbel, line 286: This part is wrong. The purpose of `some` is to skip over "Red" sequences not followed by a block (for example a Red script embedded in an HTML page). Please bring `some` back.

environment/functions.red
```diff
@@ -855,13 +855,15 @@ split-path: func [
 	reduce [dir pos]
 ]
 
-do-file: func ["Internal Use Only" file [file! url!] /local saved code new-path src][
+do-file: func ["Internal Use Only" file [file! url!] /local saved code new-path src ws][
+	ws: charset " ^-^M^/"
 	saved: system/options/path
-	unless src: find/case read file "Red" [
+	unless parse/case read file [to [src: "Red" opt "/System" any ws #"["] to end] [
```
dockimbel, line 861: You can't `do` a Red/System script, so what's the point of `opt "/System` there?

environment/console/engine.red
dockimbel: I've added it back while fixing the conflicts.

hiiamboris: @dockimbel No, it's correct. Look closer: the rule succeeds only when it finds a block `#"["`, because I externalize `to`. Rule as you formulated it `[some [[src: "Red" opt "/System" any ws #"["] to end] | skip]` in fact has no effect other than slowing down error detection:
- when `[src: "Red" opt "/System" any ws #"["]` succeeds, it's no different than `to` version: `to end` will guarantee that it never evaluates twice
- when this subrule fails once, it never succeeds, but tries to match `to` (very heavy rule) for every character in the file, resulting in O(n^2) load time ... see https://github.com/red/red/issues/4410#issuecomment-620756693 that you fixed yourself previously :D (I think you've forgotten to add regression tests though)

Issue #3563, opened 2018-10-12
Line-breaks in area on Windows add to calculated length 1 for each line-break


toomasv:
**Describe the bug**
When all of multi-line text is selected in `area`, `selected` property of the face gives different result than `length?`. This distorts calculations of cursor positions in area.

**To Reproduce**
1. In console enter: `view [a: area "Some^/line^/here" button "Show" [print [length? a/text a/selected]]]`
2. Select all text in area and press 'Show'
3. See difference in console

**Expected behavior**
`selected` should return same position as is returned by `length?`.

**Screenshots**
![selected-problem](https://files.gitter.im/red/bugs/iJJd/image.png)

**Platform version(please complete the following information):**
```
Red 0.6.3 for Windows built 11-Oct-2018/18:31:29+03:00 commit #1d43d77
```

hiiamboris:
It looks like Windows does support LF newlines. Supposedly we should do this after CreateWindow:
```
SendMessage handle 1510h 7 2
SendMessage handle 1512h 3 3
```
But bad news it's only supported on some newer W10 builds :D 
For the reference: https://docs.microsoft.com/en-us/windows/win32/controls/edit-control-window-extended-styles  https://docs.microsoft.com/en-us/windows/win32/api/commctrl/ne-commctrl-ec_endofline

toomasv:
There is still a problem with this. 
[![area-select](https://toomasv.red/images/Errors/area-select.gif)](https://toomasv.red/images/Errors/area-select.gif)

hiiamboris:
Interesting. Can't reproduce it on W7.

hiiamboris:
https://github.com/red/red/pull/4168 should fix it for good
Thanks for testing @toomasv 
I should have tested it more thoroughly, but it was so late at night ;)

PR #4152, opened 2019-11-24
FIX: issue #3563 (Line-breaks in area on Windows add...)

hiiamboris:
Test snippet:
`view [a: area "Some^/line^/here" button "Show" [print [length? a/text a/selected]] button "SEL" [a/selected: 1x0 + any [a/selected 0x1] print [length? a/text a/selected]]]`


qtxie:
Thanks for fixing it. :+1:

Issue #617, opened 2013-12-17
object! iteration missing


Kaj-de-Vos:
red>> foreach [x y] context [a: 'b] [?? x ?? y]

**\* Runtime Error 1: access violation
**\* at: 0809C4B1h

R3:

> > foreach [x y] context [a: 'b] [?? x ?? y]
> > x: a
> > y: b
> > == b

dockimbel:
I think it is worth considering the addition of `/into` refinement to all `*-of` accessors, as they are producing a new block (except `body-of` on functions). In such case, iterating over objects directly will not be needed anymore.

dockimbel:
Closing this ticket, as object semantics will be reexamined in the language spec work and iterability will be considered.

PR #4169, opened 2019-12-02
FIX: issue #3910 (Wrong error message when using foreach wrongly on map)

### LINKED ISSUE #3910: Wrong error message when using foreach wrongly on map
**Describe the bug**
`foreach` on a map value should provide a block of 2 words as counters. If only one word is given, the error message is not clear and even misleading

**To reproduce**
```
>> m: #(a: 1)
== #(
    a: 1
)
>> foreach p m []
*** Script Error: foreach does not allow map! for its 'word argument
*** Where: foreach
```

**Expected behavior**
```
*** Script Error: foreach requires a block of 2 words for its 'word argument
```

```
all
```

9214:
`foreach <word> <non-empty map> <block>` now reports an invalid argument:
```red
>> foreach value #(a b)[]
*** Script Error: invalid argument: value
*** Where: foreach
*** Stack:  

>> foreach value #()[]
```

Note that the last example is pertinent to _all_ cases with an empty `series` argument, regardless of its type:
```red
foreach [whatever #really /I don't/care][][]
```

This PR closes #3910. I'm not completely satisfied with the change though, as it makes `while [foreach-next][...]` loop less tight than it should be. An alternative solution is to move check into second `either` branch in `foreach*` before the loop kicks in.

Related tickets: https://github.com/red/red/issues/2279, https://github.com/red/red/issues/617, https://github.com/red/REP/issues/58.

### DIFF
--- a/runtime/natives.reds
+++ b/runtime/natives.reds
@@ -281,15 +281,17 @@ natives: context [
 	foreach*: func [
 		check? [logic!]
 		/local
-			value [red-value!]
-			body  [red-block!]
-			size  [integer!]
+			value  [red-value!]
+			series [red-value!]
+			body   [red-block!]
+			size   [integer!]
 	][
 		#typecheck foreach
 		value: stack/arguments
+		series: stack/arguments + 1
 		body: as red-block! stack/arguments + 2
 		
-		stack/push stack/arguments + 1					;-- copy arguments to stack top in reverse order
+		stack/push series								;-- copy arguments to stack top in reverse order
 		stack/push value								;-- (required by foreach-next)
 		
 		stack/mark-loop words/_body
@@ -310,6 +312,8 @@ natives: context [
 				]
 			]
 		][
+			if TYPE_OF(series) = TYPE_MAP [fire [TO_ERROR(script invalid-arg) value]]
+			
 			while [foreach-next][						;-- foreach <word!>
 				stack/reset
 				catch RED_THROWN_BREAK	[interpreter/eval body no]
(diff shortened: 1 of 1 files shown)


hiiamboris:
Why not put the check before this line?
https://github.com/red/red/blob/98b7537c03b25c49016b3f4f83690073c755afb1/runtime/natives.reds#L313)

9214:
@hiiamboris to keep the symmetry with `foreach-next-block` and to prevent `foreach value #()[...]` from erroring out. By putting check before the loop you get this:
```red
>> foreach [1 2 3]#()[] ; no error, although this is invalid
>> foreach word #()[] ; errors out on this, but not on the above one
*** Script Error: invalid argument: word
*** Where: foreach
*** Stack: 
```
Not saying that this is bad, just inconsistent with other datatypes.

hiiamboris:
Let's make block error out too? If it's not a block of exactly 2 words.

9214:
@hiiamboris "exactly 2 words" is relevant to `map!` only. Checking of block format needs to be done for all cases (see 2nd snippet in my original message) and IMO is not needed, as it (a) will error out if `series` argument is not empty (b) adds overhead; if it's empty then there's simply no iteration, and `'word` argument is ignored.
```red
>> foreach [1 2 3][][]
>> foreach [1 2 3][whoops][]
*** Script Error: invalid argument: 1
*** Where: foreach
*** Stack:  
````
Not for me to decide anyway; what this PR does is a fix of a single issue.

PR #4195, opened 2019-12-20
CHORE: refactor `wait` native

loziniak:
removes OS dependency from `runtime/natives.reds`

### DIFF
--- a/runtime/datatypes/structures.reds
+++ b/runtime/datatypes/structures.reds
@@ -321,7 +321,7 @@ red-date!: alias struct! [
 red-time!: alias struct! [
 	header 	[integer!]								;-- cell header
 	padding	[integer!]								;-- for compatibility with date!
-	time	[float!]								;-- 64-bit float
+	time	[float!]								;-- 64-bit float, in seconds
 ]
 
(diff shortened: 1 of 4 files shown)


9214:
[General contribution guidelines](https://github.com/red/red/wiki/%5BDOC%5D-Contributor-Guidelines#general-contribution-guidelines), p. 5.

loziniak:
It's a refactor, so neither a FIX, nor FEAT. Which one should I use?

hiiamboris:
I'm unsure `if time < 1.0 [time: 1.0]` is a good change. A value of zero is used in Windows to yield the thread's remaining time slice to other threads and may return immediately. Why forbid it?

greggirwin:
FEAT or FIX is sometimes a hard choice. Let's call this a FIX, since it's not adding anything new, just improving the implementation.

dockimbel:
What is this  "CHORE:" prefix doing in the commits history?

loziniak:
@greggirwin it's already merged, so probably too late for that? I'll try to remember that next time. Thanks.

PR #4225, opened 2020-01-07
FEAT: calendar widget

9214:
This PR introduces native monthly calendar widget (Win32 & Cocoa backends) and new `calendar` face which supports selection of a single `date!` value via `data` facet.

#### Relevant documentation

* Win32: [month calendar control](https://docs.microsoft.com/en-us/windows/win32/controls/month-calendar-control-reference);
* Cocoa: [NSDatePicker](https://docs.microsoft.com/en-us/windows/win32/controls/month-calendar-control-reference) (calendar is its graphical style);
* View: https://github.com/red/docs/pull/205.

#### Known bugs and features

* `loose` on macOS is broken, but that seems to be a generic issue with Cocoa https://github.com/red/red/issues/4226;
* `color` is supported on Windows (changes background color), but not on macOS;
* increasing calendar's size on Windows changes its rendering style, but has no effect on macOS.

#### GTK support

@bitbegin see [here](https://developer.gnome.org/gtk3/stable/GtkCalendar.html).


hiiamboris:
Works like a charm on W7.
Although there's some random white border around it:

9214:
@hiiamboris looks like there are some extra margins (so you see a background), although the calendar is of the minimum possible size and should not have any.

image

I pushed a fix that sets default background to system-specific color (the same that `panel` has), see if it works for you.

hiiamboris:
Naisu! :+1:

PR #4255, opened 2020-01-22
FIX: [Windows macOS] issues #4244, #4245 and #4246

### LINKED ISSUE #4244: [View] incorrect treatment of logical values in RADIO and CHECK
**Describe the bug**

When it comes to logical values, Red's motto is "everything is true except `false` and `none`". This does not hold in case of `radio` and `check` buttons:

* on creation, both treat only `true` as truthy and everything else as falsey;
* after creation, `check` treats `none` as truthy.

**To reproduce**

On creation:

```red
view [radio data 'a check data 'b]
```

After creation:

```red
view [c: check button [c/data: none]]
```

Clicking on a button sets a checkbox to a toggled state.

This issue affects all View backends.

**Expected behavior**

I expect these widgets to follow the standard Red semantics. `data` facet should contain only `logic!` (which switches widget into toggled/untoggled state), and any other value's datatype should be overridden to it.

**Platform version**

The latest release, all platforms.

### LINKED ISSUE #4245: [View] CHECK treats NONE as truthy
**Describe the bug**

Setting `data` facet of `check` face to `none` switches it to a toggled state.

**To reproduce**

```red
view [c: check data none button [c/data: none]]
```

Note that pre-setting it with `none` has no effect, facet needs to be set after low-level instantiation of the face. This issue affects all View backends.

**Expected behavior**

In fact, `none` is supposed to indicate the 3rd state for a tri-state checkbox. This functionality is available on all platforms (see [here](https://github.com/red/red/issues/4244#issuecomment-575745396)), but currently is not properly implemented.

**Platform version**

The latest release, all platforms.

### LINKED ISSUE #4246: [View] RADIO emits superfluous events on Windows and macOS
**Describe the bug**

Toggled `radio` face, when clicked on:

* On Windows: switches between off/on states and emits two `change` events for each.
* On macOS: generates one `change` event.

**To reproduce**

```red
view [radio [probe face/data]]
```

Click on it multiple times and it will print out `true`, despite no actual change taking place in `data` facet. On Windows, it often prints `false` and `true` at once.

**Expected behavior**

I expect toggled `radio` to emit no events when clicked on. Or, if that behavior is consistent with the design of other widgets, then `radio` on GTK should always emit an event when clicked on, regardless of its state. In any case, on Windows, there shouldn't be any double events and state switching.

**Platform version**

The latest release, Windows 10 and macOS 10.14 (Mojave).

9214:
Fixes #4244, closes #4245 and resolves #4246. In addition to that, this PR properly implements 3-state mode for `check` face (described [here](https://github.com/red/docs/pull/211)).

For the reference: the state-transitioning behavior of 3-state check box widget is the same on all platforms
```
→ indeterminate → checked → unchecked →
```

But at least on Windows 10 by default it is
```
← indeterminate ← checked ← unchecked ←
```

Which

* Violates [documented](https://docs.microsoft.com/en-us/windows/win32/controls/button-types-and-styles) behavior, and, as such, does not match the user's expectations:
  > Repeatedly clicking a three-state check box toggles it from checked to cleared to indeterminate     and then repeats the cycle.
* Is inconsistent with two other platforms.

So I had to manually re-map states on each `BN_CLICKED` notification for a 3-state check box.


modules/view/backends/windows/events.reds
```diff
@@ -638,18 +640,46 @@ process-command-event: func [
...
+			unless zero? evt [make-event current-msg 0 evt]	;-- should be *after* get-facet call (Windows closing on click case)
 		]
 		BN_UNPUSHED [
 			type: as red-word! get-facet current-msg FACE_OBJ_TYPE
 			if type/symbol = radio [
 				current-msg/hWnd: child					;-- force child handle
-				make-event current-msg 0 EVT_CHANGE
+				
+				;-- ignore double-click (fixes #4246)
+				unless BST_UNCHECKED <> as integer! SendMessage child BM_GETSTATE 0 0 [
```
hiiamboris: man you're clearly abusing negations here ;)

modules/view/backends/windows/gui.reds
```diff
@@ -937,15 +937,27 @@ set-logic-state: func [
 	state  [red-logic!]
```
hiiamboris: I would suggest `red-value!` type as you're passing `none` as well.

9214: So you'd rather return `0` and `1` instead of `true` and `false` in `switch`, just to save an extra `integer!` cast at the expense of clarity? The function is called `set-LOGIC-state` and had this interface on all backends way before my PR.

modules/view/backends/windows/events.reds
9214: You're welcome to come up with a better solution.

hiiamboris:
I can confirm the order on W7 is as you designed: 
> → indeterminate → checked → unchecked →

modules/view/backends/windows/events.reds
hiiamboris: I don't know it all looks like jumping through the hoops to me... If it (radio) is UNPUSHED, means it was (maybe) checked during PUSHED event, right? Then why we even bother with reacting to UNPUSH event? But I'm lazy to recompile it and see... So I'd take your word on it.
I'm just sayin that `if =` reads better than `unless <>` :/

PR #4277, opened 2020-02-01
FIX: issue #4271 ([GTK] NO-BUTTONS window flag has no effect)

### LINKED ISSUE #4271: [GTK] NO-BUTTONS window flag has no effect
**Describe the bug**

`no-buttons` flag removes all window buttons on Windows and macOS, but has no effect on GTK.

**To reproduce**

```red
view/flags [] 'no-buttons
```

`no-min` and `no-max` also have no effect.

**Expected behavior**

I expect all of the buttons to be removed.

**Platform version**

```
Red 0.6.4 for Linux built 30-Jan-2020/13:22:38+01:00
```

9214:
This PR fixes #4271, but only partially and in a two-fold manner, here's why:

1. Window decoration flags can be set separately with combination of [`gdk-window-set-decorations`](https://developer.gnome.org/gdk3/stable/gdk3-Windows.html#gdk-window-set-decorations) and [`GdkWMDecoration`](https://developer.gnome.org/gdk3/stable/gdk3-Windows.html#GdkWMDecoration) in `realize` signal handler (i.e. after allocation of low-level GDK resources), but it still leaves window at the mercy of a window manager, which is free to ignore all the (un)set bits when applying decorations; per documentation referenced above:

   > Most window managers honor a decorations hint of 0 to disable all decorations, but very few honor all possible combinations of bits.

1. GTK itself offers only coarse-grained management of window's appearance:
   * [`gtk-window-set-decorated`](https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-set-decorated) to disable decorations altogether (i.e. remove window's frame with decoration hint of 0);
   * [`gtk-window-set-type-hint`](https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-set-type-hint) with [`GdkWindowTypeHint`](https://developer.gnome.org/gdk3/stable/gdk3-Windows.html#GdkWindowTypeHint) to prompt WM to decorate window according to its functional purpose (i.e. window's type);
   * [`gtk_window_set_deletable`](https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk_window_set_deletable) to mark window as 'deletable', hinting WM to hide 'Close' button.

Consequently, https://github.com/red/red/commit/df43497db2387a4a5193dc68fd3fa986e9891c75 addresses **(1)** and sets decoration hints manually in GDK window, while **(2)** https://github.com/red/red/commit/a21f9a270148757f115636a82ba2ad9792cd7c9d relies on high-level GTK API.

Testing on `xfwm4` gave the following results:
* Most of the manually set hint combinations are not honored, rendering **(1)** ineffective;
* `gtk_window_set_deletable` hides 'Close' button; 
* `GDK_WINDOW_TYPE_HINT_UTILITY` window type hint hides 'Min' button.

What this means with regard to View is:
* `no-buttons` removes only 'Close' and 'Min' buttons;
* `no-min` removes 'Min' button;
* `no-max` is ignored since it's apparently impossible to hide the 'Max' button when the window is marked as resizable.

Examples:

| `resize` | — | `no-buttons` | `no-min` | `no-max` |
|:-:|-|-|-|-|
| ✗ | image | image | image | image |
| ✓ | image | image | image | image |

---

This PR can be merged when:

- [ ] Interested contributors tested these changes on their WM of choice and reported back;
- [ ] If **(1)** has no effect, then I'll revert the respective commit, to avoid cluttering codebase with redundant code;
- [ ] If there are any significant differences across WMs, the least we can do is to document them.

### DIFF
--- a/modules/view/backends/gtk3/events.reds
+++ b/modules/view/backends/gtk3/events.reds
@@ -1024,6 +1024,7 @@ connect-widget-events: func [
 			gobj_signal_connect(widget "focus-in-event" :focus-in-event widget)
 			gobj_signal_connect(widget "focus-out-event" :focus-out-event widget)
 			gobj_signal_connect(widget "configure-event" :window-configure-event widget)
+			gobj_signal_connect(widget "realize" :window-realize widget)
 		]
 		sym = slider [
 			gobj_signal_connect(widget "value-changed" :range-value-changed widget)
--- a/modules/view/backends/gtk3/gtk.reds
+++ b/modules/view/backends/gtk3/gtk.reds
@@ -1278,6 +1278,10 @@ GPtrArray!: alias struct! [
 			window		[handle!]
 			mode		[logic!]
 		]
+		gtk_window_set_type_hint: "gtk_window_set_type_hint" [
+			window		[handle!]
+			mode		[integer!]
+		]
 		gtk_window_move: "gtk_window_move" [
 			window		[handle!]
 			x			[integer!]
@@ -2930,7 +2934,10 @@ GPtrArray!: alias struct! [
 			win 		[handle!]
 			cursor		[handle!]
 		]
-
+		gdk_window_set_decorations: "gdk_window_set_decorations" [
+			window		[handle!]
+			flags		[integer!]
+		]
 		;; Useless since already called inside pango_cairo_create_context
 		; pango_cairo_font_map_get_default: "pango_cairo_font_map_get_default" [
 		; 	return: 	[handle!]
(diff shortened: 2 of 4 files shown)


bitbegin:
image
image
image
image
image

ubuntu 18.04

bitbegin:
better than before

9214:
Thanks @bitbegin, can you also check all these flags with `resize`, to see if 'Max' button gets hidden?

bitbegin:
image
image
image

9214:
@bitbegin so my **(1)** workaround is ignored on Ubuntu's WM also... @qtxie I think you merged a bit too earlier, as I haven't cleaned up the redundant code per my comments :wink: will do it in another PR.

Issue #4126, opened 2019-11-07
#[type!] support is quite narrow


hiiamboris:
**Describe the bug**

Direct loading of datatype! values is only supported for a few types:
```
>> foreach type to block! any-type! [x: attempt [load rejoin [ "#[" type "]" ]] print [pad type 15 "=>" mold x]]
datatype!       => none
unset!          => none
none!           => none!
logic!          => logic!
block!          => block!
paren!          => paren!
string!         => string!
file!           => none
url!            => none
char!           => char!
integer!        => integer!
float!          => none
word!           => word!
set-word!       => set-word!
lit-word!       => lit-word!
get-word!       => get-word!
refinement!     => refinement!
issue!          => issue!
native!         => native!
action!         => action!
op!             => op!
function!       => function!
path!           => path!
lit-path!       => lit-path!
set-path!       => set-path!
get-path!       => get-path!
routine!        => routine!
bitset!         => bitset!
object!         => none
typeset!        => none
error!          => none
vector!         => none
hash!           => none
pair!           => none
percent!        => none
tuple!          => none
map!            => none
binary!         => none
time!           => none
tag!            => none
email!          => none
handle!         => none
date!           => none
port!           => none
image!          => none
event!          => none
```

**Expected behavior**

Every known datatype supported.

**Platform version (please complete the following information)**
```
Red 0.6.4 for Windows built 18-Oct-2019/15:33:41+03:00 commit #ed913ef
```

9214:
This is expected, since construction syntax is far from being complete.

meijeru:
See #652 and #3666. At the time of #652 (back in 2014) this was squarely dismissed. I think some patience is required.

hiiamboris:
This doesn't look like a big deal to me. Just add stuff to `escaped-rule` of lexers. It shouldn't take 5 years to do so.

meijeru:
Frankly, compared to the FULL implementation of value constructors (`#[<type> <spec>]`), this is indeed a minor thing. It is that full implementation that is going to bring real benefits, also in connection to `mold`.

greggirwin:
I don't imagine a PR for handling all datatypes will be rejected, but it's _possible_. As noted, it's easy and doesn't require any design work. The reason it likely hasn't been done is that it hasn't been needed, and won't really be important until we have the general solution defined. There was a time, before `map!` syntax was decided on, that changing construction syntax was an open design question, and may still be. 

[...]

PR #4279, opened 2020-02-02
FEAT: support for handle loading

hiiamboris:
Addresses https://github.com/red/red/issues/4126#issuecomment-581180069 by making handle! loadable after mold:
```
>> mold/all load mold/all load "#[handle! 10h]"
== "#[handle! 00000010h]"
```

I didn't change the R2 lexer. Not sure how - no handles in R2. Or if there's a point it that.

### DIFF
--- a/environment/lexer.red
+++ b/environment/lexer.red
@@ -262,6 +262,18 @@ system/lexer: context [
 		n
 	]
 
+	make-handle: routine [
+		start	[string!]
+		end		[string!]
+		/local
+			n	  [integer!]
+			value [red-value!]
+	][
+		n: make-hexa start end
+		value: as red-value! integer/box n
+		set-type value TYPE_HANDLE
+	]
+
 	make-char: routine [
 		start	[string!]
 		end		[string!]
@@ -870,6 +882,7 @@ system/lexer: context [
 					| "routine!"	(value: routine!)
 				]
 				| "none" 			(value: none)
+				| "handle!" some ws s: hexa-rule (value: make-handle s e)
 			] pos: any ws #"]"
 		]
 


Oldes:
I don't think that `handle` should be loadable! Only `null` handle could be, so system could recreate it if needed, but even with it I'm not so sure.

hiiamboris:
And then how do you propose to save and load data containing handles?

Oldes:
Even more... I don't think that `handle` should display it's value (if not used in some debug mode). But internally `handle` should hold it's type, so for example when I use `handle` to hold internal `sound`, I will not be allowed to use such a handle as a `window` pointer (which is case with View).

Oldes:
Loading data which contains `handle` should convert this value to `none` or create `handle` with `null` value.

hiiamboris:
IMO there are tons of use cases for accessing values of handles, some of which I'm using myself. From adding features to the View subsystem or altering it's behavior, to communicating OS objects between processes, to dumping the data for debugging and who knows what - it depends on how many values eventually become handles in Red.

OTOH, loading a handle as zero is a straight way to the crash.

Oldes:
My opinion is, that you should never directly access handle's value as it is internal thing. I don't say that loading handle as a zero is the best thing. Of course that the internal implementation would have to handle such a `handle` as it would have to handle `none` value in place of handle.

Oldes:
And I should note, that it was me, who bring `handle!` type into `Red` as I was heavy `handle` user in all my bindings, which are still part of the [red/code](https://github.com/red/code/tree/master/Library).

I still think, that current `handle` is not good enough. @dockimbel wanted it to be just a immutable `integer!`, but if there would be more handle types used, these should carry its own type and one should be able to specify, that for example function `play-bass-sound` is accepting `handle` of type `bass-sound!` and not `sqlite-database!` handle.

And internally `handle` could work with GC and be able to have registered destructor per its type, which GC could call in case that handle is not used anymore, so one would not have to be responsible to manually call routines just to free resources.

hiiamboris:
I support the typed handle idea. And GC - in case the memory was allocated by Red. Although currently GC only sees series.

hiiamboris:
Another example of where loadable handle comes handy: https://gitter.im/red/sandbox?at=5ed75b50b101510b2034c012

Oldes:
> Another example of where loadable handle comes handy: https://gitter.im/red/sandbox?at=5ed75b50b101510b2034c012

How is it related?

hiiamboris:
@Oldes There's an `area` that lists the face contents. This area is loaded back when changed. Toomas excluded `state` from it for simplicity. 

But ideally if you want a live object editor, you would have to scan the whole object deeply for handles, figure out how to change the data around them, but not touch them.
And then, what if you have e.g. a block `[a: some face  b: other face]` and you edit this block live by exchanging values of `a` and `b`. How would you preserve handles in this case? As usual, exceptions destroy simple solutions.

dockimbel:
`handle!` is an opaque low-level reference by definition. There should be no direct way to create it from Red level, only from R/S. As its created during a Red session, it's specific to that session, so there is no point in serializing it, nor displaying it (except for debugging purposes).

Associating a low-level type information to each handle would be a useful improvement.

PR #4283, opened 2020-02-04
FEAT: fetch-next supports set-words/paths, get/lit-args, objects...

hiiamboris:
Extension to `preprocessor/fetch-next` in the light of :point_up: [February 1, 2020 6:15 PM](https://gitter.im/red/bugs?at=5e3596233aca1e4c5f5a11d4)

### DIFF
--- a/utils/preprocessor.r
+++ b/utils/preprocessor.r
@@ -136,13 +136,20 @@ preprocessor: context [
 		arity
 	]
 
-	value-path?: func [path [path!] /local value i] [
+	value-path?: func [path [path!] /local value i item] [
 		repeat i length? path [
-			set/any 'value either i = 1 [get/any first path][select value pick path i]
-			unless any [								;-- select-able types
-				series? get/any 'value
-				find [object! port! error! map!] type?/word get/any 'value
-			][
+			set/any 'value either i = 1 [get/any first path][
+				set/any 'item pick path i
+				case [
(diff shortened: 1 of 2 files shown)


utils/preprocessor.r
```diff
@@ -122,39 +128,88 @@ preprocessor: context [
...
+				case [
+					get-word? :item [set/any 'item get/any to word! item]
+					paren?    :item [set/any 'item do item]
+				]
+				either integer? :item [pick value item][select value :item]
+			]
+			unless find [								;-- select-able types
+				block! paren! path! lit-path! set-path! get-path!
+				object! port! error! map!
+			] type?/word get/any 'value [
```
dockimbel: You could use a `typeset!` here to make `find` perform in constant time.

dockimbel:
Great work! The changes look good, though the tests on Linux are failing.

utils/preprocessor.r
hiiamboris: Yes, good idea!

hiiamboris:
Thanks for the review @dockimbel ☻
Must be OK now.

loziniak:
Perhaps `image!` tests for Linux could be moved to GTK branch?

hiiamboris:
@loziniak it doesn't matter in this case. I just wanted a path that cannot be `select`-ed. `image/size` was just the 1st thing that came to my mind, by `pair/y` works the same.

PR #4295, opened 2020-02-09
FIX: libRed.red was not updated if libRed directory exists

endo64:
This PR fixes the below issue:

When using encapped Red toolchain to build `libRed` (`red build libRed`), if there is already a folder named libRed, files inside the folder were not updated to the files in the repo, hence the `libRed.dll` built was not the latest.


endo64:
Holly crap, messed with the commits, need to close this PR and re-issue a clean one.

PR #4296, opened 2020-02-15
FIX: support `match tag!` in console engine

bitbegin:


dockimbel:
@bitbegin @qtxie Why are `block-rule`, `paren-rule` and `tag-rule` duplicating so much same code? Why not factorize all that rather than copy-pasting?

bitbegin:
refactored by this https://github.com/red/red/pull/4435

Issue #4205, opened 2019-12-24
RANDOM/SEED is not satisfactory


hiiamboris:
**Describe the bug**

Run this:
```
Red []                                
                                      
loop 10 [
	random/seed form now/time/precise
	probe random 10000
]
loop 10 [
	random/seed now/time/precise
	; random/seed (t: now/time/precise t/second)    ;) same
	wait 1e-3
	print [t random 10000]
]
```

Here's 3 results:
```
2753            
3781            
1267            
2651            
4113            
4804            
2503            
2771            
7648            
9492            
0:25:19.7836 479
0:25:19.7846 479
0:25:19.7856 479
0:25:19.7866 479
0:25:19.7876 479
0:25:19.7886 479
0:25:19.7896 479
0:25:19.7906 479
0:25:19.7916 479
0:25:19.7926 479
```
```
2753             
3781             
1267             
2651             
4113             
4804             
2503             
2771             
7648             
9492             
0:26:05.2872 7661
0:26:05.2882 7661
0:26:05.2892 7661
0:26:05.2902 7661
0:26:05.2912 7661
0:26:05.2922 7661
0:26:05.2932 7661
0:26:05.2942 7661
0:26:05.2952 7661
0:26:05.2962 7661
```
```
2753             
3781             
1267             
2651             
4113             
4804             
2503             
2771             
7648             
9492             
0:26:16.4108 1290
0:26:16.4118 1290
0:26:16.4128 1290
0:26:16.4138 1290
0:26:16.4148 1290
0:26:16.4158 1290
0:26:16.4168 1290
0:26:16.4178 1290
0:26:16.4188 1290
0:26:16.4198 1290
```

So, as can be seen:
- /seed does not use fractional part of seconds when given `time!`
- /seed does nothing at all when given a `string!` (look how results are all the same thoughout all program runs)

**Expected behavior**

- either it should return an error when given an argument it can't seed with, or it should use it in full
- should use the fractional part of `float!` and `time!`

**Platform version (please complete the following information)**
```
Red 0.6.4 for Windows built 20-Dec-2019/19:03:46+03:00 commit #544a6e1
```

loziniak:
```
>> loop 10 [probe to binary! to float! now/time/precise wait 1e-3]
#{40E955695B8348F5}
#{40E95569659AF9EC}
#{40E955696FAA4767}
#{40E9556979ACFFA8}
#{40E9556983B3E9A7}
#{40E955698DAE3E6C}
#{40E9556998288052}
[...]

PR #4297, opened 2020-02-20
FIX: seeding random with time! and float! issue #4205

loziniak:


hiiamboris:
@loziniak maybe better `sp/1 xor sp/2`?

loziniak:
I thought about it. But couldn't find any arguments for or against. Do you have any?

hiiamboris:
1) harder to find repeatable seeds (e.g. `Me+N` may produce the same same seed for all N or something alike)
2) big-endian support, where the order of ints within a float is reversed

loziniak:
Isn't it that `sp/1` is always the same as `sp/value`?

hiiamboris:
In my examples - yes. Generally - no.

loziniak:
I misread your first post, didn't notice the *xor*, just thought you want me to replace `sp/value` with `sp/1`, sorry. Sure, xoring is great idea.

qtxie:
@loziniak Nice fix! Thank you! 👍

PR #4300, opened 2020-02-23
FEAT: garbage collect images on Win32 backend

hiiamboris:
Implementation of garbage collection of images for Win32.

Background: image data on Windows is stored outside of Red memory segment, in chunks allocated by the GDI+, to which the only links are the image handles. GC has no knowledge of any external resources associated with the data (in the future this can be audio/video data or anything).

I extend memory management with the ability to call a custom deallocation routine, when a node is about to be freed: 
https://github.com/red/red/blob/3ad9ed799f22687c402dca3f0f7af86e35690516/runtime/allocator.reds#L354-L359

For that I have to utilize another bit in series flags:
https://github.com/red/red/blob/3ad9ed799f22687c402dca3f0f7af86e35690516/runtime/definitions.reds#L39

Deallocator spec:
https://github.com/red/red/blob/3ad9ed799f22687c402dca3f0f7af86e35690516/runtime/allocator.reds#L21
Deallocator has to follow the series buffer. All the other series content is up to a specific datatype to define and use. Image type only holds an OS handle there.

So for it to work I have to:
1) hold image handle in the data referred by a node
2) make node a series to make GC aware of it (previously it was a bare GDI+ handle)

Image node on Win32 becomes a series with 2 values:
https://github.com/red/red/blob/3ad9ed799f22687c402dca3f0f7af86e35690516/runtime/platform/image-gdiplus.reds#L53-L56
This brings it closer to MacOS backend where image node is also a series.
It is formally empty here, as it's offset points to it's tail.

A node is created from a handle using:
https://github.com/red/red/blob/3ad9ed799f22687c402dca3f0f7af86e35690516/runtime/platform/image-gdiplus.reds#L512-L524

The rest just makes it all work together.

One unsolved trouble here is that I haven't found a good spot to place a trigger for `collector/do-cycle`. When `recycle` kicks in it collects images properly, but if I do not call it myself while allocating a lot of images, it may run out of memory because it's not aware of any memory used by the GDI+.
Currently `do-cycle` only runs when it fills current series frame and has to compact it, and since image is seen as 8-byte series it's not gonna happen. Putting `do-cycle` into `image/make` is an overkill - `do-cycle` is too slow for that. So for now, I'm leaving it up to the user.

### DIFF
--- a/modules/view/backends/windows/base.reds
+++ b/modules/view/backends/windows/base.reds
@@ -109,6 +109,7 @@ render-base: func [
 		graphic	[integer!]
 		type	[integer!]
 		res		[logic!]
+		inode	[img-node!]
 ][
 	graphic: 0
 	res: paint-background hWnd hDC
@@ -119,11 +120,10 @@ render-base: func [
 
 	GetClientRect hWnd :rc
 	if TYPE_OF(img) = TYPE_IMAGE [
+		inode: as img-node! (as series! img/node/value) + 1
 		GdipCreateFromHDC hDC :graphic
 		if zero? GdipDrawImageRectI
-			graphic
-			as-integer img/node
-			0 0
+			graphic inode/handle 0 0
 			rc/right - rc/left rc/bottom - rc/top [res: true]
 		GdipDeleteGraphics graphic
 	]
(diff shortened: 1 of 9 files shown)


greggirwin:
The explanation to go with the PR is fantastic. :+1:

dockimbel:
@hiiamboris Great work! 

Could you please rebase this PR and solve the conflicts with master?

hiiamboris:
@dockimbel Sure, I'll try to find some time.

hiiamboris:
Plz fix the [`** Syntax Error: Missing ] at end-of-script` regression first](https://api.travis-ci.org/v3/job/731024267/log.txt)

hiiamboris:
@dockimbel rebased, see https://github.com/red/red/pull/4662

PR #4309, opened 2020-02-28
FEAT: random/secure

loziniak:
Based on `crypto/urandom'. [Chat @ Gitter](https://gitter.im/red/blockchain?at=5e4e388d0c4514126679983b)

### DIFF
--- a/runtime/datatypes/tuple.reds
+++ b/runtime/datatypes/tuple.reds
@@ -251,7 +251,6 @@ tuple: context [
 			array [byte-ptr!]
 			n	  [integer!]
 			size  [integer!]
-			rnd	  [integer!]
 	][
 		#if debug? = yes [if verbose > 0 [print-line "tuple/random"]]
 
@@ -264,9 +263,8 @@ tuple: context [
 			size: TUPLE_SIZE?(tp)
 			n: 0
 			until [
-				rnd: either secure? [_random/rand-secure] [_random/rand]
 				n: n + 1
-				array/n: as-byte rnd % ((as-integer array/n) + 1)
+				array/n: as-byte (-1 + _random/int-uniform-distr secure? as-integer (array/n + 1))
 				n = size
 			]
 		]
--- a/runtime/datatypes/tuple.reds
+++ b/runtime/datatypes/tuple.reds
@@ -264,7 +264,9 @@ tuple: context [
 			n: 0
 			until [
 				n: n + 1
-				array/n: as-byte _random/rand % ((as-integer array/n) + 1)
+				array/n: as-byte
+					(either secure? [_random/rand-secure] [_random/rand])
+					% ((as-integer array/n) + 1)
 				n = size
 			]
 		]
--- a/runtime/datatypes/tuple.reds
+++ b/runtime/datatypes/tuple.reds
@@ -251,6 +251,7 @@ tuple: context [
 			array [byte-ptr!]
 			n	  [integer!]
 			size  [integer!]
+			rnd	  [integer!]
 	][
 		#if debug? = yes [if verbose > 0 [print-line "tuple/random"]]
 
@@ -263,10 +264,9 @@ tuple: context [
 			size: TUPLE_SIZE?(tp)
 			n: 0
 			until [
+				rnd: either secure? [_random/rand-secure] [_random/rand]
 				n: n + 1
-				array/n: as-byte
-					(either secure? [_random/rand-secure] [_random/rand])
-					% ((as-integer array/n) + 1)
+				array/n: as-byte rnd % ((as-integer array/n) + 1)
 				n = size
 			]
 		]
--- a/runtime/datatypes/tuple.reds
+++ b/runtime/datatypes/tuple.reds
@@ -264,7 +264,7 @@ tuple: context [
 			n: 0
 			until [
 				n: n + 1
-				array/n: as-byte (-1 + _random/int-uniform-distr secure? as-integer (array/n + 1))
+				array/n: as-byte (-1 + _random/int-uniform-distr secure? (1 + as-integer array/n))
 				n = size
 			]
 		]
--- a/runtime/random.reds
+++ b/runtime/random.reds
@@ -88,6 +88,16 @@ _random: context [
 		state: state >> 18 xor state
 		state
 	]
+
+	rand-secure: func [
+		return:	[integer!]
+		/local
+			i	 [integer!]
+	] [
+		i: 0
+		crypto/urandom (as byte-ptr! :i) 4
+		i and 7FFFFFFFh									;-- range 0-2147483647 inclusive
+	]
 	
 	init: does [
 		table: as int-ptr! allocate MT_RANDOM_STATE_SIZE * size? integer!
--- a/runtime/random.reds
+++ b/runtime/random.reds
@@ -98,7 +98,37 @@ _random: context [
 		crypto/urandom (as byte-ptr! :i) 4
 		i and 7FFFFFFFh									;-- range 0-2147483647 inclusive
 	]
-	
+
+	int-uniform-distr: func [
+		secure?	[logic!]
+		max		[integer!]
+		return:	[integer!]
+		/local
+			mask	 [integer!]
+			rnd		 [integer!]
+			neg?	 [logic!]
+	] [
+		if max = 0 [return 0]
+		either max < 0 [
+			max: 0 - max
+			neg?: true
+		] [
+			neg?: false
+		]
+		mask: 7FFFFFFFh >> (30 - (as-integer floor log-2 as-float (max - 1)))
+		until [
+			rnd: either secure? [rand-secure] [rand]
+			rnd: rnd and mask
+			max > rnd
+		]
+		rnd: rnd + 1
+		either neg? [
+			0 - rnd
+		] [
+			rnd
+		]
+	]
+
 	init: does [
 		table: as int-ptr! allocate MT_RANDOM_STATE_SIZE * size? integer!
 		srand 19650218
--- a/runtime/random.reds
+++ b/runtime/random.reds
@@ -104,7 +104,7 @@ _random: context [
 		max		[integer!]
 		return:	[integer!]
 		/local
-			mask	 [integer!]
+			limit	 [integer!]
 			rnd		 [integer!]
 			neg?	 [logic!]
 	] [
@@ -115,13 +115,12 @@ _random: context [
 		] [
 			neg?: false
 		]
-		mask: 7FFFFFFFh >> (30 - (as-integer floor log-2 as-float (max - 1)))
+		limit: 2147483647 / max * max
 		until [
 			rnd: either secure? [rand-secure] [rand]
-			rnd: rnd and mask
-			max > rnd
+			limit > rnd
 		]
-		rnd: rnd + 1
+		rnd: rnd % max + 1
 		either neg? [
 			0 - rnd
 		] [
--- a/system/runtime/libc.reds
+++ b/system/runtime/libc.reds
@@ -135,7 +135,11 @@ Red/System [
 			value		[float!]
 			return:		[float!]
 		]
-		log-2:		"log" [
+		log-e:		"log" [
+			value		[float!]
+			return:		[float!]
+		]
+		log-2:		"log2" [
 			value		[float!]
 			return:		[float!]
 		]
--- a/system/runtime/libc.reds
+++ b/system/runtime/libc.reds
@@ -139,10 +139,6 @@ Red/System [
 			value		[float!]
 			return:		[float!]
 		]
-		log-2:		"log2" [
-			value		[float!]
-			return:		[float!]
-		]
 		sqrt:		"sqrt" [
 			value		[float!]
 			return:		[float!]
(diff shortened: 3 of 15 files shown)


loziniak:
I'd like to include tests for this. My idea is:
* check whether a result is in range
* check if results don't repeat (take 10 random numbers)

But it's good for `random`. I don't have an idea how to test "security".

loziniak:
Second concern: should `random/secure` be seedable? In my implementation, if it's seeded with same number, sequences of generated numbers are different.

hiiamboris:
There's an issue, not with this PR but with the use of random in general. So @loziniak while you're at it, I hope you'll fix it ;)

The issue is that `random % size` idea is absolutely wrong!
Random values are supposed to be [distributed uniformly](https://en.wikipedia.org/wiki/Uniform_distribution_%28continuous%29), but remainder breaks this expectation.

A simple example.
Our random returns uniformly distributed value in the range [0 .. 2,147e9]. What happens if I'm trying to use `random 1'000'000'000`? Values [0 .. 1e9) will be mapped uniformly across the requested range. Values [1e9 .. 2e9) as well. But values [2e9 .. 2,147e9] will only cover the range [0 .. 147'000'000]. So I will get a value in the range [0 .. 147'000'000] more often (by like 7%) than in the range [147'000'001 .. 1e9]. To make it worse, `random 1'432'000'000` will generate twice(!) more values in the range [0 .. 715'600'000) than in the range [715'600'001 .. 1'432'000'000).

To mitigate this issue I would suggest a solution like this (pseudo code, just to show the idea):
```
limit: 2147483647 / size * size
until [
   x: random
   x < size
]
x: x % size
```
This would ensure that the last segment that does not uniformly map to the requested region, is discarded.

loziniak:
@hiiamboris I see one problem with this approach. In case of some numbers, like 1'432'000'000, it could require more repetitions, than in some others. In extreme case, function will never exit, because all numbers in sequenc would be from `((2147483647 / size * size), 2147483647]` range, which is theoretically possible.

hiiamboris:
@loziniak the worst case I foresee is `random 4'000'000h` where it would require an average of two random numbers generated per single result. All other sizes will require no more than that, with small sizes requiring almost always a single random number.

That extreme case is quite a problem though... I'm not sure what is the repeat period of RNG used by Red, but let's pretend it's a true RNG and do some math... What will be the probability of it missing 1000 times in a row in a 50/50 situation?
```
>> mold 0.5 ** 1000
== "0.9332636185032189e-301"
```
1e-301. Cool!
We're able to produce about 4M random numbers per second per CPU core (on my i5 ☻). Suppose Red is super widespread and our civilization loves to crunch numbers. So we dedicate 1M cores for that working 24/7. That makes, per our planet, 4M*1M*time(seconds), i.e. 1.2e20 numbers per year. So it would take 1e281 years to witness such an event (which will be resolved in less than 1 ms anyway). That's still 1e271 times more than the life of our universe... So... ☺ I say it's worth the risk ;)

I'd add that this is not necessary for tuple randomization as non-uniformity would be at most (1 / 16'777'216) which nobody will notice. But for integers I think it's a must ;)

loziniak:
Moreover, it seems that rejection sampling is the only known method to get uniform distribution in our case:
https://cs.stackexchange.com/questions/2605/is-rejection-sampling-the-only-way-to-get-a-truly-uniform-distribution-of-random

loziniak:
Perhaps a limit on retries would be good? Let's say at most 1000 and then error? With Red's small footprint it's gonna be used on wide range of devices, including ones with low-end CPU. Perhaps there are some clues on probability of generating numbers from certain range in row by `_random/rand` or `_random/rand-secure`?

BTW Red's PRNG is a Mersenne Twister with period of *2^19937 − 1*.

hiiamboris:
A theoretically better solution from those SO links (good reads there, but I only skimmed briefly):

Although it would be silly to fight for these bits of efficiency while making a Red `random` call (which is like 1000 times slower than R/S).

> Perhaps a limit on retries would be good? Let's say at most 1000 and then error?

I don't see a reason for that

system/runtime/libc.reds
```diff
@@ -135,7 +135,11 @@ Red/System [
 			value		[float!]
 			return:		[float!]
 		]
-		log-2:		"log" [
+		log-e:		"log" [
+			value		[float!]
+			return:		[float!]
+		]
+		log-2:		"log2" [
```
9214: This does at least two things:
* breaks most of the logarithm-related [natives](https://github.com/red/red/blob/master/runtime/natives.reds#L1794);
* renders Windows build inoperable — there's [no](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/log-logf-log10-log10f?view=vs-2019) `log2` entry point in `msvcrt`.

runtime/datatypes/tuple.reds
```diff
@@ -264,9 +263,8 @@ tuple: context [
 			size: TUPLE_SIZE?(tp)
 			n: 0
 			until [
-				rnd: either secure? [_random/rand-secure] [_random/rand]
 				n: n + 1
-				array/n: as-byte rnd % ((as-integer array/n) + 1)
+				array/n: as-byte (-1 + _random/int-uniform-distr secure? as-integer (array/n + 1))
```
hiiamboris: there will be a problem here with array/n = 255

runtime/random.reds
```diff
@@ -98,7 +98,37 @@ _random: context [
...
+			neg?	 [logic!]
+	] [
+		if max = 0 [return 0]
+		either max < 0 [
+			max: 0 - max
+			neg?: true
+		] [
+			neg?: false
+		]
+		mask: 7FFFFFFFh >> (30 - (as-integer floor log-2 as-float (max - 1)))
```
hiiamboris: can this be `(30 - log-b max - 1)` ? `log-b` is translated into a `bsr` asm instruction by the compiler directly

runtime/random.reds
```diff
@@ -98,7 +98,37 @@ _random: context [
...
+		until [
+			rnd: either secure? [rand-secure] [rand]
+			rnd: rnd and mask
```
hiiamboris: I think this will be a lot slower than my proposed remainder solution, because think of small numbers, e.g let `size = 8`, then `mask = 15` and you're in average making 2 random rolls per result.

system/runtime/libc.reds
hiiamboris: I think it should be a part of a new PR instead. And just rename the `log-2` entry to `log-e` in this and other places, rather than importing `log2` which we have as much faster `log-b`

hiiamboris: By the way, `log-2` is used in R/S in only 2 places - in `log-2*` and `log-e*` natives, so not hard to replace

9214: I don't see any `log-b` in `msvcrt` exports either.

greggirwin:
> > Perhaps a limit on retries would be good? Let's say at most 1000 and then error?
> 
> I don't see a reason for that

Limits on retries are almost always a good idea. cases where something can be formally proven to never exceed N tries are an exception of course, but if there is no guarantee, it's a spinlock waiting to happen.

greggirwin:
And if it *is* formally proven, that needs to be clearly doc'd, because someone will come along later, change code, not add a retry check, and can claim ignorance about why they destroyed the world.

hiiamboris:
Well, to me it's dead code since none of us humans will live to see that error fire ;)

system/runtime/libc.reds
loziniak: @9214 I'm confused, because you posted link to entry about `log` and `log10`, and here is just similar [entry about log2](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/log2-log2f-log2l?view=vs-2019). I'd expect that log2 is similarly available. Unless I'm missing something, which is very likely, because I don't program in C, especially for Windows.

loziniak: I thought I've [dealt with natives](https://github.com/red/red/pull/4309/files#diff-6733eb7cea1c8116dace22a0b0b93542R1801) already. Is there something more left?

hiiamboris: `log2` isn't exported by the default msvcrt.dll. It's only exported since MSVC2013 (msvcr120.dll)

loziniak:
It can happen in gazillion of years, but it may happen tomorrow ;-) Probability, b**ch ;-)

runtime/random.reds
loziniak: Probably it could, but I used your algorithm anyway.

hiiamboris:
Indeed :) But so may a meteor hit the Earth (and this is so much likely than even a 100 randoms miss in a row) and we won't have the extra random call problem. Or your PC may just hang. An older RAM module produces in average 10 errors every day ([source](https://www.zdnet.com/article/dram-error-rates-nightmare-on-dimm-street/), I suspect it's much more on modern hardware. Happen in a critical space and your program goes awry.

Besides, how an error is better than an extra random call? Just toss it 1001st time in case you hit a jackpot with 1000 tosses.

Let's not lose touch with reality, guys ;)

system/runtime/libc.reds
9214: @loziniak don't expect — test, otherwise your PRs will have a hard time being accepted, regardless of the project you contribute to.

loziniak:
And what you think about alreadt mentioned topics: tests and seedability?

hiiamboris:
@loziniak 
1) Yes, please add some tests. There are even tests with hardcoded seed & output in the test suite: https://github.com/red/red/blob/87d8f523b212cb059114c12d8ab9a49884a671a0/tests/source/units/date-test.red#L822 , but testing for non-repetition is also OK
2) Seeding CSPRNG doesn't make sense I think. Seeding is predictability, and the whole point of CSPRNG is to not let the attacker predict it's output. Besides, it should use user input, network activity and other available entropy sources, so should not be seedable at all. I would go as far as to throw an error when both `/seed` and `/secure` refinements are used to not give a false impression that it's a valid thing to do.

dander:
Really interesting discussion here. I'd just point out the possibility of running on some hardware that doesn't generate the best random numbers would change the chances of this happening (I guess the chance of this is low, but I expect people would blame Red rather than their stupid hardware). Potentially even could cause the spinlock in a worst-case. But I'm sorry I don't have any actual data about it.

hiiamboris:
Are you really suggesting that there such a hardware RNG that only returns numbers in the range 4000'0000h to 7FFF'FFFFh? ☻

dander:
I don't know how it would happen, but I've heard of some pretty stupid things 😉 
Maybe it is too remote to be worth considering.

loziniak:
@dander in our case (/dev/urandom) when there is not enough entropy from hardware, OS-level cryptographically secure pseudo random generators are used. It's hard for me to imagine platform that would generate so little entropy not to be able to seed CSPRNG from time to time.

loziniak:
I don't get what's wrong with the tests...
```
% wine ./rebview.exe -qws quick-test/run-test.r tests/source/units/integer-test.red 
0009:err:module:load_so_dll failed to load .so lib "/usr/bin/../lib32/wine/l3codeca.acm.so": libmpg123.so.0: cannot open shared object file: No such file or directory
Quick-Test v0.12.0
Running under REBOL 2.7.6.3.1
002d:err:module:load_so_dll failed to load .so lib "/usr/bin/../lib32/wine/l3codeca.acm.so": libmpg123.so.0: cannot open shared object file: No such file or directory
0032:fixme:dwmapi:DwmIsCompositionEnabled 0032FE78
~~~started test~~~ integer
~~~finished test~~~  integer
  Number of Tests Performed:      563
  Number of Assertions Performed: 1825
  Number of Assertions Passed:    1825


% wine ./rebview.exe -qws quick-test/run-test.r tests/source/units/series-test.red                                   :(
0009:err:module:load_so_dll failed to load .so lib "/usr/bin/../lib32/wine/l3codeca.acm.so": libmpg123.so.0: cannot open shared object file: No such file or directory
Quick-Test v0.12.0
Running under REBOL 2.7.6.3.1
002d:err:module:load_so_dll failed to load .so lib "/usr/bin/../lib32/wine/l3codeca.acm.so": libmpg123.so.0: cannot open shared object file: No such file or directory
0032:fixme:dwmapi:DwmIsCompositionEnabled 0032FE78
~~~started test~~~ series
~~~finished test~~~  series
  Number of Tests Performed:      441
  Number of Assertions Performed: 654
  Number of Assertions Passed:    654


```

9214:
@loziniak 4 tests from `run-all-interp.red` are failing. Try to run `run-all.r` in the project's root folder and check the log file in `/quick-test`.

hiiamboris:
@loziniak Doesn't look like GH detected new commits. It didn't re-run the tests. Did you replace/remove some already pushed commits on your local branch?

loziniak:
I amended last commit with fixes and force-pushed.

hiiamboris:
You may have to squash the commits before making a new PR then.

9214:
@loziniak force push with care and only in dire need, because usually it messes up thing pretty badly if commits were already pushed to the remote.

loziniak:
Tests finally pass. BTW is it possible to check contents of `quick-test/quick-test.log` in Travis/AppVeyor failed builds? I couldn't find it.

hiiamboris:
Travis, when fails, outputs it in the "Raw log" section.

qtxie:
Thank! Great team work! 👍

PR #4320, opened 2020-03-07
 FIX: issue #4318 ([Parse] SET and COPY work even without a sub-rule)

### LINKED ISSUE #4318: [Parse] SET and COPY work even without a sub-rule
**Describe the bug**

`set` and `copy` Parse keywords usually extract a matched portion of the input from a given sub-rule into a given word. However, usage of these keywords without sub-rules is still permitted.

**To reproduce**

```red
>> parse [][copy word] word
== []
>> parse [][set word] word
== none
```

Rule by itself always succeeds:

```red
>> parse [1 2 3][[copy word] 3 integer!]
== true
>> parse [a b c][[set word] 3 word!]
== true
```

**Expected behavior**

```red
*** Script Error: PARSE - unexpected end of rule after: set / copy
*** Where: parse
*** Stack:  
```

**Platform version**

Latest.

9214:
Fixes #4318 by peeking a bit farther ahead during keyword processing and checking the presence of a sub-rule.

### DIFF
--- a/runtime/parse.reds
+++ b/runtime/parse.reds
@@ -1514,7 +1514,7 @@ parser: context [
 						]
 						sym = words/copy [				;-- COPY
 							cmd: cmd + 1
-							if any [cmd = tail TYPE_OF(cmd) <> TYPE_WORD][
+							if any [cmd + 1 >= tail TYPE_OF(cmd) <> TYPE_WORD][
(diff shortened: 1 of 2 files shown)


dockimbel:
That's the right fix!

9214:
@dockimbel [god daym!](https://www.youtube.com/watch?v=DwzZ-yMm6E0).

PR #4332, opened 2020-03-12
FIX: issue #2532 (Math doesn't follow the arithmetic precedence for %, **)

### LINKED ISSUE #2532: Math doesn't follow the arithmetic precedence for %, **
```
>> math [10 + 10 ** 2]
== 400
>> math [10 + 10 % 2]
== 0
```
Besides, shall the math dialect be implemented using expensive copy/deep and replace operations? It makes the dialect almost useless.

9214:
This complete rewrite of `math` dialect fixes #2532 and, hopefully, addresses the modest goal of evaluating given expression using PEMDAS precedence rules.

```red
>> math [1 + 2 * 3]
== 7
>> math [(1 + 2) * 3]
== 9
>> math [10 + 10 ** 2 ** 3]
== 100000010
>> math [10 + 10 % 2]
== 10
>> math [8 / 5 % 2]
== 1
```

Covering all edge-cases (like e.g. what `quote * 1` should do, or what if `/` isn't really bound to division `op!`, what if there's more than one right-associative operator of the same precedence level) is beyond the scope of this endeavor. `math`, as I see it, is intended to be a [stele](https://en.wikipedia.org/wiki/Stele) that engraves tricks of Parse trade and shows how one can make a dialect.

At least [Picosheet demo](https://github.com/red/code/blob/master/Showcase/picosheet.red) will need an update after this PR is merged.

### DIFF
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -275,22 +275,21 @@ replace: function [
 ]
 
 math: function [
-	"Evaluates a block using math precedence rules, returning the last result"
-	body [block!] "Block to evaluate"
-	/safe		  "Returns NONE on error"
+	"Evaluates expression using math precedence rules"
+	datum [block! paren!] "Expression to evaluate"
+	/local match
 ][
-	parse body: copy/deep body rule: [
-		any [
-			pos: ['* (op: 'multiply) | quote / (op: 'divide)] 
-			[ahead sub: paren! (sub/1: math as block! sub/1) | skip] (
-				end: skip pos: back pos 3
-				pos: change/only/part pos as paren! copy/part pos end end
-			) :pos
-			| into rule
-			| skip
-		]
+	order: ['** ['* | '/ | '% | '//]]
+	infix: [skip operator [enter | skip]]
+	
+	tally: [any [enter [fail] | recur [fail] | count [fail] | skip]]
+	enter: [ahead paren! into tally]
+	recur: [if (operator = '**) skip operator tally]
+	count: [while ahead change only copy match infix (do match)]
+
+	do also datum: copy/deep datum foreach operator order [
+		parse datum tally
 	]
-	either safe [attempt body][do body]
 ]
 
 charset: func [
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -279,7 +279,7 @@ math: function [
 	datum [block! paren!] "Expression to evaluate"
 	/local match
 ][
-	order: ['** ['* | '/ | '% | '//]]
+	order: ['** ['* | quote / | quote % | quote //]]	;@@ compiler's lexer chokes on '/, '% and '//
 	infix: [skip operator [enter | skip]]
 	
 	tally: [any [enter [fail] | recur [fail] | count [fail] | skip]]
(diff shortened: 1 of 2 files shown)


hiiamboris:
You win a logic obfuscation award from me ;)
It seems to work if I remove `recur` completely btw.

environment/functions.red
```diff
@@ -275,22 +275,21 @@ replace: function [
 ]
 
 math: function [
-	"Evaluates a block using math precedence rules, returning the last result"
-	body [block!] "Block to evaluate"
-	/safe		  "Returns NONE on error"
+	"Evaluates expression using math precedence rules"
+	datum [block! paren!] "Expression to evaluate"
```
greggirwin: Why `datum` rather than `body` or `input`?

greggirwin:
Certainly the most important question is what the goal and use cases for `math` are, and if `math` is even the best name. `Calc` might not be bad, thinking of spreadsheets being a primary use case. Modulo and integer divide are far less important there, though we still need exponentiation.

I don't see it as a stele (thanks for the link on that), or as a tricks of the trade example, but more on the pragmatic side. Utility first.

Keeping it simple, as a gentle introduction to `parse` is a great goal, but is at odds with deep correctness of expression evaluation in all possible cases. Here I lean toward limiting its application in order to keep it simple. Even then, in the context of spreadsheets and future ideas there, we can't ignore the mixing of "real" Red code or use by programmers rather than users putting formulae in cells.

environment/functions.red
9214: Because it lines-up with other 5-letter words I used; helps me to "see things" by inferring the Giger-esque geometry of the code 👽

`<word>: [...]` is two ribs, vertebras of `: [` form a vertical sternum, ribcage of grammar definitions hide the heart between 2nd and 5th intercostal spaces, the heart itself (`tally`) has a rhythmic, pulsating nature (`[... [fail] | ... [fail] | ...]`), the surface tension of `do also ...` (it literally pulls `datum` in two different directions) marks a skin tissue that delineates `math` from the rest of the world.

"Datum" also means "something given", passive data to operate upon, a primordial soup.

9214:
I have a more involved version that is more in line with my initial design idea:

```red
math: function [
	"Evaluates expression using math precedence rules"
	datum [block! paren!] "Expression to evaluate"
	/local operator match
][
	order: ['** [* / % //] [+ -]]
	
	infix: [skip set operator word! (do gauge) [enter | skip]]
	redex: [infix if (redex?)]
	right: [infix if (right?)]
	
	gauge: [right?: attempt [lit-word? first redex?: find to block! group operator]]
	emend: [parse datum tally]
	
	tally: [any [enter [fail] | recur [fail] | count [fail] | skip]]
	enter: [ahead paren! into tally]
	recur: [ahead right 2 skip tally]
	count: [while ahead change only copy match redex (do match)]
	
	do also datum: copy/deep datum foreach group order emend
]
```

You can now play with `order` by changing the precedence of operators, extending it, with e.g. bitwise operations, or ticking existing operators to flip their associativity.

hiiamboris:
Just to illustrate where obfuscation leads:
```
>> math [1 + 2 ** (3 * 4)]
== 531441
>> 1 + (2 ** (3 * 4))
== 4097
```

9214:
@hiiamboris I cannot reproduce that with the version that I submitted in a PR.

hiiamboris:
Yes. That one worked. It's the involved one ;)

9214:
Both of the involved ones work fine also; there was an oversight in the first one where the matched `word!` wasn't an operator but the rule succeeded anyway.

hiiamboris:
Yes, updated version's fine :+1:

PR #4340, opened 2020-03-15
FEAT: support image! to Arbitrary quadrilateral

bitbegin:
like:
```
view [base 600x600 draw [image img 100x100 300x80 50x350 400x400]]
view [base 600x600 draw [image img 100x100 300x80 50x350]]
```


dockimbel:
@qtxie If you are OK with this PR, let's merge it, as it's a very valuable feature to have.

PR #4343, opened 2020-03-16
FEAT: MONEY! datatype

9214:
This is an implementation of `money!` datatype based on 17.5 BCD numbers, i.e. it provides 22 digits of precision and 5 digits of scale, allowing one to do some macro-economic calculations (trillions) and beyond.
```red
>> $1234.56789 * 1337
== $1650617.26893
>> make money! [CCC -123 456]
== -$123.00456
```

The primary goal is accuracy, not efficiency (and anyway if you need the latter then you need to sacrifice the former and use `float!`).

Things that are finished: all scalar-related natives and actions (i.e. arithmetic, comparison, conversion). Things that are missing: `round` and `random` actions, support for currency codes, compiler support.

A dedicated test suite is written, but not included in the QuickTest framework. One can run it from the interpreter with:
```red
do %./red/tests/source/units/money-test.red
```


runtime/datatypes/money.reds
```diff
@@ -0,0 +1,1563 @@
...
+	flip-sign: func [
+		money   [red-money!]
+		return: [red-money!]
+	][
+		#if debug? = yes [if verbose > 0 [print-line "money/flip-sign"]]
+		set-sign money as integer! not as logic! get-sign money
+	]
+	
+	collate-signs: func [
+		this    [integer!]
```
greggirwin: Is there a precedent for `this/that` naming? I think we normally use `value1/value2` or `a/b`.

runtime/datatypes/money.reds
```diff
@@ -0,0 +1,1563 @@
...
+			amount: amount + 1
+		]
+		
+		either zero? count [1][count]				;-- 0 is a single digit
+	]
+	
+	;-- Slices --
+	
+	compare-slices: func [
+		this-buffer [byte-ptr!] this-high? [logic!] this-count [integer!]
```
greggirwin: OK, I see that `this/that` is a recurring theme. I'm just skimming to review, but they are visually similar, which requires more careful reading even at that.

9214: @greggirwin on a personal note: I don't have trouble with writing/pronouncement because I never really read code at the visual level, i.e. as "words" with "spellings". For me `this` and `that` are more like ← and → movements (see [kinaesthesia](https://en.wikipedia.org/wiki/Proprioception) and [tactile learning](https://en.wikipedia.org/wiki/Kinesthetic_learning)), as if I'm raising my (left) hand to point at something or (right) hand to point at something else, or swaying my body; `buffer` is like a sweeping movement, as if I'm spreading a deck of cards (bytes) on the table; `high?` is up/down, or maybe a raised index finger; `count` is bouncing N number of times, a vibration with a specific frequency. In the code `this-*` and `that-*` go hand in hand and, most of the time, are symmetrical in their usage.

On a more formal note (related to your other comment): technically these are not Red values; it would make sense to call arguments like that (pun intended) e.g. in actions or natives, as it is the case with e.g. `add`. But I agree `this-sign` and `that-sign` make more sense and adhere to naming guidelines (nouns for data). I'll do some coding-style cleanup later.

greggirwin: Thanks for explaining how you see things. Very interesting! Reminds me of this guy: https://www.youtube.com/watch?v=PPySn3slfXI

I don't see pictures in my head, only catching a glimpse if I try really, really hard, and have a very poor sense of direction. For me, this work is all about language, which is why Red clicks for me, and many other languages don't I think. So interesting how our minds work.

PR #4346, opened 2020-03-17
 FIX: issue #4272 (TO-BINARY with ANY-LIST argument behaves incorrectly except for BLOCKs)

### LINKED ISSUE #4272: TO-BINARY with ANY-LIST argument behaves incorrectly except for BLOCKs
**Describe the bug**
`to-binary` is (amongst others) defined for `any-list!` arguments; in actual fact, applying it to a `paren!` or `hash!` value gives a stack overflow error

**To reproduce**
```
>> to-binary [1 2 3]
== #{010203}
>> to-binary make hash! [1 2 3]
*** Internal Error: stack overflow
*** Where: to
*** Stack: to-binary  
>> to-binary quote (1 2 3)
*** Internal Error: stack overflow
*** Where: to
*** Stack: to-binary  
```

**Expected behavior**
Should treat hashes and parens as the equivalent block.

9214:
Fixes #4272. Conversion from `any-list!` to `binary!` internally relied on `insert` action, which for `hash!` and `paren!` cases inadvertently overflowed the stack by converting (and pushing) the same value slot over and over again. So the issue was a byproduct of a more serious problem:

```red
>> insert #{} quote (1 2 3)
*** Internal Error: stack overflow
*** Where: insert
*** Stack:  
```

It also renames `ANY_LIST` macro so that it is consistent with other macro definitions in the same group. A couple of tests are provided.

### DIFF
--- a/runtime/datatypes/binary.reds
+++ b/runtime/datatypes/binary.reds
@@ -1158,7 +1158,7 @@ binary: context [
 		]
 		
 		type: TYPE_OF(value)
-		either ANY_LIST(type) [
+		either ANY_LIST?(type) [
 			src: as red-block! value
 			s2: GET_BUFFER(src)
 			cell:  s2/offset + src/head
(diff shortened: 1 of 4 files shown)


greggirwin:
Having explanations to go with fixes is *great* @9214. Thanks.

PR #4347, opened 2020-03-17
FIX: issue #3595 (`return:` for func now chokes if it has a doc string (interpreter))

### LINKED ISSUE #3595: `return:` for func now chokes if it has a doc string (interpreter)
**Describe the bug**
```
>> fn: func [return: [integer!]][1]
== func [return: [integer!]][1]
>> fn: func [return: [integer!] "Test"][1]
*** Script Error: invalid function definition: return:
*** Where: func
*** Stack: fn  
```

**To Reproduce**
Include a doc string for a `return:` spec in a function.

This is used in the `arg-info` function in help, been there a very long time, and Red still builds regularly, so seems to be interpreter only.

**Expected behavior**
Should be a valid func decl.

Last GUI console I have that worked with this syntax was 09-Jul-2018.

```
Red 0.6.3 for Windows built 16-Nov-2018/15:15:54-07:00 commit #8b3128e
```

9214:
Resolves #3595. Docstring after `return:` spec is now allowed, but only if it is located at the tail of the function spec (see newly added tests for examples).

### DIFF
--- a/runtime/datatypes/function.reds
+++ b/runtime/datatypes/function.reds
@@ -772,7 +772,7 @@ _function: context [
 						value: value + 1
 					]
 				]
-				TYPE_SET_WORD [							;-- only return: is allowed as a set-word!
+				TYPE_SET_WORD [								 ;-- only return: is allowed as a set-word!
 					w: as red-word! value
 					if words/return* <> symbol/resolve w/symbol [
 						fire [TO_ERROR(script bad-func-def)	w]
@@ -781,10 +781,14 @@ _function: context [
 					next2: next + 1
 					unless all [
 						next < end
-						TYPE_OF(next) = TYPE_BLOCK		;-- return: must have a type spec
-						any [							;-- This allows a return: spec before each refinement
-							next2 = end
-							TYPE_OF(next2) = TYPE_REFINEMENT
+						TYPE_OF(next) = TYPE_BLOCK			 ;-- return: must have a type spec
+						any [
+							next2 = end						 ;-- return: with type spec is enough
+							TYPE_OF(next2) = TYPE_REFINEMENT ;-- This allows a return: spec before each refinement
+							all [							 ;-- docstring is allowed if at the tail
+								TYPE_OF(next2) = TYPE_STRING
+								next2 + 1 = end
+							]
 						]
 					][
 						fire [TO_ERROR(script bad-func-def) value]
@@ -804,7 +808,7 @@ _function: context [
 		]
 		check-duplicates spec
 	]
-
+	
 	count-locals: func [
 		node	[node!]
 		offset	[integer!]
--- a/tests/source/units/function-test.red
+++ b/tests/source/units/function-test.red
@@ -812,7 +812,9 @@ if system/state/interpreted? [                      ;-- not yet supported by com
 	--test-- "fsv16"
         --assert error? try [do [func [a return: b][]]]
 
-
+	--test-- "#3595"
+		--assert function? func [return: [block!] "string"][]
+		--assert error? try [do [func [return: [block!] "string" word][]]]
 ===end-group===
 
 ~~~end-file~~~


meijeru:
This improvement makes it impossible to write:
```
f: func [... return: [integer!] "some comment" /local a b][...]
```
That seems a rather arbitrary restriction. What is the rationale?

9214:
@meijeru you tell me, it follows your [specification](https://github.com/meijeru/red.specs-public/blob/master/specs.adoc#function) to the letter. 🤷‍♂️ 

I propose:
```
<function-spec> ::= [<docstring>° <argument-spec> <return-spec>° <optional-arguments>*]
```

Perhaps with an additional restriction that `<function-refinement>` is either `/local` or `/extern`. If that sounds reasonable, I'll update the spec verifier.

qtxie:
@meijeru This restriction is not caused by this PR. Before this PR, you cannot even write this:
```
f: func [... return: [integer!] "some comment"][...]
```

hiiamboris:
@9214 I would prefer if we could either freely mix refinements and `return:`, or if `return:` always came before the first refinement. People will anyway try to make it most readable, so I don't see a point in extra constraints (which are also extra CPU cycles). But the main thing is *to have func specs **easily** extensible* programmatically, e.g. `compose [(old-spec) /my-ref my-arg]` should not implement a full-featured spec parser to determine how to properly rebuild it.

meijeru:
@9214 "My" specification is not _prescriptive_ but tries to be _descriptive_. Writing it has brought to light many opportunities for improvement and/or removal of inconsistencies. If the final implementation is as free as possible, as would be welcomed by some of us, I will be happy to document it in the spec.

greggirwin:
@hiiamboris having func specs easily processed programmatically is nice, but not as important as having a clear spec for humans to follow when writing them. It may be a bit much to expect a simple `compose` to always work.

Maybe we need to revisit this. I was in a mindset of compatibility when I read @9214's suggestion, but does it really make sense to have `return:` come before optional args? Another way to ask is with examples. Which of these make sense for a human to write?

```
fn-1: func [
	return: [ref!] "You had it coming"
	a [word!] "What is this about?"
	/b b-arg [string!] "Note it"
	/c c-arg [file!] "Include this"
	/local d
	/extern e
][...]

fn-2: func [
	a [word!] "What is this about?"
	return: [ref!] "You had it coming"
	/b b-arg [string!] "Note it"
	/c c-arg [file!] "Include this"
	/local d
	/extern e
][...]

fn-3: func [
	a [word!] "What is this about?"
	/b b-arg [string!] "Note it"
	return: [ref!] "You had it coming"
	/c c-arg [file!] "Include this"
	/local d
	/extern e
][...]

fn-4: func [
	a [word!] "What is this about?"
	/b b-arg [string!] "Note it"
	/c c-arg [file!] "Include this"
	return: [ref!] "You had it coming"
	/local d
	/extern e
][...]

fn-5: func [
	a [word!] "What is this about?"
	/b b-arg [string!] "Note it"
	/c c-arg [file!] "Include this"
	/local d
	return: [ref!] "You had it coming"
	/extern e
][...]

fn-6: func [
	a [word!] "What is this about?"
	/b b-arg [string!] "Note it"
	/c c-arg [file!] "Include this"
	/local d
	/extern e
	return: [ref!] "You had it coming"
][...]

```
1: :+1:?  2: :-1: 3: :-1: 4: :+1: 5: :-1: 6: :-1:?

That said, you can put `/local` before other refinements, so we _can_ say that the model is "required args before any optional args, but `/local /extern return:` can appear anywhere. Let's think about this in light of extended func constructors as well.

greggirwin:
Freeform can be great, but think about the natural order (at least in English) for adjectives. 

https://dictionary.cambridge.org/grammar/british-grammar/adjectives-order

hiiamboris:
2 :+1: 4 :+1:

hiiamboris:
> That said, you can put /local before other refinements, so we can say that the model is "required args before any optional args, but /local /extern return: can appear anywhere. Let's think about this in light of extended func constructors as well.

Mandatory arguments come before optional arguments because there's no other way (unless you also want a reserved refinement e.g. `/mandatory` to declare that what follows is not optional).

`/extern` is only a hint to `function` constructor, which will be gone once it processes the spec.
`/local` is a normal refinement like any other. No one ever enforced that `/local` comes after other refinements, yet as you can see people put it there. Discipline comes out of a desire to make the code beautiful for the mind, not from a strict ruleset.

`return:` is an anomaly. It is mandatory (as all functions return something) so it belongs categorically to other mandatory arguments. It is it's set-word appearance that makes it possible to put it anywhere. And since it does not control arity, it is also optional. So I would use it like in your `fn-2`. This is also how R/S uses it. 

If you wish to instill some consistency in writing the spec, the best way to do that IMO is make function constructors reorder the spec, putting return in a place where it looks better (same way it can place `/local` at the end). People will learn visually from what they see. Another tool for that is style guide.

meijeru:
Like hiiamboris, I think fn-2 and fn-4 are the most appropriate, because most "natural" way of showing the function's properties. In order to leave the writers the choice between these two, the rule _could_ be:
mandatory arguments first, then optional arguments  and return spec in any order, with one return spec only. As someone said, people will probably put `/local` at the end in any case. I am not sure where `/extern` should go -- have never used it.

greggirwin:
We may also get func attrs (`[throw catch]`), which could have a keyword as well. In Rebol they have to come first in the spec.

greggirwin:
R/S uses it like v4, doesn't it? There are no optional args there.

greggirwin:
`Return:` is an "output". Intermixing it with other inputs doesn't make sense to me.

hiiamboris:
> R/S uses it like v4, doesn't it? There are no optional args there.

Or like that. v2 and v4 become the same without `/b` and `/c`

greggirwin:
If there can be only one, do we all agree that it should be v4 then, WRT `return:`? 

Next we can tackle `/local /extern` ordering. :^)

hiiamboris:
Unneeded work IMO ;)

greggirwin:
@hiiamboris do you mean enforcing v4 or `/local /extern` ordering?

hiiamboris:
Both I guess.

9214:
Allowing `return:` to be placed before (or in the middle of) argument list will cause a problem with `op!` in its current state. Basically it won't be able to recognize binary functions as such since `set-word!` preemptively terminates the checking.
https://github.com/red/red/blob/b45271621e78d6f24e4c0e4d6f1f0acc6cf61640/runtime/datatypes/op.reds#L58

dockimbel:
The function spec has a an external interface part and an internal one. `return:` marks the end of the external interface specification, therefore it can only be placed as in `fn-4`. `fn-1` seems unnatural to me as you specify first the IN interface then the OUT one. `fn-2` and `fn-3` are a minefield for the Red internal runtime code, and don't make sense for the reader. `fn-5` and `fn-6` are breaking the separation between "external" and "internal" interfaces.

hiiamboris:
> The function spec has a an _external_ interface part and an _internal_ one. `return:` marks the end of the external interface specification, therefore it can only be placed as in `fn-4`.

I find this a bit contradictory though.

If we group IN (arguments, refinements) and OUT (return, extern) parts, `/local` becomes a delimiter between the two, as it's neither IN nor OUT by it's meaning. `/local` however being as normal refinement as any other should not be separated from other refinements, while in `fn-4` `return:` splits refinements into 2 groups. That at best means `return:` can appear anywhere between any refinements. So the order would be: 
`arguments -> refinements, local and return in any order -> extern`

If we group external (arguments, refinements) and internal (locals), then I don't know where extern and return stand. I think both belong to the external part (return on par with arguments, extern by it's name). In this case the order becomes: 
`arguments -> refinements -> extern -> return -> locals` which separates locals from refinements.

I'm writing this if only to show that these both notions seem confusing, and that unordered spec will be easier to explain (more precisely, `arguments -> everything else` order) to the programmer and less headaches to deal with (e.g. when writing spec-modifying code). History of Red and Rebol shows that people write the spec for readability even without strict ordering imposed, and few of those who do not - their code is generally such a mess that no rules can help it.

Issue #2213, opened 2016-09-06
INSERT in PARSE cannot insert word’s value, insert word’s name instead


rebolek:
```
red>> value: "1234"
== "1234"
red>> string: "abcd"
== "abcd"
red>> parse string [insert value]
== false
red>> string
== "valueabcd"
```

Zamlox:
it can insert the value of an evaluated expression:
```
red>> parse string [insert (value)]
== false
red>> string
== "1234abcd"
```

dockimbel:
For the sake of performance, I will add this ability to `insert` and `change`, as it will be orders of magnitude faster than going through the evaluation of a paren expression.

PR #4349, opened 2020-03-18
FIX: issue #4200 ([Parse] CHANGE <position> <expression> does not work properly with WORD! values)

### LINKED ISSUE #4200: [Parse] CHANGE <position> <expression> does not work properly with WORD! values
Related: https://github.com/red/red/issues/4153

**Describe the bug**

In Parse, `change` rule can modify input with arbitrary value, either literal one or that coming from evaluated expression. If literal is a `word!`, its value will be fetched and used for modification. Thus, the only way to insert actual `word!` value is by using `paren!` expression, e.g.:

```red
>> bar: 'baz
== baz
>> parse foo: [][change none bar] foo
== [baz]
>> parse foo: [][change none 'bar] foo
== ['bar]
>> parse foo: [][change none (quote bar)] foo
== [bar]
```

However, this does not work if an input position is used instead of a rule - word is fetched as if it was used literally.

**To reproduce**

```red
>> bar: 'baz
== baz
>> parse foo: [][mark: change mark (quote bar)] foo
== [baz]
```

**Expected behavior**

```red
>> bar: 'baz
== baz
>> parse foo: [][mark: change mark (quote bar)] foo
== [bar]
```

**Platform version**

9214:
Fixes #4200.

The story unfolds as follows:
* `change` keyword can change input starting from a specified position;
* if you give `change` a `paren!` expression, it will change input with the result of that expression;
* if you give `change` a `word!` value, it will change input with the value to which this word is set.

But:
* If `change` changes input starting from a specified position;
* And if `paren!` returned a `word!`, or if `word!` was set to another `word!`;

Then this `word!` will be fetched **again**, and the result will be used to change the input.

The culprit was a single (now deleted) line that did exactly that — fetched word's value if that word was returned from `paren!` or fetched from another `word!`.

Relevant tests are provided.

### DIFF
--- a/runtime/parse.reds
+++ b/runtime/parse.reds
@@ -1719,7 +1719,6 @@ parser: context [
 									copy-cell as red-value! input base 	;@@ remove once OPTION? fixed
 									input/head: new/head
 									PARSE_SAVE_SERIES
-									if TYPE_OF(value) = TYPE_WORD [value: _context/get as red-word! value]
 									actions/change input value base as-logic max null
 									if s-top <> null [stack/top: s-top]
 									PARSE_RESTORE_SERIES
--- a/tests/source/units/parse-test.red
+++ b/tests/source/units/parse-test.red
@@ -2746,15 +2746,19 @@ Red [
 		x4200-mark:  x4200-block
 		
 		parse x4200-block [x4200-mark: change x4200-mark x4200-word]
-		--assert x4200-block = [foo]
+		--assert x4200-block = [foo]									;-- word's value
 		clear x4200-block
 		
 		parse x4200-block [x4200-mark: change x4200-mark (x4200-word)]
-		--assert x4200-block = [foo]
+		--assert x4200-block = [foo]									;-- result of expression
 		clear x4200-block
 		
 		parse x4200-block [x4200-mark: change x4200-mark ('foo)]
-		--assert x4200-block = [foo]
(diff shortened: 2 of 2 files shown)


greggirwin:
I love a good story. :^)

dockimbel:
@9214 It was done on purpose, see https://github.com/red/red/issues/2213. You should use `git blame` more often to search the commit history for getting more info about a piece of code.

I think returned word should be evaluated as it is now. If the user wants to insert a word, he needs to return a lit-word then.

9214:
@dockimbel this PR was closed prematurely.

```red
>> set [foo bar][bar qux]
== [bar qux]
>> parse block: [][change none foo] block
== [bar]
>> parse block: [][change none (foo)] block
== [bar]
```
:point_up: this is a feature that you've mentioned.

```red
>> set [foo bar][bar qux]
== [bar qux]
>> parse block: [][mark: change mark foo] block
== [qux]
>> parse block: [][mark: change mark (foo)] block
== [qux]
```
:point_up: and this is a bug that this PR fixed.

Instead of evaluating `foo` and inserting `bar` (as proposed in #2213 and as can be seen in the first example), it evaluates `foo`, takes hold of `bar`, _and then_ evaluates `bar` again and inserts `qux` to which it refers (as documented in #4200 and as can be seen in the second example). 

This is inconsistent both with how `paren!` behave in Parse and Red evaluation semantics.

Issue #4190, opened 2019-12-18
Crash in `fire-on-set` on facet update


loziniak:
([original gitter post](https://gitter.im/red/bugs?at=5df9eab18dfce538b5c8aca1))
After release-compiling this code and clicking a button the app crashes:
```
Red [Needs: 'View]
a: make face! [
    type: 'base color: red size: 20x20
    showw: function [
        parent [object!]
        offs [pair!]
    ] [
        self/offset: offs
        self/parent: parent
        append parent/pane self
    ]
]

view [
    panel 500x500 [
        button "show a" [a/showw face/parent 30x30]
    ]
]
```

Stack trace:
```
*** Runtime Error 98: assertion failed
*** in file: XXXred/gtk/red-github/runtime/datatypes/object.reds
*** at line: 384
***
***   stack: red/object/fire-on-set 00000062h 080951FEh 00000000h FFE69840h
***   stack: red/object/fire-on-set F54FBFB0h 091DCF94h 091E43E4h 091E43D4h
***   stack: red/object/fire-on-set* 091DDF24h 091DCF94h
***   stack: ctx||458~showw F7EE3C8Ch
***   stack: red/_function/call 091E43A4h F7EE3C8Ch
***   stack: red/interpreter/eval-code 091E43A4h F531C788h F531C788h false F531C758h F531C7B0h F54FA000h
***   stack: red/interpreter/eval-path F531C7B0h F531C768h F531C788h false false false false
***   stack: red/interpreter/eval-expression F531C768h F531C788h false false false
***   stack: red/interpreter/eval F531C6F0h true
***   stack: red/interpreter/eval-function 091E4374h F531C6F0h
***   stack: red/_function/call 091E4374h F7EE4428h
***   stack: red/interpreter/eval-code 091E4374h F5284A38h F5284A38h false 00000000h 00000000h 091E42D4h
***   stack: red/interpreter/eval-expression F5284A18h F5284A38h false false false
***   stack: red/interpreter/eval 091E4364h true
***   stack: red/natives/do* false -1 -1 -1
***   stack: red/interpreter/eval-arguments F54F94C0h F52849F0h F52849F0h 00000000h 00000000h
***   stack: red/interpreter/eval-code F54F94C0h F52849E0h F52849F0h false 00000000h 00000000h F54F94C0h
***   stack: red/interpreter/eval-expression F52849E0h F52849F0h false false false
***   stack: red/interpreter/eval 091E4344h true
***   stack: red/natives/try* true 0
[...]

PR #4351, opened 2020-03-19
[view] handle firing on-set functions properly when using self inside objec's method

loziniak:
With tests.

Make compiler generate `object/loc-ctx-fire-on-set* ctx||459 exec/~parent` instead of `object/fire-on-set* ~self exec/~parent`

### DIFF
--- a/compiler.r
+++ b/compiler.r
@@ -3370,14 +3370,24 @@ red: context [
 					object/fire-on-set*
 				] to logic! local-word? first back back tail path
 				
-				parent: either 2 < length? path [		;-- extract word from parent context
-					breaks: [-12 -9 -6 -1]
-					set [obj fpath] object-access? copy/part path (length? path) - 1
-					ctx: second obj: find objects obj
-					['word/from ctx get-word-index/with pick tail path -2 ctx]
-				][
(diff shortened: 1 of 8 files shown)


loziniak:
Output from tests on Travis CI:
```
~~~started test~~~ Red Compiler Regression tests part 5

Supplied source - compiler error


-=== Red Compiler 0.6.4 ===- 

Compiling /red/quick-test/runnable/qt-test-comp20063756735.red ...
...using libRedRT built on 19-Mar-2020/20:07:08
*** Compilation Error: undefined word face! 
*** in file: /red/quick-test/runnable/qt-test-comp20063756735.red
*** near: [face! [
        fn: does [self/parent: 'boom]
    ] 
    fc/fn 
    print fc/parent
]
```
Does this mean, that testing View on Linux is not possible in current state of CI?

loziniak:
Any insights?
Fixes #4190

9214:
> Does this mean, that testing View on Linux is not possible in current state of CI?

@loziniak likely yes (see https://github.com/red/red/pull/4348#issuecomment-600634678), but it probably have changed with the introduction of GTK backend; try to sync your branch with the upstream, that will trigger CI re-run.

PR #4357, opened 2020-03-21
FEAT: grant wish #1956 (complement should be allowed on tuples)

### LINKED ISSUE #1956: WISH: complement should be allowed on tuples
This wish is in line with several other wishes, which arise from a review of the datatype vs action matrix.
See #1262, #1263, #1267.

9214:
Fulfills and closes #1956. `complement` on `tuple!` values is now allowed and matches Rebol behavior:
```red
>> complement 254.252.252.248
== 1.3.3.7
```

I didn't find any `tuple!`-specific test suite and don't know where else to put them (if that's needed at all).

### DIFF
--- a/environment/actions.red
+++ b/environment/actions.red
@@ -197,8 +197,8 @@ and~: make action! [[
 
 complement: make action! [[
 		"Returns the opposite (complementing) value of the input value"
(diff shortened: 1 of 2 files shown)


hiiamboris:
Regression tests file comes to the rescue in such cases ;)

PR #4354, opened 2020-03-20
FIX: issue #4197 ([Parse] COLLECT trashes series buffer)

9214:
Fixes #4197 by forbidding `collect into/after` on the following combinations:

- Into `binary!`;
- From `any-block!` into `any-string!`.

Thus, only the following combinations are accepted:

- Into `any-block!`;
- From `any-string!` into `any-string!`.

New type of error is provided to signal such cases:
```red
>> stash: <stash> parse [series][collect into stash keep skip]
*** Script Error: PARSE - COLLECT INTO/AFTER expects a series! of compatible datatype
*** Where: parse
*** Stack:  
```

`ANY_SERIES_PARSE?` macro is defined as a nuclear option for cases where Parse tries to interact with series that it does not support (e.g. #4194); I changed all occurrences of `ANY_SERIES?` inside `parse.reds` to it (so ideally this PR should be merged after #4194).

This PR also provides relevant tests and fixes the misbehaving #3951 test that breaks the testing framework after the first run, by resetting preprocessor's execution context.

PR #4360, opened 2020-03-22
FEAT: grant wish #3055 (DRAW supports PERCENT! for SCALE)

### LINKED ISSUE #3055: WISH: DRAW supports PERCENT! for SCALE
Currently [documentation](https://doc.red-lang.org/en/draw.html#_scale) states that `scale` supports `number!` values, but it (and [`transform`](https://doc.red-lang.org/en/draw.html#_transform) with `<scale-x>` and `<scale-y>` args) will choke up on `percent!`, however:
```Red
>> number!
== make typeset! [integer! float! percent!]
```

9214:
Fulfills and closes #3055. It does not cover [deprecated API](https://doc.red-lang.org/en/draw.html#_deprecated_api) though, and I don't think it should.

Relevant tests can be provided, but really, `percent!` and `float!` are one and the same, so there shouldn't be any problem with that.

### DIFF
--- a/modules/view/draw.red
+++ b/modules/view/draw.red
@@ -224,6 +224,16 @@ Red/System [
 				any [TYPE_OF(pos) = type1 TYPE_OF(pos) = type2]
 			][cmd: pos]
 		]
+		
+		#define DRAW_FETCH_NUMBER [
+			cmd: cmd + 1
+			if any [
+				cmd >= tail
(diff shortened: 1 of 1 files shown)


9214:
Hmm, failing macOS on Travis looks like a [generic](https://github.com/red/red/pull/4354#issuecomment-601908836) issue.

dockimbel:
@qtxie Please review this PR also.

PR #4362, opened 2020-03-22
FIX: issue #3766 (It is possible to unset a word using SET with BLOCK without /ANY refinement)

### LINKED ISSUE #3766: It is possible to unset a word using SET with BLOCK without /ANY refinement
**Describe the bug**
It is possible to unset a word using `set` with `block`?
```
>> b: 1 set [A B] reduce [4 ()]
== [4 unset]
>> b
*** Script Error: b has no value
*** Where: catch
*** Stack:
```
While it is not possible when using `set` with `word` directly:
```
>> b: 1 set 'b ()
*** Script Error: b needs a value
*** Where: set
*** Stack:  

>> b
== 1
```
It should be possible only when used with `/any` refinement.

```
Red 0.6.4 for Windows built 26-Nov-2018/12:26:10+01:00
```

9214:
Fixes #3766.

Red now follows Rebol3 semantics for `set [<word>*] <value>`: `unset!` values inside `block!`, `map!` or `object!` trigger an error unless `/any` refinement is specified; in case of an error all words retain their values (see provided tests for examples).

`foreach <block> <body>` is an interesting edge-case that mimics `set/any` in its operation. In the future this can be changed by flipping a logical flag.


hiiamboris:
Thanks for the foreach idea ;) I thought - what is the simplest way to set multiple values to chars in a string? Like `set [a b c] "abcdefgh"`. Foreach is the answer!

PR #4363, opened 2020-03-23
FEAT: grant wish #4131 (as string! should work on binary!)

### LINKED ISSUE #4131: WISH: as string! should work on binary!
`to-string! <binary>` uses utf-8 decoding, which is fine for most purposes. However, when one wants to decode a binary value that is encoded using ISO-8859-1 or Windows-1252, and this is quite common in e-mail transmission, one should like to use _direct_ conversion of each byte into an 8-bit character. The mechanism is there: `as` allows `binary! -> string!` in its type-spec, but the implementation refuses this particular case, for no apparent reason.

9214:
Fulfills and closes #4131. `as` now can coerce `any-string!` to `binary!` and vice versa. Use-cases and rationale for that are provided in the relevant ticket.

### DIFF
--- a/environment/natives.red
+++ b/environment/natives.red
@@ -844,8 +844,8 @@ sign?: make native! [[
 
 as: make native! [[
 		"Coerce a series into a compatible datatype without copying it"
-		type	[datatype! block! paren! any-path! any-string!] "The datatype or example value"
-		spec	[block! paren! any-path! any-string!] "The series to coerce"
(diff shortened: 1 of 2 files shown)


9214:
Per @dockimbel's comment, `binary!` to `string!` conversion with `as` is unreliable, because it conflicts with internal string encoding, and will be eventually delegated to a special codec system. I propose to limit it only to `string! -> binary!` case.

9214:
Closing per https://github.com/red/red/issues/4131#issuecomment-705713920.

PR #4371, opened 2020-03-26
FEAT: continued work on MONEY! datatype

9214:
New features, in no specific order:
* Support for currencies;
* Currency-aware operations (arithmetic, comparison);
* `code` and `amount` datatype accessors (both with path notation and `pick`);
* `as-money` and `make money! <currency code>` constructors;
* Formatting options (`mold/all`, `system/options/money-digits`);
* Bugfixes in datatype conversion routines;
* Relevant tests.


environment/system.red
```diff
@@ -158,7 +158,8 @@ system: context [
 				bad-loop-series:	["Loop series changed to invalid value:" :arg1]
 				;bad-decode:		"missing or unsupported encoding marker"
 				;already-used:		["alias word is already in use:" :arg1]
-				;wrong-denom:		[:arg1 "not same denomination as" :arg2]
+				wrong-denom:		[:arg1 "not same denomination as" :arg2]
+				bad-denom:			["invalid denomination:" :arg1]
```
greggirwin: While longer, would using full `*-denomination` be better, since `denom` could also be interpreted as `denominator`?

runtime/datatypes/money.reds
```diff
@@ -67,8 +66,15 @@ money: context [
 		S_END
 	]
 	
+	#define SWAP_ARGUMENTS(this-argument that-argument) [
```
greggirwin: Why not just `a b` or `arg-1 arg-2` for arg names?

runtime/datatypes/money.reds
```diff
@@ -843,58 +999,51 @@ money: context [
 	;-- Comparison --
 	
 	compare-money: func [
-		this    [red-money!]
-		that    [red-money!]
-		return: [integer!]
+		this-money [red-money!]
```
greggirwin: Looks like there is no standard for naming in `compare` funcs for types. Some use `value1/2`, some `arg2/arg2`, and at least pair uses `left/right`. If we don't normalize now, it should go on a list for the future.

runtime/datatypes/money.reds
9214: @greggirwin I don't follow, what makes you think `this` and `that` are the issue? I took your previous suggestion into account and renamed arguments as nouns.

greggirwin:
Good work on all this @9214.

runtime/datatypes/money.reds
greggirwin: I thought I said that `this/that` were the issue, but maybe that wasn't clear. The reason being that they are similar visually and, at least for me, less distinctly meaningful.

9214: I think I already explained that it's not an issue for me, on the contrary. I write code in a way that helps me to maintain it (recall e.g. your convention of putting `==` suffixes/prefixes in Parse rules), as the only person who will likely ever do that maintenance; and I don't *read* [the code](https://www.youtube.com/watch?v=qAYXgbSlSv0) the way you do. I think we can agree that being aware of one's shticks and leveraging them is not a bad idea.

Lexically, it's IMO on the same level of meaningfulness as `arg1` or `a`, albeit more verbose.

greggirwin: In this case, the code is not just for you, but for the team and contributors.

lexer.r
```diff
@@ -690,10 +690,7 @@ lexer: context [
 		reset: does [clear stk]
 		
 		clean-up: does [
-			unless empty? stk [
```
qtxie: This fix is reverted. The unit test failed. https://github.com/red/red/runs/537371435

PR #4381, opened 2020-04-04
FIX: missing copy in split function

endo64:
This PR adds missing `copy` to `split` function to fix below bug:
```
>> a: split "a," ","
== ["a" ""]
>> b: split "b," ","
== ["b" ""]
>> same? a/2 b/2
== true
```

### DIFF
--- a/tests/source/environment/functions-test.red
+++ b/tests/source/environment/functions-test.red
@@ -390,6 +390,7 @@ Red [
 		--assert ["a" "b" "c"] = split "a-b-c" "-"
 		--assert ["a" "c"] = split "a-b-c" "-b-"
 		--assert ["a-b-c"] = split "a-b-c" "x"
+		--assert false = do compose [same? (second split "a," ",") (second split "b," ",")] ;PR #4381 missing copy in split function
 ===end-group===
 
 ===start-group=== "dirize tests"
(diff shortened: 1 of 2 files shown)


hiiamboris:
Please add a regression test too

greggirwin:
Good fix. Please do add tests, then I'll merge ASAP.

tests/source/environment/functions-test.red
```diff
@@ -390,6 +390,7 @@ Red [
 		--assert ["a" "b" "c"] = split "a-b-c" "-"
 		--assert ["a" "c"] = split "a-b-c" "-b-"
 		--assert ["a-b-c"] = split "a-b-c" "x"
+		--assert false = do compose [same? (second split "a," ",") (second split "b," ",")] ;PR #4381 missing copy in split function
```
9214, line 393:
```
block: split "," comma
--assert not block/1 == block/2
```

hiiamboris: not gonna work though

9214: Care to elaborate?

hiiamboris: `==` does not test for copied or not, should be `=?`. But even then, it copies all parts right now except the last one (see the parse expression). So this succeeds in current build as well:
```
>> not block/1 =? block/2
== true
```

PR #4386, opened 2020-04-07
FIX: issue #4383 (to-hex is not doing proper bounds check for its `/size` value).

### LINKED ISSUE #4383: to-hex is not doing proper bounds check for its `/size` value
**Describe the bug**
when `to-hex/size` has `size <= 0`, it should throw an out-of-range error. 
```red
>> to-hex/size 1 0
== #
>> to-hex/size 1 -1
== #0123456789ABCDEF
```
It does truncation to maximum 8 hex digits:
```red
>> to-hex/size 1 8
== #00000001
>> to-hex/size 1 99
== #00000001
```
That is probably ok, but could be documented in doc-string.

9214:
Fixes #4383 by throwing an error on non-positive `size` argument. FYI, Rebol3 does almost the same:
```red
>> to-hex/size 123 1
== #B

>> to-hex/size 123 0
== #???

>> to-hex/size 123 -1
** Script error: invalid argument: -1
** Where: to-hex
** Near: to-hex/size 123 -1
```

### DIFF
--- a/runtime/natives.reds
+++ b/runtime/natives.reds
@@ -1614,13 +1614,13 @@ natives: context [
 		/local
 			arg	  [red-integer!]
 			limit [red-integer!]
-			buf   [red-word!]
 			p	  [c-string!]
 			part  [integer!]
 	][
 		#typecheck [to-hex size]
 		arg: as red-integer! stack/arguments
 		limit: arg + size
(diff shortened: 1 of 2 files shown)


9214:
The [logs](https://travis-ci.org/github/red/red/jobs/672137973#L1406) in failing tests are interesting, I'll take a look at that later.

PR #4404, opened 2020-04-17
FEAT: update of MONEY! datatype.

9214:
See the details in `money!` documentation https://github.com/red/docs/pull/224:
* Arithmetic operations with `percent!` are allowed;
* Stricter rules for `money!` literals and conversion from `float!`, `string!`;
* `=?` on two `money!` values returns a `logic!` value, indicating whether they are identical;
* Sorting is applied lexicographically for different currencies and by amounts for same currencies.


9214:
Preliminary work on `ref!` accidentally crept in. I force-pushed from a relevant branch to overwrite these changes.

Issue #2812, opened 2017-06-14
image: alpha channel not correctly displayed


meijeru:
In tuple representation, the fourth element is 255 - A, that is understood (default 0 means opaque, hence 255).
However, in binary representation, the last byte of `/argb` differs from the byte of `/alpha` see below. 
```
>> i: make image! 1x1
== make image! [1x1 #{FFFFFF}]
>> i/1
== 255.255.255.0
>> i/rgb
== #{FFFFFF}
>> i/argb
== #{FFFFFFFF}
>> i/alpha
== #{00}  ;;; should be #{FF}
```

qtxie:
Yes, you're right. The alpha channel specify the `Opacity` in ARGB color.

meijeru:
I think the tuple should still have 4th element 255 - A, since the default with 3-element tuples is 4th element = 0, and the default for A is 255. Thus 255 - A, as currently for tuples, is OK.

qtxie:
> Yes, you're right. The alpha channel specify the `Opacity` in ARGB color. So
> 
> * `i/1` should give `255.255.255.255`.
> * ` i/argb` should give `#{FFFFFFFF}`.
> * `i/alpha` should give `#{FF}`.

I think we should do this change. I don't see a good reason to make it different in Red than the rest of the world. As @hiiamboris points out, 

> The alpha component specifies the transparency of the color: 0 is fully transparent, and 255 is fully opaque.

It's same definition on all the platforms we support, also the same on Web CSS.

Issue #2684, opened 2017-05-14
inconsistency in terminology RGBA vs ARGB


meijeru:
The function which makes a 4-tuple out of 4 integers is called `as-rgba`, suggesting that transparency is the 4th and last byte of a pixel. The refinement of an image value which produces the binary sequence containing 4 bytes for each pixel is called `/argb` suggesting that the transparency byte is up front.

Which of the two suggests better the reality?

qtxie:
I propose to use ARGB in all places. So we need to do the following change:
1. 4-tuple value means ARGB instead of RGBA.
2. rename function `as-rgba` to `as-argb`.

hiiamboris:
I now believe it's all consistent. Tuple 1.2.3.4 is R=1 G=2 B=3 A=4, i.e. RGBA:
```
>> as-rgba 1 2 3 4
== 1.2.3.4
```
`/argb` is not the order in which colors appear in the tuple, but the [ARGB order of encoding them into a binary](https://en.wikipedia.org/wiki/RGBA_color_model#ARGB_%28word-order%29)

PR #4405, opened 2020-04-18
FIX: issue #2812 (image: alpha channel not correctly displayed)

qtxie:
This PR changes the meaning of the alpha channel in color tuple.
R.G.B.A where A = 255 indicates full opacity and A = 0 full transparency.

Note: This change breaks the compatibility with Rebol.

### DIFF
--- a/runtime/datatypes/image.reds
+++ b/runtime/datatypes/image.reds
@@ -49,7 +49,7 @@ image: context [
 			pixel and 00FF0000h >> 16
 			pixel and FF00h >> 8
 			pixel and FFh
-			255 - (pixel >>> 24)
+			pixel >>> 24
 		]
 	]
 
@@ -209,7 +209,7 @@ image: context [
 			while [i <= sz][
 				pixel: data/i
 				either type = EXTRACT_ALPHA [
-					p/1: as-byte 255 - (pixel >>> 24)
+					p/1: as-byte pixel >>> 24
 					p: p + 1
 				][
 					p/1: as-byte pixel and 00FF0000h >> 16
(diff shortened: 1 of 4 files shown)


hiiamboris:
This is what I'm thinking.
If we make `255.255.255` equal to `255.255.255.255`, that would break tuple arithmetic rules. So all 3-tuples after this PR will become transparent colors (`255.0.0 = 255.0.0.0 = transparent`). Which is not useful.
All opaque colors will have to become 4-tuples. And we'll have to add extra `.255` to all colors we manually define in the code, and all predefined colors will have to be updated as well. Which is boring.

Perhaps this PR will do more harm than good, and it answers the design question: why Rebol has chosen a RGBT convention instead of RGBO.

qtxie:
> So all 3-tuples after this PR will become transparent colors.

Nope. How to interpret the colors is controlled by the applications. In view, `view [base 255.0.0]` is still opaque red color. In image!, when you use 3-tuples, you only set R.G.B channels, nothing to do with alpha channel.

> why Rebol has chosen a RGBT convention instead of RGBO

I think just because the image library which Rebol used use this format.

hiiamboris:
Okay but this problem still remains:
```
>> 255.255.255 = 255.255.255.0
== true
>> 255.255.255 = 255.255.255.255
== false
```
So when testing for a color equality one would have to trim the tuple length accordingly, e.g.
```
>> 255.255.255 = (255.255.255.255 * 1.1.1)
== true
```
Something we should note somewhere then.

greggirwin:
Absolutely something to note in tuple docs, as they ignore trailing zeros when comparing.
```
>> 255.255.255 = 255.255.255.0.0.0.0
== true
```
In the context of colors, we have to think (and I haven't yet) about whether opacity is a component of a "color". It isn't, intuitively to me, but a related value for pixels in images.

hiiamboris:
Related https://github.com/red/red/issues/2684

PR #4411, opened 2020-04-22
FEAT: REF! datatype.

9214:
`ref!` is a very thin wrapper on top of `string!`, so the work on it is practically finished. Compiler support is implemented, a couple of tests are also provided.


compiler.r
```diff
@@ -302,6 +302,7 @@ red: context [
 	float-special?: func [value][value/1 = #"."]
 	tuple-value?:	func [value][value/1 = #"~"]
 	money-value?:	func [value][value/1 = #"$"]
```
greggirwin, line 304: For money, does this hold true for `USD$1` format? I haven't worked on the compiler code, so my questions are from an outsiders view.

tests/source/units/lexer-test.red
```diff
@@ -180,7 +180,31 @@ Red [
 		    %//docs/notes
 		]
 		forall out [--assert file? out/1]
-
+	
+	--test-- "tr-19.1"
```
greggirwin, line 184: Nice.

tests/source/units/lexer-test.red
```diff
@@ -180,7 +180,31 @@ Red [
...
+		out: transcode {
+			@
+			@.
+			@.:
+			@.:;comment
+			@/-&!|*~`\
```
greggirwin, line 190: Does this test all the supported non-alphanum chars? e.g. `@%?$.

tests/source/units/mold-test.red
```diff
@@ -13,117 +13,117 @@ Red [
 
 ===start-group=== "string-basic"
 
-	--test-- "mold-string-baseic-1"
```
greggirwin, line 16: Thanks for taking the time to clean up the typos. 👍

utils/redbin.r
```diff
@@ -216,16 +216,17 @@ context [
...
-		type: select [
-			string! TYPE_STRING
-			file!	TYPE_FILE
-			tag!	TYPE_TAG
-			url!	TYPE_URL
-			ref!	TYPE_REF
-			email!	TYPE_EMAIL
-			binary! TYPE_BINARY
-		] type?/word str
+	emit-string: func [str [any-string!] /root /ref /local type unit header][
```
greggirwin, line 219: Might be good to include a comment to say why ref is a special case here.

compiler.r
9214: @greggirwin values that cannot be directly mapped to R2 datatypes are stored as issues with custom encodings (a character prefix like the ones you see here). For `money`, the currency is encoded either as 3 letters or as `...` for a generic currency.

tests/source/units/lexer-test.red
9214: It tests only supported characters, like the rest of the `transcode` tests in the same category.

greggirwin: Thanks. Again, comments can help.

utils/redbin.r
9214: Given that this is an implementation detail relevant only for 2-3 people in the world that work on R2-based compiler...

Series are stored in Redbin payload and need to be properly encoded — and, at a minimum, that requires a datatype ID; this is what  `emit-string` does for `any-string!` values.

All the values except for `ref!` are passed to it literally, and Rebol2 can dispatch on their type and map them to a low-level datatype ID enumeration. `ref!`, however, is passed as an `issue!` with internal encoding and needs special handling, as there's no `ref!` in Rebol2; otherwise, I had to implement `emit-ref` and copy-paste the body from `emit-string`, since they are nearly identical.

greggirwin: How about just pasting that wonderful explanation in as a comment? :^)

> Given that this is an implementation detail relevant only for 2-3 people in the world that work on R2-based compiler...

It's a balancing act. Don't view comments as the enemy, or even a necessary evil, but development in general as more of a living engineering notebook. Sometimes the work itself is all the explanation needed, but quite often hints and reminders are very helpful when reloading your wetware cache when working on many things, or having been away from something for a while.

dockimbel:
@greggirwin Are we good for merging this?

greggirwin:
Per @hiiamboris' note on Gitter, it seems ref only works in the console right now. That's an important fix. I have the same problem here.

9214:
> ref only works in the console

Previous PR was just an interim sync between @dockimbel and me, so that he can implement proper lexer support for `ref!`; I don't understand the point of reporting "bugs" in it. This pull request finishes the work started, and all the tests in it were passing last time I ran them.

dockimbel:
@greggirwin @hiiamboris did not seem to be using this PR for testing, so his report is irrelevant until this PR is merged.

greggirwin:
"Bugs" is a channel on gitter for reporting and discussing problems. There's no official ticket for this. It can be confusing, as I pulled `fast-lexer` myself thinking I'd be up to date to check ref functionality. 

@dockimbel if I need to pull `ref` branch to test, I can do that, but this is still new, and goes with new lexer, so I'm OK with merging if you are.

Issue #4424, opened 2020-05-03
Incorrect output of to-csv


endo64:
**Describe the bug**
See below

**To reproduce**
```
>> o: context [a: "x x"]
== make object! [
    a: "x x"
]
>> to-csv values-of o
== {"x x"^/}
>> to-csv values-of o
== {"""x x"""^/}
>> to-csv values-of o
== {"""""""x x"""""""^/}
>> b: ["x x"]
== ["x x"]
>> to-csv b
== {"x x"^/}
>> to-csv b
== {"""x x"""^/}
>> to-csv b
== {"""""""x x"""""""^/}
```

**Expected behavior**

**Platform version (please complete the following information)**
```
Red 0.6.4 for Windows built 29-Apr-2020/15:24:35+03:00 commit #db7f36a
```

dockimbel:
@rebolek This one is for you. I'm putting Gregg as the proxy in "Assignees".

greggirwin:
Objects and maps share the same encoder, so affects those too. What we need to do is vet the input, to make sure it's what `to-csv` expects, and also put that in the wiki (and eventually docs and articles).
```
>> mm: load-csv/as-columns s
== #(
    "A" ["a" "1" "11"]
    "B" ["b" "2" "22"]
    "C" ["c" "3" "33"]
)
>> to-csv mm
== {A,B,C^/a,b,c^/1,2,3^/11,22,33^/}
>> to-csv mm
== {A,B,C^/a,b,c^/1,2,3^/11,22,33^/}
>> to-csv mm
== {A,B,C^/a,b,c^/1,2,3^/11,22,33^/}
```
I thought at first it was `/skip` related, that you needed to include it in the call, but it seems like it may be a special case where there's only one string in the block.

rebolek:
Merge request with fix created https://github.com/red/red/pull/4425

PR #4425, opened 2020-05-05
FIX: #4424 is gone now

rebolek:


environment/codecs/csv.red
```diff
@@ -65,7 +65,7 @@ context [
 		/extern quote-char double-quote quotable-chars
 	][
 		quot?: false
-		unless string? value [value: form value]
+		value: either string? value [copy value][form value]
```
endo64, line 68: Could be `form`ed anyway: `value: form value`. Direct `form`ing is only slower 10 milliseconds for one million iteration.

rebolek: True, `form` copies string, so your version works same and is shorter. I change it.

PR #4426, opened 2020-05-05
FEAT: update on REF! datatype.

9214:
@dockimbel I hesitated with [Line 1366](https://github.com/red/red/compare/fast-lexer...9214:ref?expand=1#diff-1887889cb99ae2e2424362bd4bb1d136R1366), because it listed only a selected few of `any-string!` datatypes; either that was intentional or it was a remnant of old code that should have been replaced by `ANY_STRING?` check.


dockimbel:
It was intentional, so should not be changed for now.

PR #4435, opened 2020-05-09
FIX: refactor cascaded delimiters match

bitbegin:


bitbegin:
@dockimbel i refactor the match rules, but don't support `<` tag!, i need a open/close event for tag! datatype.

dockimbel:
@bitbegin Thanks. I need to study that case, I'm not sure we should provide an event for that type.

Issue #4432, opened 2020-05-07
[CRASH] involving Parse, Reduce & GC


hiiamboris:
**To reproduce**

Variant 1 (always crashes):
```
Red []
s: "long long long string"
b: reduce [s s [1]]
print "HERE"
color: parse copy "." [change skip (b)]
print "THERE"
```
`red 1.red`:
```
HERE

*** Runtime Error 1: access violation
*** at: 7517AB60h
```
Or `gui-console 1.red`
```
HERE

*** Runtime Error 1: access violation      
*** Cannot determine source file/line info.
***                                        
```

---
Variant 2 (crashes only with GC on):
```
Red []
s: 1.21319613701372131
b: reduce [s s [1]]
color: parse copy "." [change skip (b)]
print "HERE"
cache: make block! 1'000'000
print "THERE"
```
Run it in debug console `gui-console 2.red`:
```
HERE
root: 5181/6970, runs: 0, mem: 2599840                                                                 
*** Runtime Error 98: assertion failed                                                                 
*** in file: /D/devel/red/red-src/red-master/runtime/allocator.reds                                    
*** at line: 365                                                                                       
***                                                                                                    
***   stack: red/free-node 00000062h                                                                   
***   stack: red/free-node 00000020h                                                                   
***   stack: red/cross-compact-frame 02C40004h 009E4FD0h                                               
***   stack: red/collect-frames 1                                                                      
***   stack: red/collector/do-mark-sweep                                                               
***   stack: red/collector/do-cycle                                                                    
***   stack: red/alloc-series-buffer 1000000 16 0                                                      
***   stack: red/alloc-series 1000000 16 0                                                             
***   stack: red/alloc-cells 1000000                                                                   
***   stack: red/block/make-at 02A59144h 1000000                                                       
***   stack: red/block/make 02A59144h 02A59154h 5                                                      
***   stack: red/actions/make 02A59144h 02A59154h                                                      
***   stack: red/actions/make*                                                                         
***   stack: red/interpreter/eval-arguments 02B404B8h 02D98970h 02D98990h 00000000h 00000000h          
***   stack: red/interpreter/eval-code 02B404B8h 02D98950h 02D98990h true 00000000h 00000000h 02B404B8h
***   stack: red/interpreter/eval-expression 02D98950h 02D98990h false true false                      
***   stack: red/interpreter/eval-expression 02D98940h 02D98990h false false false                     
***   stack: red/interpreter/eval 02A59104h true                                                       
[...]

PR #4097, opened 2019-10-19
FIX: issues #3369, #4088, #4087, #4085, #4083, #4078, #4913

hiiamboris:
Fixes #3369 , fixes #4078 , fixes #4083 , fixes #4085 , fixes #4087 , fixes #4088 , fixes #4100 , fixes #4913

This PR fixes numerous crashes in series-related functions, and includes a total rewrite of `change` and `take`. Lots of tests included.

Series funcs behavior now follows [the one I outlined in red/red](https://gitter.im/red/red?at=5da5b83b870fa33a4df7e0b7) in all edge cases. How it all works is seen from the tests, so I'm not duping it here.

`next`, `skip`, `at` I do not plan to modify until we agree on some sort of series model.
`take/last/part` behavior is unchanged, only fixed, although I'm not happy with it ([reasons](https://gitter.im/red/red?at=5da5861f809de9699f3ec707))

**Status**

This PR is ready for review, though a bit more work is ~required~ desirable:
- `move` func ~has no tests at all, and it's~ very unintuitive, I'd love to rewrite it as well (see #3527 ); no tests for it so far; ~`alter` has no tests either~
- ~I didn't test select/pick/random/min/max yet~
- ~I didn't run tests except `series-test.red` yet~
- ~All tests succeed when run from the console, but some heisenbug still remains, causing `run-all-interp.red` (when compiled only) to stop just after `fun-12` test, executing only 1687 of over 10k tests, ignorantly reporting a false success~

P.S. Added memory protection (see below)

PR #4438, opened 2020-05-11
FIX: issue #4432 ([CRASH] involving Parse, Reduce & GC)

qtxie:
Those fixes are from PR #4097.  @hiiamboris 👍

### DIFF
--- a/runtime/datatypes/string.reds
+++ b/runtime/datatypes/string.reds
@@ -993,7 +993,7 @@ string: context [
 			(as byte-ptr! s1/offset) + (offset << (log-b unit1)) + h1
 		]
 		either all [keep? diff?][
-			assert s1 <> s1
+			assert s1 <> s2
 			p2: (as byte-ptr! s2/offset) + h2
 			limit: p2 + (size2 << (log-b unit2))
 			while [p2 < limit][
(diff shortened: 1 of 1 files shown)


runtime/datatypes/string.reds
```diff
@@ -2741,10 +2751,11 @@ string: context [
 			type: TYPE_OF(cell)
 			either type = TYPE_CHAR [
 				char: as red-char! cell
+				s: GET_BUFFER(str)
 				either part? [				;-- /part will insert extra elements
-					s: insert-char s str/head + added char/value
+					insert-char s str/head + added char/value
 				][
-					s: overwrite-char s str/head + added char/value
+					overwrite-char s str/head + added char/value
```
dockimbel, line 2758: Why removing the `s:` part?

greggirwin:
Saw a comment from @dockimbel about `s:` being removed. My guess is that it's because `s` wasn't used after that, and updating it leads you think it is.

qtxie:
@greggirwin Yes. `s` wasn't used after that and we always fetch `s` before using it. So `s` always points to valid memory.

dockimbel:
@qtxie @greggirwin I removed my comment a few minutes after posting it.

PR #4441, opened 2020-05-14
FEAT: relax format of function's return spec.

9214:
A follow-up to https://github.com/red/red/pull/4347. Now `return: [...]`, with or without a docstring, can be followed by a refinement; after that it's free-form.


dockimbel:
Per my comment there https://github.com/red/red/pull/4347#issuecomment-697303029, I won't accept this change.

PR #4450, opened 2020-05-18
support matrix-order for macOS

bitbegin:


dockimbel:
We have decided to drop the `matrix-order` command, see PR https://github.com/red/red/pull/4596

So I'm closing this PR.

PR #4149, opened 2019-11-23
FIX: issue #3624

hiiamboris:
Fixes #3624

PR #4452, opened 2020-05-18
FIX: issues #4448 and #4451.

### LINKED ISSUE #3624: Difference in how compiler and interpreter evaluate replace/case/all quote :a/b/A/a/B [a] 'x
**Describe the bug**
Difference in how compiler and interpreter evaluate replace/case/all quote :a/b/A/a/B [a] 'x
```text
replace/case/all quote :a/b/A/a/B [a] 'x - compiled
::a/b/A/x/B

replace/case/all quote :a/b/A/a/B [a] 'x - interpreted
:x/b/A/x/B
```

**To Reproduce**
Run this code:

```
Red[]

print "replace/case/all quote :a/b/A/a/B [a] 'x - compiled"
probe replace/case/all quote :a/b/A/a/B [a] 'x
print ""
print "replace/case/all quote :a/b/A/a/B [a] 'x - interpreted"
probe do [ replace/case/all quote :a/b/A/a/B [a] 'x ]
``` 

**Expected behavior**

commit 1e0d5ec2360b336ef8111546e3221024dc28e31f (HEAD -> master, upstream/master)

### LINKED ISSUE #4448: GET does not support get-paths/set-paths
:point_up: [May 16, 2020 9:36 PM](https://gitter.im/red/bugs?at=5ec032c890f6db31beee53b5)

**Describe the bug**

Get should work not only for paths but also for get-/set-/lit-paths (as otherwise one always has to do extra checks and coerce the path)
```
>> get quote :system/script
*** Script Error: get does not allow get-path! for its word argument
*** Where: get
*** Stack:  

>> get quote system/script:
*** Script Error: get does not allow set-path! for its word argument
*** Where: get
*** Stack:  

>> get quote 'system/script
*** Script Error: get does not allow lit-path! for its word argument
*** Where: get
*** Stack:  
```

**Expected behavior**

Same as
```
>> get 'system/script
== make object! [
    title: none
    header: none
    parent: none
    path: none
    a...
```

```
Red 0.6.4 for Windows built 16-May-2020/14:49:13+03:00
```

### LINKED ISSUE #4451: [Compiler] GET-PATH! literals are malformed
**Describe the bug**
`get-path!` literals are compiled as `get-path!` with `get-word!` at the head, as in:
```red
>> to get-path! [:foo bar]
== ::foo/bar
```

**To reproduce**
The following snippet:
```red
probe quote :foo/bar
```

When compiled, results in:
```red
::foo/bar
```

**Expected behavior**

```red
:foo/bar
```

**Platform version**

W10, `master` branch, commit 172c5fb1.

9214:
Closes #4448 and fixes #4451. #4448 was a trivial change, while #4451 was caused by Rebol2's peculiarity, mainly a complete absence of `get-path!` datatype.

Compiler encodes all `any-path!` values into Redbin payload as `block!`s, and treats `get-path!`s as `path!`s with `get-word!` at the head. This causes trouble with decoding, because `path!` allows only `word!` as its first element.

The (partial?) fix for that is to coerce `get-word!` to `word!` before storing `get-path!` into Redbin payload: this resolves the problem, at least for the cases where `get-path!` is `quote`d.

Tests for both issues are provided.

Also fixes #3624 (but see https://github.com/red/red/pull/4149).


hiiamboris:
Nice that you also fix `set` along the way. I just noticed it too lacked any-path support.

dockimbel:
@9214 The changes in %compiler.r are garbage, they don't even pass the loading stage... I've reverted them until you push a new corrected PR.

PR #4483, opened 2020-05-30
FEAT: improve docstrings of conditional natives.

### LINKED ISSUE #4469: ANY/ALL docstrings need clarification
`ANY` help brings you to the wrong conclusion that if a value is not `none` or `false` than it is true (logic).

```
>> help any
USAGE:
     ANY conds

DESCRIPTION: 
     Evaluates, returning at the first that is true. 
     ANY is a native! value.

ARGUMENTS:
     conds        [block!] 

>> any [false 22 33]
== 22
```
*...Evaluates, returning at the first that is* **true**.

But in Redbol logic:
```
>> true = 22
== false
>>
```

The same is needed for `ALL`
```

>> help all
USAGE:
     ALL conds

DESCRIPTION: 
     Evaluates, returning at the first that is not true. 
     ALL is a native! value.

ARGUMENTS:
     conds        [block!] 
```

I would suggest changing these help texts.

Discussion is [here](https://gitter.im/red/docs?at=5ecc27e6225dc25f54a5926e)

9214:
Resolves #4469 and addresses the points made in that issue.

### DIFF
--- a/environment/natives.red
+++ b/environment/natives.red
@@ -11,7 +11,7 @@ Red [
 ]
 
 if: make native! [[
-		"If conditional expression is TRUE, evaluate block; else return NONE"
+		"If conditional expression is truthy, evaluate block; else return NONE"
 		cond  	 [any-type!]
 		then-blk [block!]
 	]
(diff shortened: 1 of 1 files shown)


hiiamboris:
In the PR:
> all : Evaluates, returns the first falsy value; else NONE

What was proposed in https://github.com/red/red/issues/4469:
> all : Evaluates and returns the last value if all are truthy; otherwise NONE

PR #4484, opened 2020-05-30
FEAT: new semantics for ALL and ANY natives.

9214:
* `any` evaluates expressions one-by-one and returns the first truthy value, if any, else the last falsy value; defaults to `none`;
* `all` evaluates expressions one-by-one and returns the first falsy value, if any, else the last truthy value; defaults to `true`.

This differs from previous behavior, where both `any` and `all` defaulted to `none` and coerced falsy values to `none`.

The PR is missing compiler's support: for static cases where `all`/`any` is followed by a `block!`, compiler unrolls nested chain of expressions into a conditional tree of `if ... [...]` or `either ... [...][...]`. Current behavior conflicts with that logic and I haven't quite figured out yet how to rewrite it.

Nevertheless, I decided to draft it out and see if it's worth completing or if I should give up on trying to improve these short-circuit evaluation primitives.

### DIFF
--- a/environment/natives.red
+++ b/environment/natives.red
@@ -36,14 +36,14 @@ either: make native! [[
 ]
 	
 any: make native! [[
-		"Evaluates, returning at the first that is true"
+		"Evaluates, returns the first truthy value, if any, else the last falsy value; defaults to NONE"
 		conds [block!]
 	]
 	#get-definition NAT_ANY
 ]
 
 all: make native! [[
-		"Evaluates, returning at the first that is not true"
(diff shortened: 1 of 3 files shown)


dockimbel:
Where's the rationale for such change? Is there a REP for it?

9214:
@dockimbel I used [this](https://github.com/red/red/issues/4469#issuecomment-635464683) comparison as a guide, plus the rest of the discussion in #4610.

I am closing the PR and will open the REP with proper rationale for the change so that it can be discussed and decided upon.

PR #4487, opened 2020-06-01
FIX: refactor conditional usages of typeset macros.

9214:
This PR is informed mostly by `ref!`-related bugs and issues that come up time and again: it was implemented with an assumption that runtime relies on `TYPE_ANY_STRING` and `ANY_STRING?` macros. As can be seen in the applied changes, this did not hold in the majority of cases.

I also took the liberty to simplify `ANY_SERIES?` macro, remove dead `CHECK_COMPARE_OTHER` definition that wasn't used in the codebase, and refactor conditions that use `ANY_WORD?`.

Among the way the following inconsistency was fixed:
```red
>> head change "" <a>
== "a"
>> head insert "" <a>
== "<a>"
```

Now both actions preserve angle brackets of `tag!` values, the same way Rebol does.

### DIFF
--- a/runtime/datatypes/binary.reds
+++ b/runtime/datatypes/binary.reds
@@ -1300,13 +1300,7 @@ binary: context [
 					src: GET_TUPLE_ARRAY(cell)
 				]
 				default [
-					either any [
-						type = TYPE_STRING				;@@ replace with ANY_STRING?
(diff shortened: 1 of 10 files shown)


greggirwin:
Always good to leverage typesets. :+1:

Issue #4505, opened 2020-06-08
[Core] [Regression] interpreter started consuming a lot of RAM


hiiamboris:
**Describe the bug**

Try this script:
```
Red []
recycle/off
b: reduce [o: object []]
probe stats
loop 1000000 [find/same b o]
probe stats
```
Output of May 28th build `36857eb87` and before:
```
2221848        
2222324        
0:00:00.470027 
```
Current build `3d23acd0f`:
```
2240204      
234240912    
0:00:01.33608
```

**Expected behavior**

Stable memory usage.

**Platform version (please complete the following information)**
```
Red 0.6.4 for Windows built 8-Jun-2020/20:17:38
```

dockimbel:
@qtxie Please have a look to see which commit caused that.

qtxie:
This commit: https://github.com/red/red/commit/c2f0d8999dbd79ef458cfbe6e5d7c42f46d8b1fd

dockimbel:
It is caused by the copying of the call value on the stack, so that the arguments cache datastructures are linked to the value on the stack instead of the original value, resulting in losing the cache once the call ends, and having to rebuild it at every next call.

hiiamboris:
@dockimbel we are in the new chapter of the saga... ;)

```
Red []
; 1 % 2 - 3		;) spell of protection
recycle/off
ix: random collect [repeat i 10000 [keep i]]
print ["here" stats / 1e6 "M"]
sort/compare ix func [a b][loop 10 [1 - 1] a <= b]
print ["there" stats / 1e6 "M"]
recycle
print "freed"
```
Output:
```
here 3.110888 M
there 328.279812 M
freed
```
---
```
Red []
; 1 % 2 - 3		;) spell of protection
recycle/off
ix: random collect [repeat i 50000 [keep i]]
print ["here" stats / 1e6 "M"]
parse/trace ix [any skip] func [e m r y s][
	loop 10 [1 - 1]
	yes
]
print ["there" stats / 1e6 "M"]
recycle
print "freed"
```
Output:
```
here 4.683952 M
there 352.695444 M
freed
```
---
```
Red []
; 1 % 2 - 3		;) spell of protection
recycle/off
ix: append/dup "" "1 " 10000
print ["here" stats / 1e6 "M"]
transcode/trace ix func [e i t l k][
	loop 10 [1 - 1]
	yes
]
print ["there" stats / 1e6 "M"]
recycle
print "freed"
```
Output:
```
here 2.64916 M
[...]

PR #4529, opened 2020-06-17
FEAT: scalable metered reactivity

### LINKED ISSUE #4166: bugs in `react?`
**Describe the bug**

1. `react?/target` only works after `is`, not after `react`:
```
>> r: reactor [x: y: 1]
>> react [r/y: 2 * r/x]
== [r/y: 2 * r/x]
>> react? r 'x
== [r/y: 2 * r/x]
>> react?/target r 'y
== none
>> react?/target r in r 'y
== none
>> react?/target r 'r/y
*** Script Error: react? does not allow path! for its field argument
*** Where: react?
*** Stack: react?  
```
There should also be tests for this in %reactivity-test.red.

2. `react?/target` can be tricked:
```
>> r: reactor [x: 1 y: is [x * 2]]
>> o: object [on-change*: :r/on-change* x: 1 y: 2]
>> react?/target r 'y
== [x * 2]
>> react?/target o 'y
== [x * 2]            ;) bug - should be none
>> r/x: 3
== 3
>> r
== make object! [
    x: 3
    y: 6
]
>> o
== make object! [
    x: 1               ;) unaffected by the reaction, as it should be
    y: 2
]
```
To fix (2) `=?` should be used here: https://github.com/red/red/blob/19853a8b3cc3ed3b75960d21693d45519e8f2f6a/environment/reactivity.red#L233

3. Same problem with `react/unlink`:
https://github.com/red/red/blob/19853a8b3cc3ed3b75960d21693d45519e8f2f6a/environment/reactivity.red#L296
```
>> r: reactor spec: [x: 1 y: is [x * 2]]
>> o: object [on-change*: :r/on-change* x: 1 y: 2]
>> dump-reactions
1
:---  Source: object [x y]
   Field: x
  Action: [x * 2]
  Target: y
>> react/unlink last spec o
== [x * 2]          ;) react/unlink accepted `o` and removed a reaction on `r`
>> dump-reactions
>>
```

4. I'm lazy to invent a scenario to trip this bug, but this line doesn't seem logical at all:
https://github.com/red/red/blob/19853a8b3cc3ed3b75960d21693d45519e8f2f6a/environment/reactivity.red#L300 
So, if it found a `reaction` but the first found object didn't match the one expected, it should stop searching? Should be rewritten.

**To reproduce**

```
clear-reactions
r: reactor [x: y: 1]
react [r/y: 2 * r/x]
react? r 'x
react?/target r 'y
react?/target r in r 'y
react?/target r 'r/y
```
```
clear-reactions
r: reactor [x: 1 y: is [x * 2]]
o: object [on-change*: :r/on-change* x: 1 y: 2]
react?/target r 'y
react?/target o 'y
```
```
clear-reactions
r: reactor spec: [x: 1 y: is [x * 2]]
o: object [on-change*: :r/on-change* x: 1 y: 2]
dump-reactions
react/unlink last spec o
dump-reactions
```

**Expected behavior**

(1) `react?/target r 'y` should yield  `[r/y: 2 * r/x]`
(2) `react?/target o 'y` should yield `none`
(3) `react/unlink last spec o` should yield `none`
(4) I suggest rewriting the while loop to be more transparent

Note that #4165 should be fixed first. If that is done, many functions in `reactivity.red` will become one-liners and should be rewritten (including this one).

```
Red 0.6.4 for Windows built 27-Nov-2019/10:17:45+03:00 commit #b2aafe6
```

### LINKED ISSUE #4176: Regression: `is` not working with paths anymore
**Describe the bug**

It's fascinating how such a trivial change: https://github.com/red/red/commit/96fb9e48e8ad03084c3cfbfc0f7ac1a43e5fd672#diff-ff0dacd9d5c1090750f41b767ae4027f
could stop the reactivity from working properly

Test snippet:
```
r: make reactor! [n: 1 c: is [1x1 * n] x: is [c/x] t: is [n * 1:0:0] hms: is [rejoin [t/hour t/minute t/second]]]
? r  r/n: 2  ? r
```

Before that commit:
```
R is an object! with the following words and values:
     on-change*  function!     [word old new /local srs]
     n           integer!      1
     c           pair!         1x1
     x           integer!      1
     t           time!         1:00:00
     hms         string!       "100.0"

R is an object! with the following words and values:
     on-change*  function!     [word old new /local srs]
     n           integer!      2
     c           pair!         2x2
     x           integer!      2
     t           time!         2:00:00
     hms         string!       "200.0"
```
After that commit:
```
R is an object! with the following words and values:
     on-change*  function!     [word old new /local srs]
     n           integer!      1
     c           pair!         1x1
     x           integer!      1
     t           time!         1:00:00
     hms         string!       "100.0"

R is an object! with the following words and values:
     on-change*  function!     [word old new /local srs]
     n           integer!      2
     c           pair!         2x2
     x           integer!      1
     t           time!         2:00:00
     hms         string!       "100.0"
```

**Expected behavior**

`x` and `hms` should be updated properly

```
Red 0.6.4 for Windows built 27-Nov-2019/10:17:45+03:00 commit #b2aafe6
```

### LINKED ISSUE #4471: Initial reaction is triggered once for every occurrence of the same path
**Describe the bug**

See the title. This makes (sometimes complex) face initialization code fire multiple times resulting in slow GUI startup. Can be worked around, but since `react` also has no locals, it becomes ugly.
```
r: reactor [x: 1]
react [all [:r/x integer? r/x probe r/x]]
r/x: 2
r/x: 3
```
Output is:
```
1
1
1
2
3
```

**Expected behavior**
```
1
2
3
```

```
Red 0.6.4 for Windows built 27-May-2020/16:53:12+03:00 commit #cfb6d30
```

### LINKED ISSUE #4507: Slow `dump-reaction` output when one of the target objects is big
**Describe the bug**
```
v: view/no-wait compose/deep [panel [b: image 500x500 (make image! 2000x2000)]]
react/link func [v b] [? b/data] [v b]
clock [dump-reactions]
```
Output:
```
4787 ms	[dump-reactions]
```

Bug source: https://github.com/red/red/blob/459046699f88e3bd97e835ece2ea4c3fe820d059/environment/reactivity.red#L192

```
Red 0.6.4 for Windows built 28-May-2020/17:51:04+03:00 commit #36857eb
```

### LINKED ISSUE #4510: Possible to create invalid reactions that will never fire
**Describe the bug**

`react` & `react/link` do not check if after the object comes a word:
```
>> react/link/later func [a b] [probe a/('data)] [a b]
== func [a b][probe a/('data)]        ;) reactivity does not support such paths
>> dump-reactions             ;) yet it does add them
1
:---  Source: object [type offset size text image color menu data enabled? visible? selected flags option
   Field: data
  Action: func [a b][probe a/('data)]
    Args: [object [type: 'face offset: none size: none text: none image: none color: none menu: non
>> a/data: 1             ;) and they become a dead weight
== 1
```
```
>> clear-reactions
== []
>> react/later [probe a/('data)]
== [probe a/('data)]
>> dump-reactions
1
:---  Source: object [type offset size text image color menu data enabled? visible? selected flags option
   Field: data
  Action: [probe a/('data)]
```

**To reproduce**
```
clear-reactions
a: make face! []
b: make face! []
react/later [probe a/('data)]
react/link/later func [a b] [probe a/('data)] [a b]
dump-reactions
```

**Expected behavior**

Should ignore anything but words after the object, as such dynamic paths are out of reactivity's scope.

```
Red 0.6.4 for Windows built 28-May-2020/17:51:04+03:00 commit #36857eb
```

hiiamboris:
**TL;DR:** ~10x faster reactivity with O(1) complexity and multiple fixes here.

**Rationale.**

Since hash lookups are *very* cheap compared to Red token evaluation time, I decided to scratch the idea of putting relations into `on-change*` and use a central reactors index (hashtable). As I expected it's even faster than `select body-of :on-change* relations` approach I used before. Tests show further 20% speedup, but the number is not reliable as I'm on another PC which may have better optimized CPU cache.

**Performance**

Main source of speedup in this PR is avoiding O(n) lookup times by splitting the unstructured index into 2:
- `reactor [relations ...]`
- `reaction [source-objects ...]`

All operations become O(1) and it really scales.
You can see the benchmark results here: [for reactors](https://gist.github.com/hiiamboris/ad914d4e18c3e4400e8f0817bcda58e0), [for faces](https://gist.github.com/hiiamboris/5241fe596dfa05ec64659d16972cbddd) . I expect ~20-30% better timings with the fix for https://github.com/red/red/issues/4505 (it was fixed in this PR's snapshot).

There's still slight linearity O(n) for block reactions due to https://github.com/red/red/issues/4466 - once it is fixed I expect it to be nearly O(1) up to huge loads.

**Other notes**

Shallow reactors do not take ownership of assigned series anymore, so if some series is watched by a deep reactor, it can now safely be assigned to any number of shallow reactors

Objects not based on reactor! template can still use `reactivity/check` to embed themselves into reactivity framework (this part was also true in the original design)

I've also come to realize that `reactivity/check` should return blazingly fast for objects without relations.
E.g. in my Spaces project I'd like to make objects possibly reactive, but there may be thousands of objects, 99% of which won't be reactive, only a few top-level ones. And I need performance cost of adding reactivity on top of it to be negligible.
This implementation satisfies that need.

*Bonus* features are: better debugging output and the ability to collect statistics.

**Debugging** summary:
`system/reactivity/debug?` now can be:
- `off`/`none` to turn debug output off
- `true` or any other truthy value to turn it on for most interesting output, namely added, removed, skipped (most valuable!) and queued reactions
- `'full` (word) adds to `yes` a more extensive output: each fired reaction, checks for reactions, on-change events. Note: part of this output won't be shown in GUI console (as it'll crash with stack overflow). Use CLI console to see all the details.
- when it can find an object inside the global context, it will refer to it by word, not by the massive `object [....]` output, so it's much easier to tell what's what.

**Metering** summary:
- `system/reactivity/metrics?: true` turns on stats collection, `false` turns it off. Note: metering (when on) slows it down by ~20%, so use only when you need it.
- `system/reactivity/metrics/reset` zeroes all stats for a fresh start
- `system/reactivity/metrics/show` prints the full statistics, e.g.:
```
***** REACTIVITY METRICS REPORT *****
Metrics collection enabled?: true
Statistical counts:
    events triggered:    80015
    reactions fired:     50000 (immediately: 25005 , queued: 24995 )
    reactions skipped:   0
Time spent in reactions: 0:00:00
Time spent in reactivity:
    total:               0:00:09.55855
    adding relations:    0:00:03.80822 (preparations: 0:00:00.901052 )
    removing relations:  0:00:00.843048
    dispatching:         0:00:04.90728
    longest queue flush: 0:00:00.0020001
Peak values:
    maximum queue size:  5000
    maximum index size:  5000
    biggest relation:    2 reactors
    most used reactor:   5000 relations
```

I also made `is` much smarter:
https://github.com/red/red/blob/a77d72c37f55805b1122ccfee0c2046552081358/environment/reactivity.red#L302-L309

I have a lot of notes on pitfalls, tricks. If this PR gets approved, I will write them down to a wiki.

And along the way:
Fixes #4510, fixes #4507, fixes #4471, fixes #4176, fixes #4166, grants wish [REP#75](https://github.com/red/REP/issues/75)

### DIFF
--- a/environment/reactivity.red
+++ b/environment/reactivity.red
@@ -3,170 +3,267 @@ Red [
 	Author: "Nenad Rakocevic"
 	File: 	%reactivity.red
 	Tabs: 	4
-	Rights: "Copyright (C) 2016-2018 Red Foundation. All rights reserved."
+	Rights: "Copyright (C) 2016-2020 Red Foundation. All rights reserved."
 	License: {
 		Distributed under the Boost Software License, Version 1.0.
 		See https://github.com/red/red/blob/master/BSL-License.txt
 	}
 ]
 
-reactor!: context [
-	on-change*: function [word old new][
-		if system/reactivity/debug? [
-			print [
-				"-- on-change event --" lf
-				tab "word :" word		lf
-				tab "old  :" type? :old	lf
-				tab "new  :" type? :new
+system/reactivity: context [
+	;-- index format: [reaction [reactors..] ...]        -- required by react/unlink 'all, dump-reactions, clear-reactions, stop-reactor
+	index:       make hash! 1000						;-- hash speeds up unlinking noticeably
+
+	;-- queue format: [reactor reaction target done]
+	queue:		 make hash! 100
+
+	debug?: 	 no
+	metrics?:    no
+	source:		 []		;-- contains the initial [reactor reaction] that triggered a chain of subsequent reactions
+
+	metrics: context [
+		max-queue: max-index: max-reactors: max-relations: 0
+		events: fired: queued: skipped: 0
+		in-add: in-remove: in-react: in-check: in-eval: longest-flush: 0:0
+
+		reset: does [
+			set self 0
+			in-add: in-remove: in-react: in-check: in-eval: longest-flush: 0:0
+		]
+
+		show: does [
+			print "^/***** REACTIVITY METRICS REPORT *****"
+			print ["Metrics collection enabled?:" metrics?]
+			print  "Statistical counts:"
+			print ["    events triggered:   " events]
+			print ["    reactions fired:    " fired "(immediately:" fired - queued ", queued:" queued ")" ]
+			print ["    reactions skipped:  " skipped]
+			print ["Time spent in reactions:" in-eval]
+			print  "Time spent in reactivity:"
+			print ["    total:              " in-add + in-remove + in-react + in-check]
+			print ["    adding relations:   " in-add + in-react "(preparations:" in-react ")"]
+			print ["    removing relations: " in-remove]
+			print ["    dispatching:        " in-check]
+			print ["    longest queue flush:" longest-flush]
+			print  "Peak values:"
+			print ["    maximum queue size: " max-queue]
+			print ["    maximum index size: " max-index]
+			print ["    biggest relation:   " max-reactors "reactors"]
+			print ["    most used reactor:  " max-relations "relations"]
+		]
+		
+		start: target: none
+		time: func [/count 'into [word!] /save /local t] [
+			t: now/precise
+			case [
+				save  [set target (get target) + t: difference t start  t]
+				count [start: t  target: into]
 			]
 		]
+
+		++: make op! func ['word value] [set word add get word value]
+		peak:        func ['word value] [set word max get word value]
+
+		register: func ['counter value] [set counter max get counter value]
+	]
+
+	--measure--: func [code] [
+		if metrics? [do bind code metrics]
+	]
+
+	--debug-print--: function [blk [block!] /full /no-gui] [
 		all [
-			not empty? srs: system/reactivity/source
-			srs/1 = self
-			srs/2 = word
-			set-quiet in self word old					;-- force the old value
-			exit
+			debug?
+			any [not full  debug? = 'full]
+			not all [no-gui  attempt [system/console/gui?]]			;-- print in GUI console overflows stack in some places
+			(
+				limit: (any [all [system/console system/console/size/x] 72]) - 10
+				blk: reduce blk
+				forall blk [
+					x: :blk/1										;@@ workaround for #4517
+					unless string? :x [
+						if function? :x [x: body-of :x]
+						all [
+							object? :x
+							pos: find/same values-of system/words x
+							x: pick words-of system/words index? pos
+						]
+						change blk
+							replace/all
+								mold/flat/part :x limit
+								"make object!" "object"
+					]
+				]
+				if 1 < overshoot: (length? s: form blk) / limit [	;-- trim longest parts equally
+					foreach x next blk [							;-- don't trim the prefix
+						if 15 < n: length? x [
+							clear skip x to integer! n / overshoot
+						]
+					]
+					clear skip s: form blk limit
+				]
+				print s
+			)
 		]
-		unless all [block? :old block? :new same? head :old head :new][
-			if any [series? :old object? :old][modify old 'owned none]
-			if any [series? :new object? :new][modify new 'owned reduce [self word]]
-		]
-		system/reactivity/check/only self word
 	]
-]
 
-deep-reactor!: make reactor! [
-	on-deep-change*: function [owner word target action new index part][
-		system/reactivity/check/only owner word
+	relations-of: func [reactor [object!]] [first body-of :reactor/on-change*]
+
+	unique-objects: function [] [						;-- used by debug funcs only
+		unique collect [foreach [_ list] index [keep list]]
 	]
-]
+	
+	relations-count: function [] [
+		sum: 0
+		foreach obj unique-objects [sum: (length? relations-of obj) / 3 + sum]
+		sum
+	]
 
-reactor:	  function [spec [block!]][make reactor! spec]
-deep-reactor: function [spec [block!]][make deep-reactor! spec]
-
-
-system/reactivity: context [
-	relations:	 make block! 1000		;@@ change it to hash! once stable
-	;-- queue format: [reactor reaction target done]
-	queue:		 make block! 100
-	eat-events?: yes
-	debug?: 	 no
-	source:		 []
-
-	add-relation: func [
+	add-relation: function [
 		obj		 [object!]
-		word
+		word     [word!]
 		reaction [block! function!]
 		targets  [set-word! block! object! none!]
-		/local new-rel
 	][
-		new-rel: reduce [obj :word :reaction targets]
-		unless find/same/skip relations new-rel 4 [append relations new-rel]
+		--measure-- [time/count in-add]
+		new-rel: head reduce/into [:word :reaction targets] clear []
+		relations: relations-of obj
+		unless find/same/skip relations new-rel 3 [
+			append relations new-rel
+			unless objs: select/only/same/skip index :reaction 2 [
+				reduce/into [:reaction objs: make block! 10] tail index
+			]
+			unless find/same objs obj [append objs obj]
+			if block? targets [							;-- react/link func .. [func objects..] case
+				foreach obj next targets [
+					unless find/same objs obj [append objs obj]
+				]
+			]
+			--measure-- [
+				peak max-relations (length? relations) / 3
+				peak max-reactors   length? objs
+				peak max-index     (length? index) / 2
+			]
+			--debug-print-- ["-- react: added --" :reaction "FOR" word "IN" obj]
+		]
+		--measure-- [time/save]
 	]
 	
-	eval: function [code [block!] /safe][
+	eval: function [code [block!] /safe /local result saved][
+		--measure-- [
+			time/count in-eval
+			fired ++ 1
+		]
+		--debug-print--/full ["-- react: firing --" either function? first code [body-of first code][code]]
 		either safe [
-			if error? set/any 'result try/all code [
-				print :result
+			if error? error: try/all [set/any 'result do code  'ok] [
+				print :error
 				prin "*** Near: "
 				print mold/part/flat code 80
-				result: none
 			]
-			get/any 'result
 		][
-			do code
+			set/any 'result do code
 		]
+		--measure-- [time/save]
+		:result
 	]
 	
-	eval-reaction: func [reactor [object!] reaction [block! function!] target /mark][
-		if mark [repend queue [reactor :reaction target yes]]
-		
-		either set-word? target [
-			set/any target eval/safe :reaction
-		][
-			eval/safe any [all [block? :reaction reaction] target]
+	eval-reaction: func [reactor [object!] reaction [block! function!] target [set-word! block! object! none!]][
+		case [
+			set-word? target [set/any target eval/safe :reaction]
+			block? :reaction [eval/safe reaction]
+			'linked          [eval/safe target]
 		]
 	]
 	
 	pending?: function [reactor [object!] reaction [block! function!]][
-		q: queue
-		while [q: find/same/skip q reactor 4][
-			if same? :q/2 :reaction [return yes]
-			q: skip q 4
-		]
-		no
+		pattern: head reduce/into [reactor :reaction] clear []
+		none <> find/same/skip queue pattern 4
 	]
-	
-	check: function [reactor [object!] /only field [word! set-word!]][
-		unless empty? pos: relations [
-			while [pos: find/same/skip pos reactor 4][
-				reaction: :pos/3
-				if all [
-					any [not only pos/2 = field]
-					any [empty? queue  not pending? reactor :reaction]
-				][
-					either empty? queue [
-						if empty? source [
-							append source reactor
-							append source field
-						]
-						eval-reaction/mark reactor :reaction pos/4
-						
-						q: tail queue
-						until [
-							q: skip q': q -4
-							either q/4 [ 				;-- was already executed?
-								clear q 				;-- allow requeueing of it
-							][
-								eval-reaction q/1 :q/2 q/3
-								either tail? q' [ 		;-- queue wasn't extended
-									clear q 			;-- allow requeueing
-								][
-									q/4: yes 			;-- mark as executed
-									q: tail queue 		;-- jump to recently queued reactions
-								]
-							]
-							head? q
-						]
-						clear queue
-						clear source
-					][
-						unless all [
-							eat-events?
-							pending? reactor :reaction
+
+	check: function [reactor [object!] field [word! set-word!] /local word reaction target][
+		--debug-print--/full/no-gui ["-- react: checking --" field "IN" reactor]
+		--measure-- [time/count in-check]
+		pos: relations-of reactor
+		unless pos: find/skip pos field 3 [exit]
+		if initial?: tail? source [reduce/into [reactor field] source]
+		until [
+			set [word reaction target] pos
+			case [
+				pending? reactor :reaction [			;-- don't allow cycles
+					--measure-- [skipped ++ 1]
+					--debug-print--/no-gui ["-- react: skipped --" :reaction "FOR" field "IN" reactor]
+					'idle
+				]
+				not tail? queue [						;-- entered while another reaction is running
+					reduce/into [reactor :reaction target no] tail queue
+					--measure-- [
+						queued ++ 1
+						peak max-queue (length? queue) / 4
+					]
+					--debug-print--/no-gui ["-- react: queued --" :reaction "FOR" field "IN" reactor]
+				]
+				'else [
+					reduce/into [reactor :reaction target yes] queue
+					--measure-- [peak max-queue 1  time/save]
+					eval-reaction reactor :reaction target
+					q: tail queue
+					while [not head? q] [
+						q: skip q': q -4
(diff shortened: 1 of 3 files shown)


environment/reactivity.red
```diff
@@ -3,246 +3,426 @@ Red [
...
+			case [
+				save  [set target (get target) + t: difference t start  t]
+				count [start: t  target: into]
+			]
 		]
-		system/reactivity/check/only self word
+
+		incr: func ['word value] [set word add get word value]
+		peak: func ['word value] [set word max get word value]
+		++: make op! :incr								;@@ workaround for #4526
```
greggirwin, line 66: How is this better than just using `incr`? `Incr` itself is a mezz I've tinkered up, and decided a lit arg wasn't best (for a general version).

environment/reactivity.red
```diff
@@ -3,246 +3,426 @@ Red [
...
-	eat-events?: yes
-	debug?: 	 no
-	source:		 []
+	unique-objects: function [] [						;-- used by debug funcs only
+		unique collect [foreach [_ list] index [keep list]]
+	]
+	
+	relations-count: function [] [
+		sum: 0
+		foreach obj unique-objects [sum: (length? relations-of obj) / 3 + sum]
```
greggirwin, line 119: Might be nice to have a word for the relation record size (3), since it appears a number of places as a magic number.

environment/reactivity.red
```diff
@@ -254,96 +434,129 @@ system/reactivity: context [
...
 ]
 
+reactor!: context [
+	on-change*: function [word old [any-type!] new [any-type!]] [
+		;-- relations format: [reactor word reaction targets]
+		;;  src-word [reaction] set-word            	 -- used by `is` (evaluates reaction, assigns to target)
+		;;  src-word function [func obj1 obj2...]   	 -- used by react/link (evaluates target), one relation for every reactor in both list and func's body
+		;;  src-word [reaction] none                	 -- used by react (evaluates reaction)
+		;;  src-word [reaction] set-word/object     	 -- used by react/with (evaluates reaction, assigns to a set-word only)
+		[]												;-- relations placeholder (hash is ~10x times slower)
```
greggirwin, line 531: Why an anonymous block, rather than named?

greggirwin (commented):
Great writeup. I appreciate all the included tests and metrics.

environment/reactivity.red
hiiamboris: `++` just better visually stands out, that's all. I wanted `peak` to be an infix too, but it is used with `length? ..` expressions and as an op will require an extra set of parens so I left it like this, and `incr` in an infix form.

hiiamboris: Actually, compiler gives me the headache sometimes about `++`, so I'm open to removing it.

environment/reactivity.red
hiiamboris: Perhaps.

environment/reactivity.red
hiiamboris: I just settled on the simplest thing. Why name if the name won't be used? Then I found another reason - you don't write thing like this by accident. So it's harder to mistake a reactor by-design for reactor by-accident and modify something you don't want to. Technically we could access relations as `body: body-of :reactor/on-change*  relations: body/relations` (if it's defined as `relations: []`). So if you wish, let's do that.

environment/reactivity.red
greggirwin: I love it when the compiler is on my side. ;^) `++` is problematic, historically (in C), with pre/post ambiguities. Ops like `+=` and `-=` as standard, along with `++` and `--` is something for the Red Shoud... wiki page.

environment/reactivity.red
greggirwin: It jumped out at me because it looked like an accident. For those who find it, they have to face, front-on, the "it's all data" aspect (or lose their minds trying). Carl often used positional data, and it's more efficient to be sure...until you have to read and understand it. It's also more brittle. It's not likely that many people will touch this code, but if we don't name it, I'll ask for big warning signs like "DO NOT ADD ANYTHING BEFORE THIS BLOCK OR YOU WILL BREAK THE WORLD REACTIVELY!" 

Another aspect, in the big picture, is stuff like AOP, DbC, instrumentation. If you have a positional item at the head, you can't automatically add a prolog. This is a deep subject on the exit side, with early return points and epilogues.

hiiamboris: I'm actually adding stuff before this block. Just have to insert another block again at the head ;) But I agree, with word reference I won't have to add it.

hiiamboris:
Another side effect of this PR - less GC load during face creation.
In the master branch:
```
 >> loop 2000 [make face! []] ()                                              
root: 4655/6708, runs: 0, mem: 3250492 => 2177928, mark: 2.0ms, sweep: 1.0ms 
root: 4655/6708, runs: 1, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 0.0ms 
root: 4655/6708, runs: 2, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms 
root: 4655/6708, runs: 3, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms 
root: 4655/6708, runs: 4, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 1.0ms 
root: 4655/6708, runs: 5, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 1.0ms 
root: 4655/6708, runs: 6, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms 
root: 4655/6708, runs: 7, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms 
root: 4655/6708, runs: 8, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms 
root: 4655/6708, runs: 9, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 0.0ms 
root: 4655/6708, runs: 10, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms
root: 4655/6708, runs: 11, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms
root: 4655/6708, runs: 12, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms
root: 4655/6708, runs: 13, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 1.0ms
root: 4655/6708, runs: 14, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms
root: 4655/6708, runs: 15, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms
root: 4655/6708, runs: 16, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 0.0ms
root: 4655/6708, runs: 17, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms
root: 4655/6708, runs: 18, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 1.0ms
root: 4655/6708, runs: 19, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 1.0ms
root: 4655/6708, runs: 20, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms
root: 4655/6708, runs: 21, mem: 3250440 => 2170640, mark: 4.0ms, sweep: 0.0ms
root: 4655/6708, runs: 22, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 1.0ms
root: 4655/6708, runs: 23, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms
root: 4655/6708, runs: 24, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 1.0ms
root: 4655/6708, runs: 25, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms
root: 4655/6708, runs: 26, mem: 3250440 => 2170640, mark: 4.0ms, sweep: 1.0ms
root: 4655/6708, runs: 27, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 0.0ms
root: 4655/6708, runs: 28, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 1.0ms
root: 4655/6708, runs: 29, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms
root: 4655/6708, runs: 30, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 0.0ms
root: 4655/6708, runs: 31, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms
root: 4655/6708, runs: 32, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms
root: 4655/6708, runs: 33, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 0.0ms
root: 4655/6708, runs: 34, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms
root: 4655/6708, runs: 35, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 1.0ms
root: 4655/6708, runs: 36, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms
root: 4655/6708, runs: 37, mem: 3250440 => 2170640, mark: 2.0ms, sweep: 1.0ms
root: 4655/6708, runs: 38, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 1.0ms
root: 4655/6708, runs: 39, mem: 3250440 => 2170640, mark: 3.0ms, sweep: 1.0ms
>>                                                                           
```

In this PR:
```
>> loop 2000 [make face! []] ()                                              
root: 4826/6876, runs: 0, mem: 3249396 => 2244272, mark: 3.0ms, sweep: 0.0ms 
root: 4826/6876, runs: 1, mem: 3251932 => 2259820, mark: 3.0ms, sweep: 1.0ms 
root: 4826/6876, runs: 2, mem: 3251932 => 2259820, mark: 4.0ms, sweep: 1.0ms 
root: 4826/6876, runs: 3, mem: 5411344 => 2254624, mark: 3.0ms, sweep: 2.0ms 
root: 4826/6876, runs: 4, mem: 7570784 => 2249456, mark: 4.0ms, sweep: 4.0ms 
root: 4826/6876, runs: 5, mem: 7570784 => 2249456, mark: 3.0ms, sweep: 3.0ms 
root: 4826/6876, runs: 6, mem: 7570784 => 2249456, mark: 3.0ms, sweep: 4.0ms 
root: 4826/6876, runs: 7, mem: 7570784 => 2249456, mark: 3.0ms, sweep: 4.0ms 
root: 4826/6876, runs: 8, mem: 7570784 => 2249456, mark: 3.0ms, sweep: 4.0ms 
root: 4826/6876, runs: 9, mem: 7570784 => 2249456, mark: 3.0ms, sweep: 4.0ms 
root: 4826/6876, runs: 10, mem: 7570784 => 2249456, mark: 3.0ms, sweep: 4.0ms
>>                                                                           
```

hiiamboris:
Last commit makes tight (`r/x: r/x + 1`) loops ~5% slower but reduces reactor's memory consumption by a factor of four: from 5.7kB to 1.4kB (on master branch it's 4.5kB).

greggirwin:
Thanks for continuing work on this. I hope @dockimbel will be able to review soon.

environment/reactivity.red
```diff
@@ -3,246 +3,459 @@ Red [
...
-	on-change*: function [word old new][
-		if system/reactivity/debug? [
-			print [
-				"-- on-change event --" lf
-				tab "word :" word		lf
-				tab "old  :" type? :old	lf
-				tab "new  :" type? :new
-			]
+#local [
+#macro REL-PERIOD: func [][3]
```
greggirwin, line 14: Macros seem like overkill here, even if simple, compared to just int refs.

hiiamboris: Why? They will become integers during compilation.
TBH I just didn't want them to be "variables", as that implies one could change their values, so I used macro as a sort of constant declaration, same as R/S code does throughout.

environment/reactivity.red
```diff
@@ -3,246 +3,459 @@ Red [
...
-	on-deep-change*: function [owner word target action new index part][
-		system/reactivity/check/only owner word
+	--measure--: func [code] [
+		if metrics? [do bind code metrics]
 	]
-]
 
-reactor:	  function [spec [block!]][make reactor! spec]
-deep-reactor: function [spec [block!]][make deep-reactor! spec]
+	--debug-print--: function [blk [block!] /full /no-gui] [
```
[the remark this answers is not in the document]
hiiamboris: I started with undecorated names only to find out that I am unable to tell apart actual code that does the job from supplemental metering and debugging code, so I changed it to the way it is.

Propose better names, but please test how they look first in the editor, e.g. for `check` func. If it's cleaner I'm all for it! :)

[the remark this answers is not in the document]
hiiamboris: No more `sr/` prefix in the recent update.

hiiamboris:
I'd like to make Spaces reactive, but this has been ignored for almost two years already.

greggirwin:
I think @dockimbel has looked at it, but wasn't sure about the design. Please weigh in when you can @dockimbel.

PR #4536, opened 2020-06-22
FIX: support more delimeter-match for console engine

bitbegin:


bitbegin:
waiting `"#([]22)"` to be fixed

environment/console/engine.red
```diff
@@ -140,27 +139,56 @@ system/console: context [
...
+				]
+				if all [												;-- block! paren! map! have open-event, so just match delimiters
+					any [
+						type = block!
+						type = paren!
+						type = map!
+					]
+					delimiter-map/:type = last delimiters
+				][
+					throw 'break
```
dockimbel, line 160: What are those `throw 'break`? There is no `catch` for those values, only for `'stop`...

Issue #4203, opened 2019-12-23
String CORRUPTiON involving MAKE, PARSE and Unicode source file


hiiamboris:
**Describe the bug**

Run this as a standalone script (`red myfile.red` or whatever):
```
Red []

; without this line there's no string corruption (any unicode char does the trick):
; ☺

; at least one argument in spec must be 2 or more chars long
spec: [yz [float!] x]

probe collect [
	=arg=: [
		set name word!
		opt [set types block!]
		(keep probe reduce [name  probe make typeset! types])
	]
	parse spec [any [=arg=]]
]
```

It prints to console:
```
make typeset! [float!]                              
[yz make typeset! [float!崠]                         
make typeset! [float!]                              
[x make typeset! [float!]]                          
[yz make typeset! [float!崠 x make typeset! [float!崠]
```
Note the hieroglyphics ;)

**Expected behavior**

Expected `]` (5Dh) in place of char `崠` (5D20h)

**Platform version (please complete the following information)**
```
Red 0.6.4 for Windows built 20-Dec-2019/19:03:46+03:00 commit #544a6e1
```

qtxie:
FYI. No problem in the fast-lexer branch.

hiiamboris:
Great!

hiiamboris:
Since lexer is merged, this only waits for a regression test.

theSherwood:
Adding the regression test is on hold until we have a dedicated script for handling file creation from within tests.

PR #4583, opened 2020-07-21
TESTS: regression-test-red.red - issue #4203

theSherwood:


theSherwood:
This pull request needs some work.

theSherwood:
This is on hold on account of @hiiamboris wanting a dedicated script for handling creating and deleting files in regression tests.

hiiamboris:
To clarify, I just want temp junk to be contained within /quick-test/runnable/, that's all.

theSherwood:
@hiiamboris Ah. Okay. So we don't need a dedicated script; we just need to make sure that we are writing files to that directory?

theSherwood:
The test file is now being created in `/quick-test/runnable/`.

theSherwood:
We have failing tests as of commit 272c9ffa741aed85191b474d75c85f0001ab3182.

9214:
Likely this test itself is failing.

theSherwood:
@9214 Is 272c9ff passing all tests for you? It's entirely possible I made a mistake tracking this down.

theSherwood:
Interesting, this commit was passing for me locally, but master was not. I tracked it back to 272c9ff as the first commit on master where tests started failing. I must have borked my local repo somehow.

theSherwood:
Hmm. I just ran all tests 5 times on master. Iterations 1-3 failed. 4 and 5 succeeded.

9214:
@theSherwood can you check the content of `./quick-test/quick-test.log` file for failed iterations?

theSherwood:
If I can get it to fail again, I will. I didn't think to check it on the previous failed runs, and then the logs were overwritten by the successful ones.

theSherwood:
@9214 I ran `rebol -qs run-all.r` on 272c9ff

I got two failing tests: `run-all-comp1` and `run-all-interp`.

These are the relevant results in the log:

```
signify failure
~~~started test~~~ run-all-comp1




*** Syntax Error: (line 1) invalid issue at #abc:)
*** Where: transcode
*** Stack: 
```

```
signify failure
~~~started test~~~ run-all-interp
*** Syntax Error: (line 1) invalid issue at #abc:)
*** Where: transcode
*** Stack: --assert 
```

The mode of failure appears to be the same as what we have in Appveyor. No tests in these files were successfully run. On a successful run, there are lots of tests in these files. So it looks like Appveyor is experiencing the same Syntax Errors as what I'm getting.

9214:
Thanks @theSherwood, that's helpful.

https://github.com/red/red/blob/272c9ffa741aed85191b474d75c85f0001ab3182/tests/source/units/lexer-test.red#L407

https://github.com/red/red/blob/02776dcf50c085b3fe9c207df88bdaceb8c5b3c0/tests/source/units/lexer-test.red#L407

If you sync your branch with the upstream, resolve potential merge conflicts and re-run all the tests (including the one that you wrote), does the problem still persist?

hiiamboris:
IIRC `#abc:` thing was forbidden a few days ago because of it's ambiguity in paths: is `map/#abc:` a path with an issue ending with colon or a set-path with an issue ending with letter?

Still a bit weird though:
```
>> m/#abc: 1
*** Syntax Error: (line 1) invalid issue at #abc: 1
*** Where: case
*** Stack: load
>> m/#abc
== none 
```
Perhaps set-paths should be extended?

theSherwood:
@9214 I merged master into my branch. It fails in the same way now.

9214:
@theSherwood assuming that by "fails in the same way" you mean that the logs look the same, how does the test `"tr-47"` look like (`./tests/source/units/lexer-test.red` file)? Are you sure you cleaned up all auto-generated test files before running QT framework again? It happens with me sometimes that I forget to purge them and end up re-running the old tests.

theSherwood:
I have once again pulled merged master into my branch, deleted all auto-generated test files (`./test/source/units/auto-tests/`, `./quick-test/runnable`), and run `rebol -s run-all.r`.

The first time all tests passed. I then, once again deleted all auto-generated test files and ran `rebol -s run-all.r`. This time, tests failed.

### console:
```
lexer auto-test files generated
checking to see if interpreter test files need generating
61 files were generated
Quick-Test v 0.12.0
REBOL  2.7.8.2.5
This test started at 7-Aug-2020/21:12:10.669628-6:00
Complete Red Test Suite

ok - preprocessor tests..............................39 / 39
** - run-all-comp1*red********************************0 / 1 **
ok - run-all-comp2.red.............................5246 / 5246
```

### test-logs:
```
signify failure
~~~started test~~~ run-all-comp1




ok
*** Script Error: OK has no value
*** Where: do
*** Stack: do-file 
```

and 

```
signify failure
~~~started test~~~ run-all-interp


ok
*** Script Error: OK has no value
*** Where: do
*** Stack: --assert do-file 
```

### lexer-test.red:
```
	--test-- "tr-47" --assert error? try [transcode "(#abc:)"]
```



I'm going to go through the process one more time.

theSherwood:
Okay. I went through it all once more. Deleted auto-generated files and ran `rebol -s run-all.r`. I got the same failures that I got on the last run.

On a somewhat unrelated note: running `rebol -s run-all.r` does annoying things with window focus. Is anyone else getting that?

theSherwood:
I did have a problem with the test. I was getting an error at `do test-file`. However, this worked when I ran the test alone with `rebol -s quick-test/run-test.r "<file>"`. It would fail when I ran `rebol -s run-all.r`. In order to get it to pass, I changed the test to use `do read test-file`, which I'm no longer sure is even testing the correct thing. I tried testing this with an older version of the compiler to make sure it the test would fail (using `rebol -qs run-all.r --binary <path-to-compiler>`), but quick-test would simply hang (before reaching the tests in question).

Thoughts on how I should proceed?

hiiamboris:
Such things should not appear in a healthy code. I suggest using `qt-verbose: yes` flag and locating the test where it hangs, and then a particular line (manually) and trying to identify the cause. Although all my bets are on https://github.com/red/red/issues/4604 currently

9214:
@theSherwood what was the error signature with `do test-file`?

theSherwood:
@hiiamboris and @9214  I will definitely take a look at that and try to get to the bottom of it.

I found that we also are getting an Access Error.
```
ok - preprocessor tests..............................39 / 39
** - run-all-comp1*red********************************0 / 1 **
ok - run-all-comp2.red.............................5246 / 5246
```

This is from the test log:
```
*** Access Error: cannot open: <path>/red/tests/quick-test/runnable/test.red
*** Where: write
*** Stack: 
```

We get that error if we run `rebol -s tests/run-all.r`, which is what Appveyor is running. I have been running `rebol -s run-all.r` locally, which is passing. The reason for the different behavior is in `quick-test/quick-test.red` where I am setting the value of `runnable-dir` using `what-dir` (though I get the same problem with `system/options/path`). 

`runnable-dir` is the path to `quick-test/runnable` where I'm putting files generated by tests.

When `rebol -s tests/run-all.r` is run, `runnable-dir` ends up referring to an erroneous path: `.../red/tests/quick-test/runnable/`. Whereas `rebol -s run-all.r` causes it to refer to the correct path: `.../red/quick-test/runnable/`. And we clearly need it to work from both commands. 

I could correct the path if `rebol -s tests/run-all.r` is run, but what I have in mind is just string manipulation, which feels like a really brittle solution to this issue. I'm still as yet largely unfamiliar with red in general and with this codebase; so does anyone have a better idea of how to address this issue?

---

In any case, I think this is why I've been getting different behavior from appveyor at various points.

theSherwood:
@hiiamboris As for the hanging tests when run with an old build of the compiler, I ran the following:

```
rebol -qtws run-all.r --binary <path to compiler> --do "verbose: yes"
```

My results were not hugely useful, unfortunately. When I ran the above command, instead of hanging, the window crashed. However, from what I could see of the traces, it crashed at different places on different runs. I'm not sure what to do with that.

9214:
@theSherwood let's take a few steps back and ask ourselves if we really need to create a file in order to reproduce the issue.
The answer is: we don't.

```red
--test-- "#4203"
	--assert "[yz make typeset! [float!]]" = do {
		; ☺
		mold collect [
			parse [yz [float!]][
				set name word!
				set types block!
				(keep reduce [name make typeset! types])
			]
		]
	}
```
Pasted in the regression test file will work as well. E.g. on 544a6e12 where the issue was originally reported I get:
```red
"[yz make typeset! [float!崠]"
```
And the test fails both when run alone (compiled and interpreted) and as a part of the Red test suite:
```
~~~started test~~~ regression-test
ok
===group=== regressions #2001+
--test-- #4203 FAILED**************
~~~finished test~~~  regression-test
  Number of Tests Performed:      236
  Number of Assertions Performed: 457
  Number of Assertions Passed:    456
  Number of Assertions Failed:    1
****************TEST FAILURES****************
```

hiiamboris:
> does anyone have a better idea of how to address this issue?

I would probably go up the directory tree until I find `quick-test/`.

> However, from what I could see of the traces, it crashed at different places on different runs. I'm not sure what to do with that.

Try the #4604 fix, by merging master in, or just the fix: `git cherry-pick df42056`

@9214 I think it's better to isolate the script, as tests are already a huge twisted mess of intermingled side effects. What's the guarantee that the test won't be affected by other tests? Or vice versa?

9214:
@hiiamboris it doesn't matter if the source comes from `string!` or from `file!`, they all eventually bottom down on `scan-alt` (or `system/lexer/transcode` in the old lexer).

That's in `do`:
https://github.com/red/red/blob/4c0c317f53c4e165d5959d01fdd2e039e181ac55/runtime/natives.reds#L563-L568

That's in `do-file`:
https://github.com/red/red/blob/4c0c317f53c4e165d5959d01fdd2e039e181ac55/environment/functions.red#L879-L884

That's in `load`:
https://github.com/red/red/blob/4c0c317f53c4e165d5959d01fdd2e039e181ac55/environment/functions.red#L453

That's in `transcode`:
https://github.com/red/red/blob/4c0c317f53c4e165d5959d01fdd2e039e181ac55/runtime/natives.reds#L2838

 I just demonstrated that the issue can be reproduced on 544a6e1 even if you use a `string!`. And, frankly, we are testing an issue with old Parse-based lexer on a new lexer, I doubt there are any chances of it showing up in the reported form at all — it's not like Red can suddenly regress a year back.

I understand the need for file creation in temp directory, but perhaps it should be tackled separately from this regression test?

hiiamboris:
It's not like it will be a huge change ;) I don't care personally.

9214:
@theSherwood what about `qt/runnable-dir`? Supposedly QuickTest already knows where it is:
https://github.com/red/red/blob/4c0c317f53c4e165d5959d01fdd2e039e181ac55/quick-test/quick-test.r#L26-L33

theSherwood:
Thanks @hiiamboris and @9214 . I ended up doing the path normalizing stuff rather than using `system/options/boot` because I figured that might not work if using a compiler passed with the `--binary` flag. So everything is passing at the moment. However, we aren't out of the woods yet. 

I'm still left with a weird problem on line 2891:

`do test-file` causes the whole suite to break (as we've seen multiple times before) when `rebol -qws tests/run-all.r --batch` is run. But it causes no problems if simply running the one test file with `rebol -s quick-test/run-test.r <path to file>`. I have not been able to figure this out.

This is what we get in the logs when that fails:

```
signify failure
~~~started test~~~ run-all-comp1




ok
*** Script Error: OK has no value
*** Where: do
*** Stack: do-file
```

So instead I committed `do read test-file`, which doesn't cause `rebol -qws tests/run-all.r --batch` to fail.

hiiamboris:
Weird. Where does 'OK' come from?

9214:
I suspect preprocessor is involved: `do <file>` runs it, `do <string>` does not. And in fact, the error message looks awfully familiar. This is a signature of one nasty little fella:
https://github.com/red/red/blob/00840830f1103649c176f8becffd2afb398ea628/tests/source/units/parse-test.red#L2727

@theSherwood try to put `#reset` right after `WTF`. I already fixed it twice in two different PRs, but none of them were merged, so... 🤷‍♂️

theSherwood:
@9214 I made that change, and it passed locally and it passed appveyor. I think Travis must be responding to the recent GTK stuff. If that's the case, then this ought to finally be ready to go!

9214:
@theSherwood cool! I think the issue with ARM builds was resolved, so let's re-run the tests (sync your branch with the upstream, then push the changes) and check that they are greenlit.

theSherwood:
@9214 I merged in master and pushed it. But Travis-ci is still failing. If I'm understanding the details correctly, there's still something wrong with ARM tests.

hiiamboris:
This PR introduced some erroneous paths:
```
~~~started test~~~ run-all-interp


ok
*** Access Error: cannot open: %/red/quick-test/runnable/arm-tests/red/quick-test/runnable/test.red
*** Where: write
*** Stack: 
```
But on ARM only

Issue #4119, opened 2019-11-03
Confusing error message on encountering $ inside block


meijeru:
I was formulating a parse rule for strings not incorporating a `$` sign. I would have liked to call the rule `non$` but `$` is not allowed in words. That is fair, but when such a non-word is inside a block, the interpreter gives a confusing error message. It should of course have been the same as for the first case.

```
>> non$
*** Syntax Error: invalid value at "$"
*** Where: do
*** Stack: load 
>> rule: [ any non$ "something else"]
*** Syntax Error: missing #"]" at "$]"
*** Where: do
*** Stack: load 

```

greggirwin:
`\` as well. 
```
>> rule: [ any non\ "something else"]
*** Syntax Error: missing #"]" at {\ "something else"]}
*** Where: do
*** Stack: load 
```

meijeru:
Most likely all non-allowed characters in `word!` values gives this confusing message, so it must be a single imperfection in the lexer. Given that the lexer is being redesigned, one should remember to test this with the new lexer as soon as it is in operation, so as to hopefully discard this issue then.

9214:
I propose we close it.

```red
>> $non
*** Syntax Error: (line 1) invalid money at $non
*** Where: transcode
*** Stack: load 
>> non\
*** Syntax Error: (line 1) invalid character at \
*** Where: transcode
*** Stack: load 
````

meijeru:
I am OK with closing it if there are the tests I mentioned.

theSherwood:
Tests were added. So I suspect that this can be closed.

PR #4584, opened 2020-07-21
TESTS: regression-test-red.red - issue #4119

theSherwood:
Expected:
```
>> [$non]
*** Syntax Error: (line 1) invalid money at $non]
*** Where: case
*** Stack: load 
```
(As opposed to a less readable error)

### DIFF
--- a/tests/source/units/lexer-test.red
+++ b/tests/source/units/lexer-test.red
@@ -734,8 +734,10 @@ Red [
 	--test-- "scan-68" --assert none!	 = scan "#[none]"
 	--test-- "scan-69" --assert integer! = scan "#[integer!]"
 	--test-- "scan-70" --assert error!	 = scan "#[int!]"
-	--test-- "scan-71"  --assert refinement! = scan "/v:"
-	--test-- "scan-72"  --assert refinement! = scan "/value:"
+	--test-- "scan-71" --assert refinement! = scan "/v:"
+	--test-- "scan-72" --assert refinement! = scan "/value:"
+	--test-- "scan-73" --assert error!   = scan "$non"
+	--test-- "scan-74" --assert error!   = scan "\non"
 
 ===end-group===
 ===start-group=== "scan/fast"
--- a/tests/source/units/regression-test-red.red
+++ b/tests/source/units/regression-test-red.red
@@ -2853,20 +2853,6 @@ comment {
 		--assert tail? next next next next i4056
 		--assert tail? next back next tail i4056
 
-	--test-- "#4119"
-		word-money-4119:  try [load {$non}]
-		block-money-4119: try [load {[$non]}]
-		--assert to-logic find to-string word-money-4119 "(line 1) invalid money at $non"
-		--assert to-logic find to-string block-money-4119 "(line 1) invalid money at $non]"
-		word-slash-4119:  try [load {\non}]
-		block-slash-4119: try [load {[\non]}]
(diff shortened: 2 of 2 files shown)


greggirwin:
I guess we need to think about tests and what we compare in errors. e.g. `to-logic find to-string ...`. Looks like other tests check more directly against the error, e.g. `/id  /type`.

theSherwood:
That's a good point. I first wrote the test checking against `/id`, `/type`, `/arg1`, `/arg2`, and `/arg3` but it seemed pretty verbose (several more checks) and that it somewhat obfuscated the purpose of the test, which was to test that that the error was readable. Would you prefer that I check against `/id` etc?

hiiamboris:
IMO just `error? try load []` would have been enough. Error messages may change, and that incurs additional housekeeping.

theSherwood:
It was reliably producing an error before. The issue was to make sure the error was human readable, if I understand correctly.

hiiamboris:
Yeah. I wrote too quickly. Fixed my message.

9214:
#4119 itself was concerned with the old lexer and is no longer relevant. The provided tests are written using the new lexer, and since one cannot get the original error messages anymore I'm not sure how useful they are.

hiiamboris:
You change error messages -> tests fail -> you look up the issue and check if what you did make sense -> you change tests. Or you change the lexer -> it reports something else -> tests fail -> you look up the issue and realize you broke it ;)

9214:
↑ and that is not as overstretched as it might sound. There are many tests in the codebase that rely on something that is bound to change (docstrings, argument names, error messages, etc) and that you end up fixing after introducing some change.

greggirwin:
First, can we all agree that #4119 can be closed? 

Second, Oy. :^) Test suites are mainly there for regression testing, but also act as a body of knowledge, and proof that things were tested at least once. Anything not tested is open to suspicion. Regression testing implies a commitment, a guarantee related to something we expect *not* to change, not just "how things are". Anybody who has seen documentation related to an application or web site that is out of date knows that some things *will* change. This is something to consider long term, when multiple versions of Red in the wild are all correct, but may have different rules, tests, and guarantees.

Even in the original ticket, it's subjective. Was the error message that bad? 

I say we check for error aspects that are important, rather than readability and specific text. If the text changes, it shouldn't break anything. If a user tests against it, that's their problem. If we have tests for it, they can rightly hold us accountable, and we have more work to do all around. Noting "Error *messages* may change." nudges people down the righteous path.

theSherwood:
Okay. I can change it to check against the `/id`, etc. of the errors.

9214:
@theSherwood the only type of error that you can get after an attempt to load a literal value is lexical; checking error's ID and type is redundant in such case. So I'd rather extend lexer tests:
```red
>> error! == transcode/scan "$non"
== true
>> error! == transcode/scan "\non"
== true
```

hiiamboris:
Old transcode was erroring as well:
```
>> system/lexer/transcode "non$" [] no
*** Syntax Error: invalid value at "$"
*** Where: do
*** Stack: run transcode  
```

9214:
@hiiamboris likely because `money!` wasn't implemented at the time.

theSherwood:
Okay. I removed the tests and added some simple lexer tests.

greggirwin:
:+1: Note that I recently posted some thoughts to the team on `all-word!` syntax, e.g. `/v:` syntax.

PR #4585, opened 2020-07-21
FEAT: Add Linux-musl-x86 target

vazub:
Support for building Red console with musl, instead of glibc. Tested with Ubuntu (x86) (having musl installed alongside glibc), but should work for any distro that keeps musl in typical FSH, e.g. under "/lib/ld-musl-i386.so.1"


rebolek:
@vazub I've tested it with Alpine Linux under LXD and I get an error:

```
Error relocating ./red: __fxstat: symbol not found
```

I also tried to run this under TurrisOS (OpenWRT based distro), but Red is experiencing some problems on ARM currently, so I will report if it works once Red works properly on ARM again.

vazub:
@rebolek can you provide the output of ldd when you use it on red binary?

rebolek:
@vazub 
```
alpine:~# ldd ./red
/lib/ld-musl-i386.so.1 (0xf7f04000)
libc.so.6 => /lib/ld-musl-i386.so.1 (0xf7f04000)
libm.so.6 => /lib/ld-musl-i386.so.1 (0xf7f04000)   
libpthread.so.0 => /lib/ld-musl-i386.so.1 (0xf7f04000)
libcurl.so.4 => /usr/lib/libcurl.so.4 (0xf7e7f000)
libnghttp2.so.14 => /usr/lib/libnghttp2.so.14 (0xf7e56000)
libssl.so.1.1 => /lib/libssl.so.1.1 (0xf7dd0000)
libcrypto.so.1.1 => /lib/libcrypto.so.1.1 (0xf7b6d000)
libbrotlidec.so.1 => /usr/lib/libbrotlidec.so.1 (0xf7b60000)
libz.so.1 => /lib/libz.so.1 (0xf7b46000)
libbrotlicommon.so.1 => /usr/lib/libbrotlicommon.so.1 (0xf7b23000)
Error relocating ./red: __fxstat: symbol not found
```

vazub:
@rebolek  I've verified console working on:
* Alpine (Docker, i386/alpine, 3.12.0, libcurl installed)
* Alpine (Hyper-V VM, i386/alpine, standard, 3.12.0, libcurl installed)

In both cases, lib versions in use were:
`libcurl-7.69.1-r0 x86`
`musl-1.1.24-r8 x86`

Alpine is currently the only major distro I could find that was built with musl in mind and still provides x86 images, therefore it is the main target for this improvement. Trying console compiled for musl in other distros wouldn't work without additional distro sorcery, as they were not intended for multilib as a generic scenario - there are lots of quirks with dynamic linker for each specific distro, with libcurl4 paths issues etc. In any case, I don't think this should be a problem, as other distros are served with generic Glibc Red compilation target.

The only caveat with the console so far is that `read %.` command doesn't work - probably something with underlying IO implementation in Red, that assumes Glibc. This is something that needs further analysis, but otherwise the target works.

9214:
https://github.com/red/REP/issues/27

vazub:
I've investigated and fixed the IO issue, but it probably makes sense to close this PR and open a new proper one.

Issue #4572, opened 2020-07-10
compiler does not process OP! aliases properly


9214:
**Describe the bug**
Even with a workaround for https://github.com/red/red/issues/2867 in place compiler still cannot handle `op!` aliases properly.

**To reproduce**
Compile the script below:
```red
Red []

set 'o get '//

probe 8 o 3 == 2
probe try [1 o 0]
```

Result:
```red
8
0
```

**Expected behavior**
```red
true 
make error! [...]
```

**Platform version**
 a1d1495, W10.

hiiamboris:
Compiler is it's own DSL ;)

9214:
[In my world, everyone's a pony...](https://www.youtube.com/watch?v=CJr4kPMVvpU)

dockimbel:
It is possible to improve the compiler to support an explicit form like `set 'o get '//`. However, it is easy to defeat it by indirectly referencing the `op!` value, like in:
```
new: first reduce [get '//]
set 'o :new
```
There is no way the compiler could guess that `o` will be set to an `op!` during static analysis in the general case.

As a reminder, only a subset of Red semantics are statically analyzable, therefore compilable. This set is quite large and covers most of the code written by users. Scripts using a more dynamic constructs should not be fed to the compiler, they should be evaluated only or compiled in encapped mode to fully preserve their semantics. That's the best we can achieve for now.

Issue #4571, opened 2020-07-10
crash of Rebol toolchain


9214:
**Describe the bug**
Among the numerous Heisenbugs and issues with Rebol2 tooclhain that I encountered in my lifetime, this is the one that I managed to consistently reproduce.

**To reproduce**
1. Open Rebol/View console session;
1. Attempt to [compile](https://github.com/red/red#running-red-from-the-sources-for-contributors) the script below using Red command-line front-end:
    ```red
    Red [] ///: :- 0 /// 0
    ```
1. Receive an error message:
    ```red
    *** Red Compiler Internal Error: Syntax Error : Invalid path -- ops/
    *** Where: none
    *** Near:  {^V^@^@^@V}
    ```
1. Retry step 2;
1. Observe the crash:
   <img src="https://user-images.githubusercontent.com/15716466/87178991-4d156f00-c2de-11ea-9755-d98671fe5364.png" width="500">

**Expected behavior**
Successful compilation. The reason for the original error message is probably compiler not being able to handle `///` syntax.

**Platform version**
 a1d1495, W10.
```red
>> rebol/version
== 2.7.8.3.1
```

dockimbel:
There is nothing we can do for now about it until R2 gets an update to fix that bug, or until we drop R2 dependency.

Issue #4570, opened 2020-07-10
compiler trips up on OP!


9214:
**Describe the bug**
Compiler thinks that `op!` call is a `function!` call and compiles an expression with missing left argument.

**To reproduce**
Compile the script below:
```red
Red [] quote + 0 0
```
The 2nd `0` is mandatory because the compiler insists that it's an argument to `+` "function call".

Result:
```red
*** Script Error: + does not allow word for its value1 argument
*** Where: +
*** Stack:
```

And that's the best-case scenario, because `+` is derived from `add` action that has a form of type-checking... unlike, say, `>>` derived `shift-right` routine:

```red
Red [] quote >> 0 0
```

```red
*** Runtime Error 98: assertion failed
*** in file: .../red/runtime/datatypes/integer.reds
*** at line: 66
***
***   stack: red/integer/get 00000062h
***   stack: red/integer/get 02BD32D4h
```

**Expected behavior**
Successful compilation; word set to `op!` is properly escaped by `quote`.

**Platform version**
a1d1495, W10.

dockimbel:
@9214 I let you write the regression test(s).

Issue #4569, opened 2020-07-10
compiler ignores setting a word to an object


9214:
**Describe the bug**
If one instance of a word is bound to function's context, and then another instance with the same spelling is set to an object — magic happens.

**To reproduce**
The script below, once compiled:
```red
Red []

bind 'foo has [foo]['WTF]
foo: object []

probe foo
probe :foo

```
Results in:
```red
WTF
make object! []
```

**Expected behavior**
```red
make object! []
make object! []
```

**Platform version**
a1d1495, W10.

greggirwin:
Good example. :+1:

9214:
Another one:
```red
Red []

block: reduce ['foo func [/bar]["Definitely not bar."]]
foo:  context [bar: does ['bar]]
probe foo/bar
```
When compiled:
```red
"Definitely not bar."
```

dockimbel:
@9214 In your second example, the following line:
```
block: reduce ['foo func [/bar]["Definitely not bar."]]
```
has no effect on the rest of the code.

9214:
@dockimbel if what you say is true, then the "Definitely not bar." message appears out of nowhere 😉 I expect the compiled code to call `bar` within object `foo`, instead the anonymous function within `block` is called, as if referenced by key `foo`.

PR #4586, opened 2020-07-22
FEAT: Redbin codec.

### LINKED ISSUE #4431: [Compiler] Redbin encoder does not preserve new-line flags in ANY-STRING! values
**Describe the bug**
`any-string!` values lose their new-line flags when encoded into Redbin payload during compilation.

**To reproduce**
Compile the code snippet below:

```red
Red []

probe [
    a
    :b c:
    'd /e #f
	
    "a"
    @b %c
    <d> e@f
]
```

Output:
```red
[
    a
    :b c:
    'd /e #f "a" @b %c <d> e@f
]
```

**Expected behavior**
```red
[
    a
    :b c:
    'd /e #f
    "a"
    @b %c
    <d> e@f
]
```

**Platform version**
```
Red 0.6.4 for Windows built 7-May-2020/6:07:31+02:00 commit #c149a36
```

9214:
This PR introduces a runtime codec for refined and extended Redbin format, and accompanies it with a test suite; for details, see the relevant [documentation](https://github.com/red/docs/blob/master/en/redbin.adoc).

Compiler's encoder written in Rebol is left unmodified — as such, there are now two supported versions of Redbin format incompatible with each other. Because of this incompatibility, and because the new format introduced many new features, I took the liberty to bump its version to 2. Said features include:

* Added support for (almost) all Red datatypes, extended support for existing ones;
* Properly implemented support for cycles and references for indirect datatypes;
* Proper encoding of binding information for `any-word!` values.

### Known limitations
Described [here](https://github.com/red/docs/blob/master/en/redbin.adoc#unsupported).

### Pending bugs
Issues that affected the codec implementation and tests one way or another: #4572, #4571, #4570, #4569, #4568, #4567, #4565, #4563, #4562, #4552, #4540, #4526, #2867, #2207, #734.

### Open design questions
* How to handle `object!`'s class IDs: currently, they are encoded and used when instantiating a decoded object, but this might lead to conflicts in objects' inheritance system;
* How to handle values to which words bound to a global context are set: right now they are not encoded, because on decoding it might cause unwanted side-effects;
* Sandboxing and Secure dialect: right now feeding decoder with malformed payload will likely lead to a [big bada boom](https://www.youtube.com/watch?v=j8WLYzA0lCs), since it implements only naive poka-yoke checks of the input; graceful handling of incorrect encoding • will degrade performance on runtime booting • is infeasible in the general case.  
* Construction syntax and how this work on Redbin format can inform its design and implementation.

---

Fixes #4431 (string decoder didn't set the new-line flag properly). Regression test is implicitly contained in "Header flags" test group inside Redbin codec test suite.


dander:
The documentation links (including "known limitations") seem to be broken.

9214:
@dander, yes, I deleted my local branch since it has been merged into [master](https://github.com/red/docs/blob/master/en/redbin.adoc). I've updated the links.

Issue #2134, opened 2016-07-26
funny time! values


WiseGenius:
1187 full minutes <= `24:00` end in a `.0`:

```
red>> 00:07:59
== 0:07:59
red>> 00:08:01
== 0:08:01
red>> 00:08:00
== 0:08:00.0
```

254 full minutes < `24:00` end in `:60`. The rest are as above.

```
red>> 00:08:59
== 0:08:59
red>> 00:09:01
== 0:09:01
red>> 00:09:00
== 0:08:60
```

Every single full minute can be made to end in `:60` in the following way:

```
red>> 00:00:60              ;OK
== 0:01:00.0
red>> 00:00:00 + 60         ;OK
== 0:01:00.0
red>> (00:00:00 + 1) * 60   ;Even this is OK
== 0:01:00.0
red>> 00:00:01 * 60         ;Oops!
== 0:00:60
```

Every single full minute behaves in the following horrible way:

```
red>> t: 0:00:00 loop 59 [t: t + 1]
== 0:00:59
red>> t: 0:00:00 loop 61 [t: t + 1]
== 0:01:01
red>> t: 0:00:00 loop 60 [t: t + 1]
== 0:01:05.68434189e-14
```

There is also this behaviour:

```
red>> t: now/time/precise/utc
== 15:13:59.984
red>> t/second
== 59.98399999999674
red>> t: now/time/precise/utc
== 15:14:52.156
red>> t/second
== 52.15600000000268
```

qtxie:
Partially fix some cases. 

The last one `t/second` is not fixed.

dockimbel:
@qtxie Could we fix the last case now?

qtxie:
@dockimbel Nope. As we return float! in the last case and we use `%.16g` in `sprintf` function when printing a float.

hiiamboris:
How is the last case even wrong? `t/second` is a float, and it is displayed with the maximum available precision, what's so bad about it? Those who need a pretty fixed format it can always do so with
```
>> x: to-integer t/second + 1e-7 * 1e6  rejoin [x / 1'000'000 "." pad/left/with x % 1'000'000 6 #"0"]
== "59.984000"
```

greggirwin:
It created a non-loadable value.
```
>> t: 0:00:01 / 1000000
== 0:00:1e-6
>> 0:00:1e-6
*** Script Error: e-6 has no value
*** Where: catch
*** Stack:  

>> load mold t
== [0:00:01 e-6]
```

[...]

PR #4589, opened 2020-07-25
TESTS: issue #2134 (funny time! values)

theSherwood:


theSherwood:
This pull request does NOT contain tests for the maximum precision of `t/second`, as the opinion in #2134 seemed to be that the current behavior is expected.

Issue #4526, opened 2020-06-17
[Compiler] Internal error `defer needs a value`


hiiamboris:
**Describe the bug**

Compile this:
```
Red []                                 
c: context [++: make op! func [x y] []]
```
Compiler says:
```
-=== Red Compiler 0.6.4 ===-

Compiling D:\devel\red\red-src\red-master\environment\reactivity-bug2.red ...
*** Red Compiler Internal Error: Script Error : defer needs a value
*** Where: comp-word
*** Near:  [defer: fetch-functions skip pc -2]
```

**Expected behavior**

Compiles.

**Platform version (please complete the following information)**
```
Red 0.6.4 for Windows built 28-May-2020/17:51:04+03:00 commit #36857eb
```

9214:
The same error blocks me from compiling a 750LOC test suite.

9214:
I can no longer reproduce the original error message. @hiiamboris?

hiiamboris:
Bad if issues go away by themselves. Means they hide deeper.
Anyway:
```
Red []                                 
c: context [**: make op! func [x y] []]
```
*** Red Compiler Internal Error: Script Error : defer needs a value
*** Where: comp-word
*** Near:  [defer: fetch-functions skip pc -2]
```

PR #4593, opened 2020-07-31
FIX: issue #4563 (OP! constructed from OP! prototype is malformed). 

### LINKED ISSUE #4563: OP! constructed from OP! prototype is malformed
**Describe the bug**
`op!`'s `make` action accepts other `any-function!` values as its prototype. In some cases, `op!` created from another `op!` is malformed.

**To reproduce**
```red
shr: make op! :>>
2 shr 1
```

```red
*** Cannot determine source file/line info.
***
```

Let's engage in a [whodunit](https://en.wikipedia.org/wiki/Whodunit) narrative and see why this happens. As I said earlier, `op!` value can be constructed from any other `any-function!` value. However, `op!` parents differ in their structure:

* `native!` and `action!` do not contain body blocks but rely on native pointers.
https://github.com/red/red/blob/b45271621e78d6f24e4c0e4d6f1f0acc6cf61640/runtime/datatypes/structures.reds#L239
* `function!` and `routine!`, on the other hand, contain body block, and `routine!` can come from FFI.
https://github.com/red/red/blob/b45271621e78d6f24e4c0e4d6f1f0acc6cf61640/runtime/datatypes/structures.reds#L253

To differentiate two classes of parents in this taxonomy, `op!` header contains two bit flags:

https://github.com/red/red/blob/b45271621e78d6f24e4c0e4d6f1f0acc6cf61640/runtime/allocator.reds#L31

https://github.com/red/red/blob/b45271621e78d6f24e4c0e4d6f1f0acc6cf61640/runtime/allocator.reds#L27

Here are examples of `op!` values from each category:

| `16` | `23` | `op!` | Parent |
|:-:|:-:|:-:|:--|
| 0 | 0 | `+` | `add`, `action!`
| 0 | 1 | `>>` | `shift-right`, `routine!`
| 1 | 0 | `>` | `greater?`, `native!`
| 1 | 1 | x | x |

Now, this differentiation is needed because parents are differently structured; one class of parents contains a direct pointer, and another contains an indirect pointer. The crash happens when we create `op!` in one category (which is assumed to be the default by `make`) from `op!` in another category.

https://github.com/red/red/blob/b45271621e78d6f24e4c0e4d6f1f0acc6cf61640/runtime/datatypes/op.reds#L107

https://github.com/red/red/blob/b45271621e78d6f24e4c0e4d6f1f0acc6cf61640/runtime/datatypes/op.reds#L110

To reiterate on the original example: `>>` is an `op!` made from routine, and its body is a node pointer. `make op! :>>` ignores this fact and treats node pointer as pointer to native code, creating a ticking time bomb `shr`.

**Expected behavior**
I expect `op!` derived from another `op!` to be fully operational; to do so it suffices to add an extra check of bit flags inside `make`.

**Platform version**
 b452716

9214:
Fixes #4563 by forbidding the creation of `op!` value from `op!` prototype; the rationale for this change is described in https://github.com/red/red/issues/4563#issuecomment-652462593. A regression test is provided.

Also, it appears that `flag-native-op` is set even if `op!` is derived from `action!`. This flag is checked only by the interpreter and enables type checking for natives by pushing logical flag on a stack, so I kept it as-is (it works fine so why break it?).
https://github.com/red/red/blob/b345eb57c6be831df43f5a380416d10e95e0cb06/runtime/interpreter.reds#L415


9214:
Compiled tests are failing because of https://github.com/red/red/issues/4526.
```red
Compiling ...\red\tests\source\units\auto-tests\run-all-comp1.red ...
...using libRedRT built on 31-Jul-2020/14:41:19
*** Red Compiler Internal Error: Script Error : defer needs a value 
*** Where: comp-word 
*** Near:  [defer: fetch-functions skip pc -2] 
```
I'll apply the workaround for the time being.

PR #4594, opened 2020-07-31
FIX: issue #4537 (/LOCAL refinement injected by function constructors is malformed).

### LINKED ISSUE #4537: /LOCAL refinement injected by function constructors is malformed
**Describe the bug**
`has` and `function` inject `/local` refinement into constructed function's spec. Such refinement is an `any-word!` value, and, as such, must be bound to a context, which turns out not to be true on practice.

**To reproduce**
Any of the examples below:
```red
get first spec-of has [foo][]
get first spec-of function [][foo:]
```
Results in:
```red
*** in file: .../red/runtime/datatypes/context.reds
*** at line: 325
***
***   stack: red/_context/get 02C8A414h
***   stack: red/natives/get* false -1 -1
***   stack: red/interpreter/eval-arguments 02C8A404h 031FFCC8h 031FFCC8h 00000000h 00000000h
***   stack: red/interpreter/eval-code 02C8A404h 031FFC98h 031FFCC8h false 00000000h 00000000h 02E50C38h
***   stack: red/interpreter/eval-expression 031FFC98h 031FFCC8h false false false
***   stack: red/interpreter/eval 02C8A3F4h true
***   stack: red/natives/try* true 0
***   stack: do-file
***   stack: red/natives/do* false -1 -1 -1
***   stack: red/interpreter/eval-arguments 02C8A334h 031FF298h 031FF298h 00000000h 00000000h
***   stack: red/interpreter/eval-code 02C8A334h 031FF288h 031FF298h false 00000000h 00000000h 02E50BA8h
***   stack: red/interpreter/eval-expression 031FF288h 031FF298h false false false
***   stack: red/interpreter/eval 02C8A314h true
***   stack: ctx||564~try-do 026E1D2Ch
***   stack: ctx||564~do-command 026E1D2Ch
***   stack: ctx||564~eval-command 026E1D2Ch
***   stack: ctx||564~run 026E1D2Ch
***   stack: ctx||564~launch 026E1D2Ch
***   stack: ctx||598~launch 026E14B4h
```

Because context node is `null`.

**Expected behavior**
I expect function's refinements (including `/local`) to be bound to either global context or function's context.

**Platform version**
4d864b1d

9214:
Fixes #4537 and provides a relevant regression test.

Runtime has a set of refinements (including `/local` that gets injected by `has` and `function`) that is created internally during the booting process:

https://github.com/red/red/blob/b345eb57c6be831df43f5a380416d10e95e0cb06/runtime/datatypes/common.reds#L986  

https://github.com/red/red/blob/b345eb57c6be831df43f5a380416d10e95e0cb06/runtime/red.reds#L227

The creation of these refinements relies on `refinement/load`, which stores `refinement!` slot in root block, and then explicitly initializes its `ctx` field to `null`, which is the cause of the reported problem (on top of that it also skips `index` field).

https://github.com/red/red/blob/b345eb57c6be831df43f5a380416d10e95e0cb06/runtime/datatypes/refinement.reds#L40

To fix that, I reuse `word/load` and convert resulting slot to refinement by changing type tag in the header: internally, `word!` and `refinement!` have the same structure, and all the struct fields (including `ctx`) are properly initialized that way.

The question of whether or not `get` should accept `refinement!` values can be addressed in a separate ticket if desired — my gut tells me that `issue!` and `refinement!` are pure literal symbols, in a sense that they cannot be explicitly re-bound or set to a value by the user; prohibiting `get` on `refinement!` will also make it symmetrical with `set`.


dockimbel:
Issue #4536 has been resolved, so that this PR is not needed anymore. Moreover, the `ctx` slot for refinements can stay at `null` now that refinements can't get dereferenced anymore by `get`.

9214:
@dockimbel I beg to differ: this does not counter the cases where a malformed refinement is converted to, say, `word!` (see the provided tests).
```red
get to word! first spec-of has [foo][]
```
This PR should be reopened.

PR #4596, opened 2020-08-01
FEAT: remove `matrix-order` command in DRAW.

qtxie:
This feature is GDI+ specific, rarely used and not easy to be implemented on other platforms.

### DIFF
--- a/modules/view/draw.red
+++ b/modules/view/draw.red
@@ -30,7 +30,6 @@ Red/System [
 		line-join:		symbol/make "line-join"
 		line-cap:		symbol/make "line-cap"
 		matrix:			symbol/make "matrix"
-		_matrix-order:  symbol/make "matrix-order"
 		_append:        symbol/make "append"
 		prepend:        symbol/make "prepend"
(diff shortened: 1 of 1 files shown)


qtxie:
Changes in docs: https://github.com/red/docs/pull/240

9214:
https://github.com/red/red/pull/4450

dockimbel:
@qtxie Seems it is implemented for macOS by PR #4450. Is it possible to support it on GTK also somehow?

qtxie:
@qtxie Yes. It can be implemented. The reasons to remove it:
* It's based on a GDI+ specific feature, I don't see other graphic frameworks provide it. 
* It's not a necessary feature, almost nobody use it.
* Reduce the performance of all the drawing commands on non-Windows platforms. (Check the implementation on macOS, every draw commands need a `ctx-matrix-adapt` and `ctx-matrix-unadapt` in order to support it)

dockimbel:
@qtxie Ok, so we'll drop it.

PR #4605, opened 2020-08-05
FEAT: Add working Linux-musl target

vazub:
Support for building Red console on Linux with musl, instead of glibc. Also found and fixed previously reported IO issue - so listing dirs should work as intended now. Tested on fresh Alpine 3.12, x86,  with libcurl installed (both as a Docker image and VM install). 

P.S. The diff shows lots of changes, but this seems to be whitespace difference (LF vs CRLF) for the most part, hope it is not a blocker.


9214:
Mentioning https://github.com/red/REP/issues/27 here again since this PR addresses the issue described in it.

dockimbel:
@vazub Good job! :+1:

Issue #4575, opened 2020-07-16
Add (modified if string) to doc strings for `uppercase` and `lowercase`


greggirwin:
These funcs modify their arg if it's a string, but don't note it in their doc strings, which is helpful.

9214:
Resolved by https://github.com/red/red/pull/4606 and can be closed.

PR #4606, opened 2020-08-06
FIX: issue #4575 - (Add (modified if string)...)

theSherwood:


9214:
FYI, `DOCS` label is [reserved](https://github.com/red/red/wiki/%5BDOC%5D-Contributor-Guidelines#general-contribution-guidelines) for issues directly related to `red/docs` documentation. If PR addresses a ticket, then `FIX: issue #<number> (<description>)` format should be used.

For the fix itself, I propose to use "modified when series" instead, since we have at least one precedent with it:
https://github.com/red/red/blob/696ae437550711ba8df46dc611ae12e9f156b0f0/environment/actions.red#L26

OTOH, some functions note that in their top-level docstring:
https://github.com/red/red/blob/696ae437550711ba8df46dc611ae12e9f156b0f0/environment/functions.red#L211-L212

And others in return docstring:
https://github.com/red/red/blob/696ae437550711ba8df46dc611ae12e9f156b0f0/environment/functions.red#L551

Consistency is surely lacking.

theSherwood:
That's good to know. I figured "Docs" would make sense on account of the issue being tagged with `type.documentation`. But that makes sense. Shall we close this and I can open another pull request that is correctly named?

9214:
@theSherwood no need! It's just me being overly pedantic 😸

theSherwood:
Okay. I'll be sure to make that change to the docstring when I next get the chance.

PR #4615, opened 2020-08-10
FIX: issue #4567 (mysterious "argument type mismatch" compiler error). 

### LINKED ISSUE #4567: mysterious "argument type mismatch" compiler error
**Describe the bug**
A "what the actual fbugck?!" award nominee.

**To reproduce**
Compile the script below:
```red
Red [] objects/1
```

```red
Compiling to native code...
*** Compilation Error: argument type mismatch on calling: red/eval-int-path*

*** expected: [struct! [
        header [integer!]
        data1 [integer!]
        data2 [integer!]
        data3 [integer!]
    ]], found: [integer!]
*** in file: %utils/preprocessor.r
*** at line: 1
*** near: [1
    stack/unwind #user-code
```

Now change `objects` to some other word _and it will compile_.

**Expected behavior**
Ideally and in all cases: an "undefined symbol" error. Else a successful compilation with a subsequent run-time error.

**Platform version**
4e9d770, W10.

9214:
Fixes #4567 and provides a regression test.

To get face-to-face with the issue described in the ticket, let's pipe out the output of `--red-only` and meditate on the code emitted by the compiler:

```red
stack/mark-native ~eval-path 
eval-int-path* 1 
```

But `eval-int-path*` expects two arguments, a `red-value!` and an `integer!`, so one would rather expect to see:

```red
stack/mark-native ~eval-path 
eval-int-path* _context/get ~objects 1 
```

By its nature, the issue looks like a name collision: only `objects` gets removed, all the other words behave as expected. And sure enough, an inspection of `compiler.r` provides the first clue: 

https://github.com/red/red/blob/ab1a8ebddfac721b8687754944070f6afdeb23a1/compiler.r#L31

This stack forms a path to [shadow object](https://www.red-lang.org/2014/12/objects-implementation-notes.html) created by the compiler. Note the subtle detail: `obj-stack` _always_ contains `objects` word inside it as a root of the path. This `objects` is a prefix that compiler inserts at the head of every path that looks like an access to an object.

https://github.com/red/red/blob/ab1a8ebddfac721b8687754944070f6afdeb23a1/compiler.r#L990-L1001

Surely if compiler inserts it then it should also remove it at some point, otherwise `path!` literal will get malformed. Indeed, here's what happens in `emit-path`:

https://github.com/red/red/blob/ab1a8ebddfac721b8687754944070f6afdeb23a1/compiler.r#L1464

The `object-access?` call in question happens in `comp-path` just before that:

https://github.com/red/red/blob/ab1a8ebddfac721b8687754944070f6afdeb23a1/compiler.r#L3355-L3359

It's clear as day that `objects/1` is not an `object!` access, because `objects` is `unset!`. `obj?` in such case is set to `none`.

But the "prefix" is still removed: `objects/1` suddenly turns into `1`! Why? Because `obj-stack` is `objects` path, and `path/1 = last obj-stack` returns `true`.
```red
>> obj-stack: to path! 'objects
== objects
>> path: 'objects/1
== objects/1
>> path/1 = last obj-stack
== true
```
So, to reiterate: compiler thinks that path is an object access and removes an internal prefix that it supposedly tucked at the head of the path before, thus malforming a `path!` literal, while in fact there's no object (and hence no prefix) at all and `path!` should remain unmodified.

The fix, then, is trivial: ensure that stack is not empty before checking the prefix.


9214:
Hmph, tests [went bonkers](https://ci.appveyor.com/project/red/red/builds/34575553) on AppVeyor but pass locally.
```
Quick-Test v 0.12.0
REBOL  2.7.8.3.1
This test started at 10-Aug-2020/16:17:23.047+2:00
Complete Red Test Suite

ok - preprocessor tests..............................39 / 39
ok - run-all-comp1.red.............................7914 / 7914
ok - run-all-comp2.red.............................5246 / 5246
ok - run-all-interp.red...........................13016 / 13016
ok - Red print.......................................13 / 13
ok - Red run time errors..............................9 / 9
ok - Red compile errors...............................7 / 7
ok - base-self......................................493 / 493
ok - alias-compile....................................3 / 3
ok - cast-compile....................................28 / 28
ok - comp-err.........................................2 / 2
ok - exit-err.........................................3 / 3
ok - int-literals-err.................................1 / 1
ok - output...........................................3 / 3
ok - return-err.......................................5 / 5
ok - conditions-required-err.........................23 / 23
ok - inference-err....................................1 / 1
ok - callback-err.....................................2 / 2
ok - infix-compile....................................3 / 3
ok - not-compile......................................1 / 1
ok - print............................................3 / 3
ok - enumerations compile............................12 / 12
ok - pointer-compile..................................2 / 2
ok - namespace compiler tests.........................9 / 9
ok - compiles-ok......................................1 / 1
ok - dylib compiler...................................2 / 2
ok - define...........................................1 / 1
ok - Red/System Compiler Regression tests...........107 / 107
ok - array............................................2 / 2
ok - logic..........................................105 / 105
ok - byte..........................................3677 / 3677
ok - c-string........................................20 / 20
ok - struct.........................................571 / 571
ok - pointer.........................................75 / 75
ok - cast...........................................144 / 144
ok - alias...........................................23 / 23
ok - length..........................................11 / 11
ok - null............................................11 / 11
ok - enum............................................19 / 19
ok - float.........................................1683 / 1683
ok - float32.......................................1085 / 1085
ok - lib.............................................15 / 15
ok - get-pointer.....................................28 / 28
ok - float-pointer...................................61 / 61
ok - namespace......................................112 / 112
ok - not.............................................44 / 44
ok - size............................................38 / 38
ok - function.........................................6 / 6
ok - case............................................64 / 64
ok - switch..........................................85 / 85
ok - exit.............................................9 / 9
ok - return..........................................25 / 25
ok - exceptions......................................11 / 11
ok - modulo..........................................29 / 29
ok - math-mixed....................................1775 / 1775
ok - infix............................................2 / 2
ok - conditional.....................................15 / 15
ok - system..........................................16 / 16
ok - atomic...........................................8 / 8
ok - queue...........................................64 / 64
ok - dylib-auto.......................................7 / 7

ok - Complete Red Test Suite......................36921 / 36921
       in 0:10:00.413

The test finished at 10-Aug-2020/16:27:23.46+2:00
```

9214:
Jeez... that was the difference between Rebol2 versions. Local `2.7.8.3.1` has `last?` and `2.7.6.3.1` plonked on the server does not.

hiiamboris:
> 
> 
> Jeez... that was the difference between Rebol2 versions. Local `2.7.8.3.1` has `last?` and `2.7.6.3.1` plonked on the server does not.

Yeah, I found that out too the other day :D

PR #4616, opened 2020-08-10
FIX: issue #4204 (FIND/LAST does not honor /PART). 

### LINKED ISSUE #4204: FIND/LAST does not honor /PART
**Describe the bug**

```
>> find/part/last "a b c d e f g" " " 3
== " g"
```

**Expected behavior**

R2 (correct):
```
>> find/part/last "a b c d e f g" " " 3
== " b c d e f g"
```

```
Red 0.6.4 for Windows built 20-Dec-2019/19:03:46+03:00 commit #544a6e1
```

9214:
Fixes #4204 and extends `series!` test suite.

With `find/last` the search happens backward, from end to start, but in case of `last` combined with `part` the upper limit was incorrectly calculated: it always happened from the series end rather than from the boundary delineated by `/part` argument.

What's more important is that the other limit wasn't correctly calculated either:

https://github.com/red/red/blob/20aded310d83709cb22294a4861c313495c80333/runtime/datatypes/string.reds#L1908-L1912

Note the `buffer - limit` part. This means that the lower boundary resides outside the series buffer, somewhere in `series!` meta info or even before it. Knowing that this meta info is 20 bytes in width and likely contains `01` byte in the header (a bit flag that specifies that series is used), we can, then, using Latin1 string:
```red
>> find/last/part "Cthulhu fhtagn!" #"^(01)" 20 + 1
== "^A^@^@ ^XÅm^B^P^@^@^@ðÈ^W^CÿÈ^W^CCthulhu fhtagn!"
```

Conceptually, one boundary is always fixed at the index of the series, and the other is either located at the tail or shifted by `/part` to a specific position. Provided fix implements this logic.

### DIFF
--- a/runtime/datatypes/string.reds
+++ b/runtime/datatypes/string.reds
@@ -1887,13 +1887,9 @@ string: context [
 			if step < 1 [fire [TO_ERROR(script out-of-range) skip]]
 		]
 		if OPTION?(part) [
-			limit: either TYPE_OF(part) = TYPE_INTEGER [
+			sz: either TYPE_OF(part) = TYPE_INTEGER [
 				int: as red-integer! part
-				if int/value <= 0 [						;-- early exit if part <= 0
-					result/header: TYPE_NONE
-					return result
-				]
-				int/value << (unit >> 1)
+				int/value
 			][
 				str2: as red-string! part
 				unless all [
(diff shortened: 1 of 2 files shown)


9214:
As usual, there's some Heisenbug with the tests:
```red
probe none? probe find/last/part "abc" #"^(01)" 20 + 1
```
returns expected `true none` when compiled and interpreted alone, but yields `false ""` when compiled in a series test suite. Looks like stack corruption. Will look into it tomorrow.

9214:
Nope, that was `find` not putting a hard cap on `/part` argument. In the case above it simply overshoots over series tail: when compiled alone, `#"^(01)"` is not found because there's no series allocated past `"abc"`, but inside series suite it likely finds it somewhere nearby.

PR #4634, opened 2020-08-20
FIX: issue #4627 (reported values in error messages are not properly displayed).

### LINKED ISSUE #4627: reported values in error messages are not properly displayed
**Describe the bug**
Script errors typically mention Red values in their message. At some point, [regression](https://github.com/red/red/commit/fdf7a3ace7a623eabd214712d0a1d9b2a847bcae#diff-ee5dd4447fbcfc9c5056cbcd379763b4L149) made said messages less informative.

**To reproduce**
```red
>> null < []
*** Script Error: cannot compare   with 
*** Where: <
*** Stack:  
```

`null` character literally follows after "compare", as if it was printed rather than `mold`ed. Same with `[]`.
```red
>> print [null []]
 ; here they are
```

**Expected behavior**
```red
>> null < []
*** Script Error: cannot compare #"^@" with []
*** Where: <
*** Stack:  
```

**Platform version**
0084083, W10.

9214:
Fixes #4627 and provides a regression test. The issue was caused by a commit fdf7a3a introduced during the work on the new lexer. It slightly changed error processing to make syntax errors more readable:

https://github.com/red/red/blob/151e6088efec77cf87ed024e7fb5660f55fa3a0b/runtime/datatypes/error.reds#L149

That is, reported values are now `form`ed rather than `mold`ed. This makes sense, because with `mold` syntax errors would look like this:

```red
>> ]
*** Syntax Error: "(line 1)" missing #"[" at "]"
*** Where: transcode
*** Stack: load 
```

OTOH, once values are `form`ed, the other types of error messages become pretty obscure and uninformative:
```red
>> 1 < [2]
*** Script Error: cannot compare 1 with 2 ; Of course you can Red, believe in yourself!
*** Where: <
*** Stack: 
>> null > []
*** Script Error: cannot compare  with ; Wat?
*** Where: >
*** Stack:  
```

To keep the best of both worlds, I haven't come up with anything smarter than dispatching on the type of error: if it's a syntax error then we `form`, else we `mold`. Some syntax errors became a little bit too quirky because of that:
```red
>> do %scratch.txt
*** Syntax Error: script is missing a Red header: scratch.txt ; Instead of %scratch.txt.
*** Where: do
*** Stack: do-file cause-error  
```
But personally I can live with that.

### DIFF
--- a/runtime/datatypes/error.reds
+++ b/runtime/datatypes/error.reds
@@ -130,14 +130,16 @@ error: context [
 		obj		[red-object!]
 		return: [red-block!]
 		/local
-			value  [red-value!]
-			tail   [red-value!]
-			buffer [red-string!]
-			type   [integer!]
+			value   [red-value!]
+			tail    [red-value!]
+			buffer  [red-string!]
+			type    [integer!]
+			syntax? [logic!]
 	][
 		value: block/rs-head blk
 		tail:  block/rs-tail blk
 		
+		syntax?: words/errors/syntax/symbol = get-type obj
 		while [value < tail][
 			type: TYPE_OF(value)
 			if any [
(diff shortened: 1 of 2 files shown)


dockimbel:
Your trade-off seems better than the current one, so I'll merge it.

PR #4644, opened 2020-09-04
FEAT: add enhex function

bitbegin:


bitbegin:
@dockimbel still need lexer support:

in `lexer.reds`, string!/file!/url! all use load-string, and set header at end.  this `load-string` is ok for string, it create a origin string!. However, we need `string/decode` the loaded string for url! and `%a/b` format file!. file! format like `%"a/b"` no need `string/decode`.  

for compiler part:

1. same issue like above, need create right literal for file! and url!
2. i changed `file/load url/load url/load-in`, and don't know how to adapt for compiler

bitbegin:
new rules:
1. url! literal like `https://www.red-lang.org/中%20文` will be decoded to `"https://www.red-lang.org/中 文"` (also stored this string!); `mold` url! will print percent-encode string for `load`, `print` url! will print origin string (stored string)

2. file! literal have two type. one is percent-encode (not include `""`) file!, for example `%/d/folder%201/%E4%B8%AD%E6%96%87`, so it will be decoded first. other one is line-string file!, for example `"d:/folder%201/中文"`, it will not be decoded, so it still stored this string; `mold` and `print` file! both will print origin string (not percent-encode)

dockimbel:
@bitbegin 
> url! literal like https://www.red-lang.org/中%20文 will be decoded to "https://www.red-lang.org/中 文"

You cannot decode a `url!` into a `string!` type, that goes against the language core principles.

bitbegin:
@dockimbel  yes, the string `"https://www.red-lang.org/中 文"` will be stored.

bitbegin:
@dockimbel i also have another idea, to store url! as it is from, for example `load %%{https://www.red-lang.org/中%20文}%%` will store `https://www.red-lang.org/中%20文`. when print it, it show `https://www.red-lang.org/中 文`; when mold it, it show `https://www.red-lang.org/中%20文`, then introduce `encode-url` function, use it like this `str: encode-url url`, this will get string! `https://www.red-lang.org/%E4%B8%AD%20%E6%96%87`.

the file! rule not change, as i described above.

qtxie:
@bitbegin 
> other one is line-string file!, for example "d:/folder%201/中文", it will not be decoded

You mean this literal format: `%"d:/folder%201/中文"` ? `"d:/folder%201/中文"` is a literal string.

I propose the following changes:

1. add enhex function, which follows the [RFC 3986 standard](https://tools.ietf.org/html/rfc3986). (we have it in this PR)
2. update dehex function, make it follows the [RFC 3986 standard](https://tools.ietf.org/html/rfc3986).

The rules for loading file!:

* For literal form 1: `%abc/efg/xyz`, decodes it with `dehex` function when loading.
* For literal form 2: `%"abc/efg/xyz"`, no decoding, loads it as it is.
* No encode when MOLD and FORM.
E.g.
```
>> %/d/folder/%E4%B8%AD%E6%96%87.txt
== %"/d/folder/中文.txt“

>> %"/d/folder/%E4%B8%AD%E6%96%87.txt"     ;@@ %E4%B8%AD%E6%96%87.txt is a valid filename
== %"/d/folder/%E4%B8%AD%E6%96%87.txt"

>> mold %"/d/folder/中文.txt“
== {%"/d/folder/中文.txt“}

>> form %"/d/folder/中文.txt“
== "/d/folder/中文.txt“
```

The rules for loading url!:
Keep it the same behavior as in master branch, just use the updated `dehex` and `enhex` function to decode and encode it.
1. print the `mold` version
2. `form url!` gets a user friendly string.

E.g.
```
>> https://www.red-lang.org/中%20文
== https://www.red-lang.org/%E4%B8%AD%20%E6%96%87
>> mold https://www.red-lang.org/中%20文
== "https://www.red-lang.org/%E4%B8%AD%20%E6%96%87"
>> form https://www.red-lang.org/中%20文
== "https://www.red-lang.org/中 文"
```

The rules for MAKE/TO url! and file!:
* take it as it is, no decoding.
```
>> to file! "%/d/folder/%E4%B8%AD%E6%96%87.txt"
== %"%/d/folder/%E4%B8%AD%E6%96%87.txt"

>> to url! "http://www.df%20fjd"
== http://www.df%2520fjd
```

bitbegin:
@qtxie seems this PR satisfy these rules.

runtime/datatypes/string.reds
```diff
@@ -136,61 +175,206 @@ string: context [
...
+			ch2		[byte!]
+			utf8?	[logic!]
+	][
+		slen: -1
+		data: as byte-ptr! unicode/to-utf8 str :slen
+		if slen = 0 [									;-- empty string case
+			return null
+		]
+		end: data + slen
+		buffer: allocate slen
```
qtxie, line 226: Better to allocate the buffer in a red-string! directly, then use `GET_BUFFER(buf)` to get a pointer to the buffer. In this way we avoid allocating twice.

qtxie (commented):
Better to get rid of `allocate` and `free`. Write directly to the destination buffer.

runtime/datatypes/string.reds
bitbegin: fixed

hiiamboris:
> ```
> >> https://www.red-lang.org/中%20文
> == https://www.red-lang.org/%E4%B8%AD%20%E6%96%87
> 
> >> mold https://www.red-lang.org/中%20文
> == "https://www.red-lang.org/%E4%B8%AD%20%E6%96%87"
> 
> >> form https://www.red-lang.org/%E4%B8%AD%20%E6%96%87
> == "https://www.red-lang.org/中 文"
> ```

I don't know what RFCs say, just curious: no way to get the original `https://www.red-lang.org/中%20文`? back ;)
I have a habit of making URLs look like this before I post them to someone (so he can read the url before clicking it)

qtxie:
@hiiamboris 
`https://www.red-lang.org/中%20文`, This is a malformed URL, it's mixed decoded characters (中 and 文) with encoded character （a whitespace). Red will decoded it correctly like the browsers and give you the right encoded URL: `"https://www.red-lang.org/中 文"`.

hiiamboris:
@qtxie I insert a space at the head of URL bar, then copy the rest after the space (on Firefox-based browsers). In any case, I think all browsers accept this non standard form.

dockimbel:
@qtxie I agree with your proposal, just not sure what you meant there:

> print the `mold` version

`print` is using `form` serialization, to make it more human-readable.

dockimbel:
@qtxie @greggirwin Are you OK for merging this PR?

qtxie:
> @qtxie I agree with your proposal, just not sure what you meant there:
> 
> > print the `mold` version
> 
> `print` is using `form` serialization, to make it more human-readable.

I mean the return values show in the console.
```
>> https://a%20b
== https://a%20b       ;@@ mold version
```

qtxie:
> @qtxie @greggirwin Are you OK for merging this PR?

I'll review the code first.

qtxie (approved):

Issue #4576, opened 2020-07-16
text-list data change, advanced use case, may lead to crash sometimes.


greggirwin:
**Describe the bug**

See comments in code below

**To reproduce**

```
Red []

; This came from using key-val data (originally a map converted to a block)
; as `data` for a list face, so when you select an item (key), the associated
; value was available. 

; What happened is that sometimes the list entries would change, with one
; being removed. TL;DR is that is seems a combination of any loop construct
; combined with a series change is the culprit.

; To test, select the any item that isn't `yikes!`, click or scroll and all
; is well. As soon as a yikes! item is selected, things go haywire. Sometimes
; eventually clearing the list or crashing.

temp-blk: copy []

groups: [
;	"yikes!"	(collect [foreach [name obj] [a b c d e f] [keep name]])
;	"yikes!"	(collect/into [foreach [name obj] [a b c d e f] [keep name]] temp-blk)
;	"yikes!"	(blk: clear [] foreach [name obj] [a b c d e f] [append blk name])
;	"yikes!"	(blk: clear [] foreach [n o] [a b c d e f] [append blk n])
;	"yikes!"	(blk: clear [] vals: [a b c d e f] forall vals [append blk vals/1])
;	"yikes!"	(foreach [name obj] [a b c d e f] [take append [] 1])
;	"yikes!"	(foreach [name obj] [a b c d e f] [append [] 1])
;	"yikes!"	(foreach [v] [1] [append [] v])
;	"yikes!"	(repeat i 1 [append [] i])
	"yikes!"	(loop 1 [append [] 1])
	
; Perhaps a clue here, with on-face-deep-change*
;	"yikes!"	(foreach [name obj] [a b c d e f] [append [] ()])
;1 	paren 	(foreach [name obj] [a b c d e f] [append [] ()])
;*** Script Error: new has no value
;*** Where: on-face-deep-change*
;*** Stack: view do-events do-actor do-safe group-selected on-face-deep-change* 
	
	"no-yike-0"	([append [] 1])
	"no-yike-1"	(blk: clear [] [append blk 'name])
	"no-yike-2"	([1])
	"no-yike-3"	(foreach [name obj] [a b c d e f] [1])
	"no-yike-4"	(foreach [name obj] [a b c d e f] [[]])
	"no-yike-5"	(foreach [name obj] [a b c d e f] [take []])

	"favorite" 	[parse append rejoin if either func]
	"all" 		#all ; what do we use to say "all docs, whether in current build or not?
[...]

PR #4659, opened 2020-09-24
FIX: issue #4576 (when face/data changed, better use change-data)

bitbegin:


bitbegin:
filter data in `on-face-deep-change*` maybe better

qtxie:
Issue fixed by commit https://github.com/red/red/commit/6b964ca942d42db1d80ecdbde2bf04caab54c930

PR #4663, opened 2020-10-01
FIX: regression in PR 4144

hiiamboris:
also provides tests to ensure `do` cannot be used to make variadic calls


dockimbel:
Thanks!

PR #4671, opened 2020-10-09
FIX: set panel with option issue in layout

bitbegin:
example:

```
Red [
	Title: "title"
	Needs: 'View
]


OpenBrowser: function [face event][
]


; View splash screen
x: layout/options [size 800x600   
    at 50x460 bg: text 150x30 black white
    at 650x450 button 100x50 white red center "P L A Y" ] [actors: context [on-up: func [face event][OpenBrowser face event]]]
view x
```

### DIFF
--- a/modules/view/VID.red
+++ b/modules/view/VID.red
@@ -761,7 +761,7 @@ system/view/VID: context [
 
 		if all [focal-face find panel/pane focal-face not parent][panel/selected: focal-face]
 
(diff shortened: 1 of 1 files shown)


dockimbel:
That's not the right fix. The SET/SOME behavior has been corrupted by a previous PR and the new https://github.com/red/red/commit/ee607d5aa3ac90c7b906e7fbe1916b133d2b4051 commit fixes it.

PR #4677, opened 2020-10-13
 FIX: issue #4196 (crash on execution of malformed routine). 

### LINKED ISSUE #4196: crash on execution of malformed routine
Related: https://github.com/red/red/issues/3171, https://github.com/red/red/issues/4192

**Describe the bug**

`clear`ing `routine!`s spec block and then calling it leads to a crash.

**To reproduce**

```red
>> clear spec-of :as-ipv4
== []
>> as-ipv4 1 3 3 7
```
```
*** Runtime Error 98: assertion failed
*** in file: .../red/runtime/interpreter.reds
*** at line: 560
***
***   stack: red/interpreter/exec-routine 00000062h
***   stack: red/interpreter/exec-routine 02D45D60h
***   stack: red/interpreter/eval-code 02D45D60h 03138268h 031382A8h false 00000000h 00000000h 02D45D60h
***   stack: red/interpreter/eval-expression 03138268h 031382A8h false false false
***   stack: red/interpreter/eval 02C42CB4h true
***   stack: ctx||504~try-do 00EC0508h
***   stack: ctx||504~do-command 00EC0508h
***   stack: ctx||504~eval-command 00EC0508h
***   stack: ctx||504~run 00EC0508h
***   stack: ctx||504~launch 00EC0508h
***   stack: ctx||536~launch 00EBF7C4h
```

**Expected behavior**

Ideally Red should ignore any changes to function's spec (see related tickets linked above).

**Platform version**

9214:
Fixes #4196 by adding an extra error check in the interpreter, which uses the content of routine's spec block for argument type checking. Crash occured when there was a mismatch between routine's arity (stored in cell's header) and the actual number of arguments in the spec block.

Regression test is provided, although it's quite devious: by malforming `as-ipv4` (and by extension `as-rgba`) it prevents these routines from being used in the rest of the tests.

### DIFF
--- a/environment/system.red
+++ b/environment/system.red
@@ -119,6 +119,7 @@ system: context [
 				bad-func-def:		["invalid function definition:" :arg1]
 				bad-func-arg:		["function argument" :arg1 "is not valid"]
 				bad-func-extern:	["invalid /extern value:" :arg1]
+				bad-routine-def:	["invalid routine definition:" :arg1]
 				no-refine:			[:arg1 "has no refinement called" :arg2]
 				bad-refines:		"incompatible or invalid refinements"
 				bad-refine:			["incompatible refinement:" :arg1]
--- a/runtime/interpreter.reds
+++ b/runtime/interpreter.reds
@@ -262,6 +262,10 @@ interpreter: context [
 			value: s/offset
 			tail:  s/tail
 			
+			if all [value = tail args <> 0][			;-- issue #4196
+				fire [TO_ERROR(script bad-routine-def) block/push-only* 0]
+			]
+			
 			until [										;-- scan forward for end of arguments
 				switch TYPE_OF(value) [
 					TYPE_SET_WORD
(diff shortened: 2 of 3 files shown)


9214:
Although, reviewing the code, it actually makes more sense to modify `spec-of` reflector for routines, so that it returns a copy of a spec block rather than the original, because tampering with it (specifically with type specifications) will definitely trip up the interpreter inside the type-checking loop: for example, by removing one of the `block!`s we can cause another segfault on `TYPE_OF(value)`.
https://github.com/red/red/blob/99910dc065e908ac9b35224f482a1efac98c8128/runtime/interpreter.reds#L279-L301
OTOH that would render `spec-of` inconsistent with the rest of `any-function!` datatypes. So, either that or hardening the loop itself with extra checks. @dockimbel what do you think?

greggirwin:
I'd rather copy it, accepting that limitation, for safety. Specs are most helpful reflectively. If someone comes up with an advanced scenario where modifying routine specs is important, they can fix it then. :^)

Issue #4613, opened 2020-08-07
compiled script crashes on MONEY! literal with user-defined currency code


9214:
**Describe the bug**
`money!` with custom currency codes rely on `system/locale/currencies/list` which is checked at run-time: if the currency is not in it then it deemed to be invalid.

Since the extension of this list with custom currencies can happen only at run-time, the compiler cannot say whether given `money!` literal is valid or not. To resolve this problem, a custom `Currencies` field was added to the script header, which is processed at compile-time to extract user-defined currencies: the pre-defined list of currencies is copied and merged with the `block!` of user-defined ones.

The logic of this processing, however, seems to be flawed.

**To reproduce**

#### 1
```red
Red [] probe bug$0
```
```red
*** Runtime Error 98: assertion failed
*** in file: .../runtime/datatypes/money.reds
*** at line: 191
***
***   stack: red/money/get-currency-code 98
***   stack: red/money/get-currency-code 0
***   stack: red/money/make-in 02643194h false 0 004C807Ch
***   stack: red/money/push false 0 004C807Ch
```

`get-currency-code` expects a 1-based index in the currency list and is given a `0` (no clue about the top-level call with `98` though, looks like an error-reporting issue; note that `98` is also an ID of runtime error being reported).

https://github.com/red/red/blob/01a2e5d03358b115d520f685b64151627230c5f6/compiler.r#L256-L261

So, if currency code is not in the extracted list, and not in the one defined inside the header, then its index is a `0`. I'd rather expect a compile-time error on invalid `money!` literal.

#### 2
```red
Red [Currencies: [bug]] probe bug$0
```

```red
*** Runtime Error 1: access violation
*** in file: .../red/runtime/datatypes/string.reds
*** at line: 940
***
***   stack: red/string/alter 02B531B4h 02C30018h -1 0 true 0
***   stack: red/string/concatenate 02B531B4h 02C30018h -1 0 true false
***   stack: red/string/form 02C30018h 02B531B4h 02B531A4h 0
***   stack: red/word/form 004C1EC8h 02B531B4h 02B531A4h 0
***   stack: red/money/form-money 02B531A4h 02B531B4h 0 false false
***   stack: red/money/mold 02B531A4h 02B531B4h false false false 02B53194h 0 0
***   stack: red/actions/mold 02B531A4h 02B531B4h false false false 02B53194h 0 0
***   stack: red/actions/mold* -1 -1 -1 -1
***   stack: probe
```

[...]

PR #4688, opened 2020-10-23
TESTS: regression test for issue #4613 (compiled script crashes on MONEY! literal with user-defined currency code).

9214:
Provides compiler regression test for #4613, also fixes a minor omission: currency symbols from the header are now uppercased and processed in "normalized" form.


9214:
```text
/bin/bash: /Users/travis/build/red/red/quick-test/runnable/Users/travis/build/red/red/quick-test/runnable/testfile.txt: No such file or directory
```
No idea where this comes from. Newly added tests don't rely on any external files.

Issue #4698, opened 2020-10-30
[Regression] CLI-only Linux executables can't be produced in devmode anymore


hiiamboris:
**Describe the bug**

I have $DISPLAY var set to an IP address of the VM host and trying to produce an exe that doesn't require a running GUI environment (XServer is down on the host):
1. Script: `1.red`: `Red [] print "hello"`
2. Compile it: `red -c -t Linux-Old 1.red`
3. Run it: `./1` -- it just **hangs** (it will not hang if XServer is running on the host!)

Whereas `red -r -t Linux-Old 1.red` produces a working exe.

Notes:
1. Alternative compilation way (assuming Linux): `red -c --config "[sub-system: 'console]" 1.red` doesn't help the matter.
2. GTK-enabled compilation works just the same: `red -c -t Linux 1.red` will produce an exe that hangs, `red -r -t Linux 1.red` will produce an exe that works regardless of the XServer state.
3. If I `unset DISPLAY`, then `./1` compiled in devmode tells me this and terminates (does not hang):
```
Unable to init server: Could not connect: Connection refused
(1:1145): Gtk-WARNING **: 20:13:05.183: cannot open display:
```

**Expected behavior**

1. Should print `hello` in both `-r` and `-c` compilation modes.
2. An error instead of hanging when it can't init the GUI.

**Platform version**
```
Red 0.6.4 for Linux built 16-Oct-2020/21:35:47+03:00 commit #df54dcd
```
W10 host running WSL2 VM with Ubuntu OS.

hiiamboris:
@9214 it's unlisted but it's in config.r ☺:
```
Linux-Old [                                                                                                                                                                                    OS:         'Linux                                                                                                                                                                         format:     'ELF                                                                                                                                                                           type:       'exe                                                                                                                                                                           dynamic-linker: "/lib/ld-linux.so.2"                                                                                                                                                       legacy:     [stat32]                                                                                                                                                                   ] 
```
`-t Linux` works the same, just tried.

Issue #4700, opened 2020-10-30
Compiled script can't load images on Linux


hiiamboris:
**Describe the bug**

1. Create an image (or use any existing image): `reddo "save/as %test.png draw 100x100 [line-width 10 pen red circle 50x50 40] 'png"`

2. Create a loader script `1.red`:
```
Red []
i: load/as %test.png 
?? i
```

3. Compile & run: `red -r -e 1.red; ./1`, output is:
```
i: %test.png
```

**Expected behavior**

A dump of image data. When interpreted it works as expected.

**Platform version**
```
Red 0.6.4 for Linux built 16-Oct-2020/21:35:47+03:00 commit #df54dcd
```

qtxie:
`View` module is needed in order to use image! on Linux. There are two ways to solve this issue:
1. Throw an error when loading image.
2. Support image! in Red/Core on Linux. Then we have to install `libgdk-pixbuf` on Linux. Ubuntu as an example:
```
$ sudo apt install libgdk-pixbuf2.0-0:i386
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following additional packages will be installed:
  libblkid1 libblkid1:i386 libbsd0:i386 libgdk-pixbuf2.0-0 libgdk-pixbuf2.0-bin libgdk-pixbuf2.0-common
  libglib2.0-0:i386 libjbig0 libjbig0:i386 libjpeg-turbo8 libjpeg-turbo8:i386 libjpeg8 libjpeg8:i386 liblzma5
  liblzma5:i386 libmount1 libmount1:i386 libpcre2-8-0:i386 libpcre3:i386 libpng16-16:i386 libselinux1:i386
  libtiff5 libtiff5:i386 libwebp6 libwebp6:i386 libx11-6:i386 libxau6:i386 libxcb1:i386 libxdmcp6:i386
  libzstd1:i386
The following NEW packages will be installed:
  libblkid1:i386 libbsd0:i386 libgdk-pixbuf2.0-0 libgdk-pixbuf2.0-0:i386 libgdk-pixbuf2.0-bin
  libgdk-pixbuf2.0-common libglib2.0-0:i386 libjbig0 libjbig0:i386 libjpeg-turbo8 libjpeg-turbo8:i386
  libjpeg8 libjpeg8:i386 liblzma5:i386 libmount1:i386 libpcre2-8-0:i386 libpcre3:i386 libpng16-16:i386
[...]

PR #4701, opened 2020-10-31
FEAT: enable image! in red/core on Linux

qtxie:
Fix issue #4698 and issue #4700.

### DIFF
--- a/environment/codecs/BMP.red
+++ b/environment/codecs/BMP.red
@@ -17,20 +17,14 @@ put system/codecs 'bmp context [
 	suffixes: [%.bmp]
 	
 	encode: routine [img [image!] where [any-type!]][
-		#if not find [Android Linux FreeBSD NetBSD Syllabe] OS [
+		#if not find [Android FreeBSD NetBSD Syllabe] OS [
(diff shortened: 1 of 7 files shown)


qtxie:
Notes this PR will add another dependency (libgdk-pixbuf2.0) in Red/Core on Linux.

hiiamboris:
Just so you know, `sudo apt install libcanberra-gtk3-module:i386` is also required right now (at least on Debian 9 VM), else it spams messages that it can't load this module

9214:
@qtxie @hiiamboris feel free to update [this](https://github.com/red/red/wiki/%5BNOTE%5D-Linux-GTK-dependencies) page as you deem fit.

hiiamboris:
Oh wow. @9214 I didn't install most of that :D

9214:
@hiiamboris I did not check if all the dependencies are required, so if it works OK for you, then please remove the ones that are redundant.

qtxie:
> Just so you know, `sudo apt install libcanberra-gtk3-module:i386` is also required right now (at least on Debian 9 VM), else it spams messages that it can't load this module

Yes if you include the view module. It's included by default in red executable unless you use `--no-view`.

Issue #4713, opened 2020-11-07
R/S floating point output inconsistencies


hiiamboris:
**Describe the bug**

The following script:
`Red/System [] print-wide [1.0 / 0.0  -1.0 / 0.0  0.0 / 0.0  "^/"]`

On Windows produces the following:
`1.#INF -1.#INF -1.#IND.0`
In 2019 builds it was producing:
`1.#INF.0 -1.#INF.0 -1.#IND.0`

On Linux it produces:
`inf.0 -inf.0 -nan.0`
On RPi:
`inf -inf 1.#NaN`

**Expected behavior**

Same output on all platforms.
`1.#INF -1.#INF 1.#NAN` ?

**Platform version**
```
Red 0.6.4 for Windows built 16-Oct-2020/12:58:41+03:00 commit #e5876af
```

dockimbel:
@qtxie Please confirm that the fix is correct.

PR #4714, opened 2020-11-09
FIX: issue #4713 (print special float! not right)

bitbegin:


qtxie:
https://en.wikipedia.org/wiki/IEEE_754-1985, All NaNs in IEEE 754-1985 have this format:

sign = either 0 or 1.
biased exponent = all 1 bits.
fraction = anything except all 0 bits (since all 0 bits represents infinity).

A better test for NaN:
```
if d/2 and 7FF00000h = 7FF00000h [
	either all [
		zero? d/1
		zero? (d/2 and 000FFFFFh)
	][
		either zero? (d/2 and 80000000h) [prin "1.#INF"][prin "-1.#INF"]
	][prin "1.#NaN"]
]
```

bitbegin:
this pr is more simple for me, and can cover all cases

qtxie:
It's hard for me to understand the code without explaining the rational of it in a PR, also no comments in the code.

bitbegin:
https://www.doc.ic.ac.uk/~eedwards/compsys/float/nan.html

bitbegin:
图片

make this check simply, treat both SNAN and QNAN as NAN

qtxie:
@bitbegin Thanks. In this PR you check those special values by cases. While I prefer to do the checking by rules. See my first comment. A more lengthy version as below:
1. if all bits in exponent part are 1, it's a special value (NaN or INF)
2. if it's special value, check fraction part
3. if all bits in fraction part are 0, it's a INF, otherwise it's a NaN
4. if it's a INF, check sign bit to distinguish -INF and +INF

Issue #3999, opened 2019-08-20
Callback function returns struct value causes crashes on macOS


qtxie:
**To reproduce**
1. Compile the following C code as a dylib on macOS.
testc.c
`clang -dynamiclib -arch i386 ~/Downloads/testc.c -o libtestc.dylib`
```
#include <stdio.h>

typedef struct {
    float x;
    float y;
    float w;
    float h;
} MyRect;

static int* callback_func;

extern void set_callback(int* ptr) {
    callback_func = ptr;
}

extern int test_callback() {
    MyRect (*p)() = (MyRect (*)())callback_func;
    MyRect rc = p();
    printf("%f", rc.x);
    return 0;
}

```
2. Compile the following Red/System file.
testrs.reds
```
Red/System []

#import [
	"libtestc.dylib" cdecl [
        set_callback: "set_callback" [ptr [int-ptr!]]
        test_callback: "test_callback" [return: [integer!]]
    ]
]

MyRect!: alias struct! [
    x   [float32!]
    y   [float32!]
    w   [float32!]
    h   [float32!]
]

callback-func: func [
    [cdecl]
    return: [MyRect! value]
    /local
        rc  [MyRect! value]
][
    rc/x: as float32! 23.0
    rc/w: as float32! 123.0
    rc
]

test: func [][
    set_callback as int-ptr! :callback-func
    test_callback
]

test
```
3. Put those two files in the same folder and run `testrs`.

It crashes as below:
```
*** Runtime Error 32: segmentation fault
*** at: A7AA6BD0h
```

**Expected behavior**
print 23.000000 in the terminal.

**Platform version (please complete the following information)**
```
macOS 10.14, latest Red version in master branch.
```
Here are the pre-built binaries:
[prebuilt-test.zip](https://github.com/red/red/files/3521519/prebuilt-test.zip)

qtxie:
FYI. It works well on Linux.

dockimbel:
The issue is caused by CLang injecting an extra stack slot pop/push when it is not needed at all (probably a consequence of an internal CLang compiler limitation):

![image](https://user-images.githubusercontent.com/411393/64078870-a2f1b180-cce0-11e9-8611-2d02e8f0d421.png)

PR #4717, opened 2020-11-10
IEEE-754 compliance in float comparison operators

### LINKED ISSUE #847: NaN comparision doesn't follow IEEE754 spec.

### LINKED ISSUE #2567: 1.#NaN tests as negative
```
>> negative? 1.#NaN
== true
>> positive? 1.#NaN
== false
>> zero? 1.#NaN
== true
```
See also #847

hiiamboris:
See PR #3902
Fixes #847, fixes #2567

### DIFF
--- a/runtime/macros.reds
+++ b/runtime/macros.reds
@@ -548,7 +548,7 @@ Red/System [
 #define GET_INT_FROM(n spec) [
 	either TYPE_OF(spec) = TYPE_FLOAT [
 		fl: as red-float! spec
-		if any [fl/value > 2147483647.0 fl/value < -2147483648.0][
+		if any [fl/value > 2147483647.0 fl/value < -2147483648.0 fl/value <> fl/value][
 			fire [TO_ERROR(script out-of-range) fl]
 		]
 		n: as-integer fl/value
--- a/runtime/macros.reds
+++ b/runtime/macros.reds
@@ -548,7 +548,11 @@ Red/System [
 #define GET_INT_FROM(n spec) [
 	either TYPE_OF(spec) = TYPE_FLOAT [
 		fl: as red-float! spec
-		if any [fl/value > 2147483647.0 fl/value < -2147483648.0 fl/value <> fl/value][
+		if any [
+			fl/value >  2147483647.0
+			fl/value < -2147483648.0
+			fl/value <> fl/value						;-- NaN check (;
+		][
 			fire [TO_ERROR(script out-of-range) fl]
 		]
 		n: as-integer fl/value
--- a/system/targets/IA-32.r
+++ b/system/targets/IA-32.r
@@ -717,7 +717,7 @@ make-profilable make target-class [
 
 	emit-not: func [value [word! char! tag! integer! logic! path! string! object!] /local opcodes type boxed][
 		if verbose >= 3 [print [">>>emitting NOT" mold value]]
-		
+
 		if object? value [boxed: value]
 		value: compiler/unbox value
 		if block? value [value: <last>]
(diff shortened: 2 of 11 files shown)


greggirwin:
I love seeing all those tests. :+1: Looks like nice work.

hiiamboris:
I broke smth in compiler.r though :D Need to figure it out...

qtxie (commented):
So this PR implemented `option 1` in https://github.com/red/red/issues/847#issuecomment-47734318 .
I'm not capable to review the code in the compiler. I reviewed all the tests, they're all OK. If the changes in the compiler are good, it's good to be merged.

hiiamboris:
`conditional-test.reds` succeeds on Bolek's RPi. Can't do nothing about it without reproducing :(

Same for the Red test suite:
```
boris@raspberrypi:~ $ ./run-all-comp1
~~~started test~~~ run-all-comp1




~~~finished test~~~  run-all-comp1
  Number of Tests Performed:      4470
  Number of Assertions Performed: 7830
  Number of Assertions Passed:    7830

boris@raspberrypi:~ $ ./run-all-comp2
~~~started test~~~ run-all-comp2


ok
~~~finished test~~~  run-all-comp2
  Number of Tests Performed:      3084
  Number of Assertions Performed: 5683
  Number of Assertions Passed:    5683
```

Will look into the Mac issue in a few days when I get more bandwidth to DL & set up a Mac VM.

runtime/macros.reds
```diff
@@ -548,7 +548,7 @@ Red/System [
 #define GET_INT_FROM(n spec) [
 	either TYPE_OF(spec) = TYPE_FLOAT [
 		fl: as red-float! spec
-		if any [fl/value > 2147483647.0 fl/value < -2147483648.0][
+		if any [fl/value > 2147483647.0 fl/value < -2147483648.0 fl/value <> fl/value][
```
dockimbel, line 551: ???

qtxie: We have a function to test NaN in float.reds: `float/NaN?`

dockimbel: I mean what is this `fl/value <> fl/value` part for?

hiiamboris: To test for a NaN apparently ;)

hiiamboris:
On Mac 10.14 VM:
```
1s-Mac:xfer a1$ ./struct-test
~~~started test~~~ struct!
~~~finished test~~~ struct!
  Number of Tests Performed:      152
  Number of Assertions Performed: 571
  Number of Assertions Passed:    571
```

hiiamboris:
Funny, this PR either breaks Darwin (Mach-O) emitter or highlights a problem in it (I didn't touch the format emitters). Even `Red/System [] probe "hello world"` segfaults with `-t Darwin`, though all big enough scripts seem to be okay. `-t MacOS` is OK too.

hiiamboris:
No, sorry. This is some VMWare glitch related to file caching. Weird issue.

hiiamboris:
So, about failing "#3999-D" and "33999-F" tests: they never worked on Mac in the first place. Both return a NaN instead of `23.0`. `FFC00000h` instead of float-32 23.0, and `FFF80000 00000000` instead of float-64 23.0. It just happened that the previous IA-32 emitter treated `nan = 23.0` as true expression.

hiiamboris:
https://github.com/red/red/issues/4726

system/targets/IA-32.r
```diff
@@ -1318,34 +1318,137 @@ make-profilable make target-class [
...
+		if cond [opcode: add-condition op copy opcode]	;-- use `op` to modify the conditional jump Jcc
+		if back? [									;-- when jumping back, offset should account for the jump instruction size
+			size: size + (length? opcode) + (pick [1 4] short?)
+			set  bind 'size bind? ctx  size			;-- export size back to emit function
+			o: do jmpofs							;-- recalculate offset with new size
+		]
+		o: do reduce [pick [to-bin8 to-bin32] short?  'o]	;-- make binary signed offset
+		rejoin [opcode o]
+	]
+
```
dockimbel, line 1348: This function cannot be accepted in its current form. The hidding implicit data passing style is obfuscation. Please pass the data you need by regular arguments and return a block of values.

Moreover, I can't make sense of the following code:
```
o: do reduce [pick [to-bin8 to-bin32] short?  'o]
```
It evaluates one of the two conversion function, but without an argument?

I think you should rewrite more cleanly this function or redesign properly the whole `emit-branch` function.

runtime/macros.reds
dockimbel: That addition deserves a pretty good comment at least. ;-)

system/targets/IA-32.r
hiiamboris: `do reduce [pick [to-bin8 to-bin32] short?  'o]` -> `do [to-bin8 o]` or `do [to-bin32 o]`

greggirwin: How about `either short? [to-bin8 o][to-bin32 o]` ? Saves `do + reduce` and is blatantly obvious.

greggirwin: I'm not a fan of `pick` for binary choices in general. Sometimes yet, but not often.

hiiamboris: :+1:

system/targets/IA-32.r
```diff
@@ -1320,29 +1320,29 @@ make-profilable make target-class [
...
+		psize [word!] "word 'size set to jump size"
+		back? [logic! none!]
+		/local opcode o jmpofs short? size
 	][
-		;; import values from the emit function
-		set [op size back?] reduce bind [op size back?] bind? ctx
-		o: do jmpofs: [size * pick [-1 1] yes = back?]		;-- convert size to signed jump offset
-		;; test if qualifies for a short jump
+		size: get psize
+		o: do jmpofs: [size * pick [-1 1] yes = back?]	;-- convert size to signed jump offset
```
greggirwin, line 1329: Please add a comment about the thunk-like use of `jmpofs`. Took me a minute to find it and see how it was used.

hiiamboris: In real code (not in diff) it's just a few lines above. Can't miss it. Not sure what kind of comment you expect ;)

greggirwin: OK. I was thinking something like "jmpofs is used to recalc later after size is adjusted."

greggirwin:
:+1:

system/targets/IA-32.r
```diff
@@ -1318,34 +1318,136 @@ make-profilable make target-class [
...
+		][
+			flip?: no									;-- condition inverted? flag
 			op: case [
-				block? op [									;-- [cc] => keep
+				block? op [								;-- [cc] => keep
 					op: op/1
-					either logic? op [pick [= <>] op][op]	;-- [logic!] or [cc]
+					either logic? op [					;-- [logic!] or [cc]
+						pick [= <>] op
+					][ op ]
```
dockimbel, line 1372: Why did you change the coding style of line 1336?

system/targets/IA-32.r
```diff
@@ -1318,34 +1318,136 @@ make-profilable make target-class [
...
+				]
+				logic? op [pick [= <>] op]				;-- test for TRUE/FALSE
+				'else 	  [
+					flip?: yes 							;-- flip unordered target along with the condition
+					opposite? op						;-- 'cc => invert condition; unordered defined by the original op
+				]
+			]
+
+			unord-jumps-to-true?: either flip? [		;-- should unordered JP jump lead to true branch?
```
dockimbel, line 1381: `unord-jumps-to-true?` is way too big name. It's like trying to fit a whole comment line into a variable's name... Please use a short name and let the comment line explain its purpose and usage.

system/targets/IA-32.r
```diff
@@ -1318,34 +1318,136 @@ make-profilable make target-class [
...
+			[#{70} #{0F80}]							;-- Jcc short/near
+			[#{7A} #{0F8A}]							;-- JP  short/near
+		]	either op = 'parity [ 3 ][ none? op ]	;-- pick row: 1 = normal, 2 = conditional, 3 = parity
+			short? 									;-- pick column
+		if all [op op <> 'parity] [
+			opcode: add-condition op copy opcode	;-- use `op` to modify the conditional jump Jcc
+		]
+		if back? [									;-- when jumping back, offset should account for the jump instruction size
+			size: size + (length? opcode) + (pick [1 4] short?)
+			set psize size							;-- update size in the caller
```
dockimbel, line 1342: There is still some obfuscated data-passing between the caller and this function using a word (`size`) bound in calling context. That's really a confusing style we don't want to have here. The code logic is already complicated enough. Just return it in the return block (yes, you need to change the way that returned block is used then, maybe the binary code forming should happen here and just return the formed bytes and size).

dockimbel (changes requested):
Is ARM backend fully supported now?

system/targets/IA-32.r
greggirwin: @dockimbel please suggest a name you want. Otherwise we just keep guessing. Personally I don't think this name is too long, and it's clear without a comment.

system/targets/IA-32.r
greggirwin: The new format actually reads better to me. The original is very dense. But then I'm also not a fan of `pick` used for logic options because I think it's less clear.

system/targets/IA-32.r
hiiamboris: Hmm. I think it's no more confusing that `incr: func [word] [set word 1 + get word]` or any other function that works with words. But okay, I'll make it super verbose ☺

hiiamboris:
> 
> 
> Is ARM backend fully supported now?

Yes. I think it fails due to issue you fixed recently with IDIV.

system/targets/IA-32.r
hiiamboris: IIRC because the line didn't fit into 56 chars and I wanted to align comments.

system/targets/IA-32.r
hiiamboris:
```
		jxx: [second set [size jump-code] construct-jump op      size back?]
		jp:  [second set [size jump-code] construct-jump 'parity size back?]
```
`jump-code` is never used but kind of tells what `second` returns here.

system/targets/IA-32.r
dockimbel: I'll retract my comment on this one, as that longer name is a minor annoyance to me, compared to the previous one with special symbol in it.

system/targets/IA-32.r
dockimbel: > The new format actually reads better to me. The original is very dense. But then I'm also not a fan of `pick` used for logic options because I think it's less clear.

It's supposed to be dense, so that all the related code fits into one page in a code editor, in order to make it easier to reason about when debugging or upgrading it.

greggirwin: The great thing about code is that one person's "dense" is another's "makes me feel dense."

Issue #4703, opened 2020-10-31
Red filesystem access is buggy on mounted Windows drives on WSL2


hiiamboris:
**Describe the bug**

This bug involves Ubuntu Linux set up in WSL2 running on a W10 host.
[WSL2](https://docs.microsoft.com/en-us/windows/wsl/install-win10) mounts Windows drives automatically as `/mnt/c`, `/mnt/d` and so on, and one is able to work with host's files.

1. `read` on directories always returns an empty block. Try `ls %/mnt/c` or `ls %/mnt/c/Windows`, etc.
2. `read` on files returns junk bytes:
```
>> write %tmp "junk"
>> read %tmp
== {junk^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@...
```
There may be more bugs, I didn't test extensively because this is unusable.
I can't tell if it's a bug in Red or in WSL2 itself, but bash & every other tool works perfectly, so... you get it.

**To reproduce**

Assuming you have the VM set up, just run the above commands in console

**Expected behavior**

Working file access.

**Platform version**
```
Red 0.6.4 for Linux built 16-Oct-2020/21:35:47+03:00 commit #df54dcd
```
Every other mount point I've tested works fine. Just Windows (NTFS) drives that produce the glitch. This is the mtab contents, if it can tell anything (last 2 lines):
```
/dev/sdb / ext4 rw,relatime,discard,errors=remount-ro,data=ordered 0 0
tmpfs /mnt/wsl tmpfs rw,relatime 0 0
tools /init 9p ro,dirsync,relatime,aname=tools;fmask=022,loose,access=client,trans=fd,rfd=6,wfd=6 0 0
none /dev devtmpfs rw,nosuid,relatime,size=6469972k,nr_inodes=1617493,mode=755 0 0
[...]

PR #4723, opened 2020-11-13
FIX: issue #4703 (use lseek to get file size for linux platform)

bitbegin:


qtxie (commented):
Nice workaround. Let's see if we can fix the stat issue first.

Issue #3562, opened 2018-10-12
Can't run Red from other programs


maximvl:
**Describe the bug**
When I try to execute Red from another program it fails with ```PROGRAM ERROR: Invalid encapsulated data.```

**To Reproduce**
1. Install Elixir
2. Clone https://gitlab.com/maxvel/reddish.git
3. In the repo run `iex -S mix`
4. In the REPL type
```
p = Reddish.start
Reddish.send_input(p, "5 + 5\n")
flush()
```

5. See message `{#PID<0.170.0>, :data, :out, "PROGRAM ERROR: Invalid encapsulated data.\n"}`
6. After this I can't even run Red as regular:
```
> red
PROGRAM ERROR: Invalid encapsulated data.
```
7. And in a new shell instance:
```
> red
** Script Error: Invalid compressed data - problem: -3
** Near: script: decapsulate 
if none? script
```

**Expected behavior**
`flush()` prints message with "10" returned by Red, Red is runnable.

**Screenshots**
![screenshot](https://i.imgur.com/LpGHVrT.png)

**Platform version(please complete the following information):**
```
Red 0.6.3 for Linux built 11-Oct-2018/3:00:23+02:00 commit #b1f01cc
```

9214:
#543, #1547

maximvl:
I also tried to create a wrapper like this:
```
#!/bin/bash
red $@
```
With it I don't get the error, but got no output either. From what I tried this works with `cat` and `ipython`, but doesn't work with `python`.

dockimbel:
If you need only the Red interpreter, compile the CLI console and use that binary instead of `red.exe`.

x8x:
I work around this issue by wrapping the Red call into `sh` like this:
```sh
sh -c "red -r code.red"
```
I suspect a path issue in certain environment ~where the console called from red binary can't find the crush decompressor~ maybe not as in the example the toolchain is called not the console.., but haven't investigated further.

beenotung:
I found similar bug on windows.
I saved red as red.exe, and add the containing folder into PATH.
I found it will raise below error when launch form cmd:
```
PROGRAM ERROR: Invalid encapsulated data.
```

## below case will trigger error
1. if the current directory is not containing red, and using `red` or `red.exe`
2. if the current directory is not containing red, and using `..\local\bin\red`

## below case will not trigger error
1. if the current directory is containing red, and using `red` or `red.exe`
2. if the current directory is not containing red, and using `..\local\bin\red.exe`

[...]

Issue #1547, opened 2016-01-23
[linux/fish-shell] Red either fails or runs depending on the way it is called


xaduha:
one@voidlinux ~> pwd
    /home/one
    one@voidlinux ~> which red
    /usr/bin/red
    one@voidlinux ~> red
    PROGRAM ERROR: Invalid encapsulated data.
    one@voidlinux ~> /usr/bin/red
    --== Red 0.5.4 ==--
    Type HELP for starting information.

    red>>

I'm using fish-shell, so it probably has something to do with the environment. Still, something is fishy.

    one@voidlinux ~> env | sort > env1.txt
    one@voidlinux ~> bash
    [one@voidlinux ~]$ env | sort > env2.txt
    [one@voidlinux ~]$ diff env*
    16c16
    < SHLVL=2
    ---
    > SHLVL=3
    22a23
    > _=/usr/bin/env

Bottomline, works under bash, doesn't work under fish when called without specifying the absolute path.

~~Doesn't happen with `red-054` or the automated build that I previously used, commit 7526673.~~

EDIT: Having a red binary in a current folder allows it to launch normally, `red` and `./red` both.

xaduha:
I don't know if it's in any way helpful, but other errors I had with this

    bash-4.3# red -d
    *** Driver Internal Error: Access Error : Bad file path: //.red/
    *** Where: throw-on-error
    *** Near:  [make-dir temp-dir]

and 

    # red -d
    ** Script Error: Invalid compressed data - problem: -3
    ** Near: script: decapsulate
    if none? script

xaduha:
Looks like it's an old issue, related to #543.

spinningarrow:
I just got this issue on macOS with the fish shell. Should I reopen it?

dockimbel:
@spinningarrow Have you looked at the linked issue https://github.com/red/red/issues/543 ?

spinningarrow:
Yes, I did. That one is closed too and I couldn't really find a proper resolution for it (other than creating an alias to it, if I understood correctly). What I'm trying to say is - is this something red can fix from its side; or failing that, provide an error message to the user when it's started incorrectly what the workaround is?

I'm also not sure why it works fine with bash but not with fish - what's the difference in the way the binary is invoked?

gour:
> I'm also not sure why it works fine with bash but not with fish - what's the difference in the way the binary is invoked?

I also wonder about it?

[...]

Issue #543, opened 2013-09-30
On Windows, Red added to the path, can't be run from another folder on cmd


SoleSoul:
## Summary

Red can't be run from a different path on Windows.
## How to reproduce
- Put red.exe in a folder that is in the path
- On the command line, navigate to another folder (not the one where red.exe is)
- type 'red'
## Expected behavior

The Red interpreter should run.
## Observed behavior

c:>red
PROGRAM ERROR: Invalid encapsulated data.

AJMansfield:
I had this same issue when I was trying to get sublime text to use red as a build system on Ubuntu 15.04.
My solution was similar; I created a shell script like this:
```bash
#!/bin/bash
red $@
```
and called that instead.

cdokolas:
For the Windows batch file, I'd add a `@` in front, like this:
```
@"%~dp0red-exe.exe" %*
```
The way I've actually used it is by dropping the executable to my `C:\Utils\` folder (it's in the `PATH` variable) and also creating a `red.bat` file there with `@"%~dp0red-063.exe" %*` in it (I kept the original filename to keep track of the version I'm using).

PR #4739, opened 2020-11-30
An alternate way to create Red binary

qtxie:
This PR contains a script to create Red binary without using Rebol SDK. The created executable has the same features as the current one, plus some advantages:

1. Use red.exe from system path (fix issue #3562 #543, #1547)
2. Use red.exe immediately after downloaded, no precompiling needed.
3. More friendly to the OS's package manager (APT, YUM, etc.)
4. Maybe we can sign it to avoid warning of some Anti-virus software.

Here are the prebuilt binaries for testing:
[red-win.zip](https://github.com/red/red/files/5614742/red-win.zip)
[red-linux.zip](https://github.com/red/red/files/5614746/red-linux.zip)
[red-mac.zip](https://github.com/red/red/files/5614747/red-mac.zip)

### DIFF
--- a/build/build-red.r
+++ b/build/build-red.r
@@ -56,7 +56,7 @@ save-files red-repo src-files
 
 script: [Red [
 	Title: "Red Programming Language"
-	Version: 0.6.6
+	Version: 0.6.4
 	Rights:  "Copyright (C) 2014-2020 Red Foundation. All rights reserved."
 	Needs: view
 	]
--- a/build/build-red.r
+++ b/build/build-red.r
@@ -1,0 +1,44 @@
+Rebol [
+	Title:   "Red binary build script"
+	Author:  "Xie Qingtian"
+	File: 	 %build-red.r
+	Tabs:	 4
+	Rights:  "Copyright (C) 2011-2020 Red Foundation. All rights reserved."
+	License: "BSD-3 - https://github.com/red/red/blob/master/BSD-3-License.txt"
+]
+
+rebol-bin: read/binary to-rebol-file system/script/args
+src-files: pick load %includes.r 8
+
+;-- save source files
+red-repo: make block! 20
+save-files: func [out src-files /local f][
+	foreach f src-files [
+		either block? f [
+			change-dir last out
+			append/only out make block! 20
+			save-files last out f
+			change-dir %..
+		][
+			append out f
+			unless dir? f [append out read/binary f]
+		]
+	]
+]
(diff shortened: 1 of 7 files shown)


hiiamboris:
`--cli` doesn't work for me on W10
Also, it's 400kB bigger now.

greggirwin:
On PR#4739 GH won't let me upload the signed, zipped version in a comment it seems. Signed version works fine. Since our cert is Redlake's, that's who will show. I think this is OK, as Redlake is there to support Red. We may just want to note that somewhere. GH won't even let me comment now.

greggirwin:
Comment finally posted. Let me try uploading again.

greggirwin:
[red-win-signed.zip](https://github.com/red/red/files/5618676/red-win-signed.zip)

qtxie:
@greggirwin Thanks. :+1: Only two Anti-Virus tools report warning after signed. https://www.virustotal.com/gui/file/747251168d2058dbe8eeef9e6c72aed08739401170b92c9b66d54b22048fcb31/detection

qtxie:
> From the code it looks like only a single prebuilt console is included. By design?

Yes. Actually I'm not sure which way is better: A single binary includes all versions of the console or several downloading options. I was prefer the later. 
* The big red button on the download page on red-lang.org provide the standard version (CLI console + View module)
* An alternate download for GUI version (red-gui.exe, GUI console + View module)
* An alternate download for Core version (red-core.exe, CLI console, No view module)

I'm not sure how many users need `--cli`. We could support it. But is it worth to add it if it's too complicated and tricky?

qtxie:
@hiiamboris
> Also, only works with Rebol/Core. Rebol/View doesn't work.

What's the error you got? Rebol/View works fine in my test.

hiiamboris:
> 
> 
> @hiiamboris
> 
> > Also, only works with Rebol/Core. Rebol/View doesn't work.
> 
> What's the error you got? Rebol/View works fine in my test.

It just quits silently. Tested on Rebol/View 2.7.8.3.1 (not the 2.7.6. used by CI)

litew:
> * An alternate download for Core version (red-core.exe, CLI console, No view module)

Just 2 cents. I'll vote for this too as this option is just perfect for deploying Red on servers without any GUI-toolkits installed.

dockimbel:
I don't like that the `%engine.red` has been filled with unrelated functions. They should go in a separate file.

qtxie:
Use separated Red binaries now. This is not needed.

PR #4782, opened 2020-12-26
Direct2D DRAW Backend

qtxie:
Use Direct2D for drawing on Windows platform instead of the GDI/GDI+ implementation.

Pros:
* Now Rich-Text and DRAW can work together on Direct2D backend
* Faster on modern GPU
* Microsoft keeps improving Direct2D, while GDI/GDI+ is obsoleted

Cons:
* Setup up the Direct2D context takes more time than GDI+
* Direct2D API is more difficult to use than GDI/GDI+ API
* May behaviors badly on some graphic card drivers.

For compatibility, specify `Config: [draw-engine: 'GDI+]` in the header to use the GDI/GDI+ backend.
```
Red [
    Config: [draw-engine: 'GDI+]
]
```


hiiamboris:
why `legacy:` and not `GUI-engine:`?

greggirwin:
Assuming we're deprecating GDI+, `legacy` makes sense and makes it clear that it's not a choice for a feature-based reason. I don't know that `32-bit` would also be legacy, as bitness is a choice based on environment, not specifically age. I also don't know the config option naming plans for when we have a non-native GUI (or more than one). So `GUI-engine` can also make sense. It comes down to big picture consistency.

qtxie:
> why `legacy:` and not `GUI-engine:`?

Direct2D is a graphic engine, just for DRAW dialect. `GUI-engine` include all the kind of faces, windows manager, etc.

dockimbel:
@qtxie Could we change `legacy:` to `2D-engine:` rather? `legacy` term is too general. Also we might want to plug other 2D engines in the future.

qtxie:
@dockimbel Ok.

loziniak:
> ```
> Red [
>     Config: [draw-engine: 'GDI+]
> ]
> ```

Is something like this possible? I find it simpler, and is more consistent with `needs: 'View`:
```
Red [
    draw-engine: 'GDI+
]
```

PR #4804, opened 2021-01-19
20x faster JSON string decoder

hiiamboris:
before/after
```
length: 1000000
698 ms  [save/as bin: #{} s1 'json]
5297 ms [s2: load/as bin 'json]
s2 == s1 : false
```
```
length: 1000000
548 ms  [save/as bin: #{} s1 'json]
267 ms  [s2: load/as bin 'json]
s2 == s1 : true
```

### DIFF
--- a/environment/codecs/JSON.red
+++ b/environment/codecs/JSON.red
@@ -64,16 +64,10 @@ context [
 	;-----------------------------------------------------------
 	;-- String
 	string-literal: [
-		#"^"" copy _str [
+		#"^"" _s:
 			any [some chars | #"\" [#"u" 4 hex-char | json-esc-ch]]
-		] #"^"" (
-			if not empty? _str: any [_str copy ""] [
-				;!! If we reverse the decode-backslash-escapes and replace-unicode-escapes
-				;!! calls, the string gets munged (extra U+ chars). Need to investigate.
-				unescape _str
-				;replace-unicode-escapes decode-backslash-escapes _str
-			]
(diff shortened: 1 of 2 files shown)


greggirwin:
Really nice work @hiiamboris. :+1:. With JSON and CSV codecs, we're defining a model for future codecs that include Red code, rather than just the `put system/codecs ` aspect. I liked having that at the top of the file, but understand that it means forward references when the rest of the code comes later. I also like having everything in one place, rather than using `#include`, but am open to thoughts here. The main thing is that we should be consistent in the layout.

@rebolek, if this looks good to you, I'll merge it. When you have a chance, would you please make the `to-json-7/load-json-7` tests mirror images of each other, and have them cover all the backslash escape chars? Thanks.

hiiamboris:
> I liked having that at the top of the file, but understand that it means forward references 

Liked it myself, but it was not working when compiled as a standalone exe (without the whole console). I guess Red compiler used old `to-json` and `load-json` and newer ones were discarded.

greggirwin:
@hiiamboris I figured you had to move it for that reason.

@rebolek sync up `to/load` tests 7-9. The don't quite match up right now. Just a nice detail. :^)

greggirwin (approved):
All looks good. I tested a construct outside the PR that I hadn't used myself, so it's always nice to learn new tricks.

giesse:
Nothing is jumping out at me as being a problem.

PR #4805, opened 2021-01-20
FIX: removed uncalled-for series expansion

hiiamboris:
See crash reports from [January 20, 2021 1:10 AM](https://gitter.im/red/codecs?at=600758c9d8bdab473947fe66)

Cause: `string/overwrite-char` extended the series where it wasn't needed or expected.

### DIFF
--- a/runtime/datatypes/string.reds
+++ b/runtime/datatypes/string.reds
@@ -674,12 +674,16 @@ string: context [
 			UCS-4 [0]
 		]
 		unit: GET_UNIT(s)
-		
-		if ((as byte-ptr! s/tail) + unit) > ((as byte-ptr! s + 1) + s/size) [
-			s: expand-series s 0
+
+		loop 2 [
+			p: (as byte-ptr! s/offset) + (offset << (log-b unit))
+			either (p + unit) > ((as byte-ptr! s/offset) + s/size) [
+				s: expand-series s 0
+			][
+				break
+			]
 		]
 
-		p: (as byte-ptr! s/offset) + (offset << (log-b unit))
 		poke-char s p cp
 		if p >= as byte-ptr! s/tail [s/tail: as cell! p + unit]
 		s
--- a/tests/source/units/json-test.red
+++ b/tests/source/units/json-test.red
@@ -193,6 +193,14 @@ Red [
         str: {{"tag":"value","<tag":"value",">tag":"value","tag<":"value","<tag<":"value",">tag<":"value","tag>":"value","<tag>":"value",">tag>":"value","a<tag>b":"value"}}
         --assert res = load-json str
 
+    --test-- "load-json-23"         ;-- crash test for strings divisable by 16
+        s: "1234567812345678"
+        --assert s = load/as mold s 'json
+
+    --test-- "load-json-24"         ;-- crash test
+        o: load/as {{"location": "関西    ↓詳しいプロ↓"}} 'json
+        --assert o/location = "関西    ↓詳しいプロ↓"
+
 ===end-group===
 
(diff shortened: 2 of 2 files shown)


runtime/datatypes/string.reds
```diff
@@ -674,12 +674,16 @@ string: context [
 			UCS-4 [0]
 		]
 		unit: GET_UNIT(s)
-		
-		if ((as byte-ptr! s/tail) + unit) > ((as byte-ptr! s + 1) + s/size) [
-			s: expand-series s 0
+
+		loop 2 [
```
qtxie, line 678: Why loop twice?

qtxie:
This PR is a good improvement for `string/overwrite-char`, but the root cause of the crash is in the `unescape` function. https://github.com/red/red/blob/master/environment/codecs/JSON.red#L131-L135
```
130	    default [                           ;-- pass both chars
131	        string/overwrite-char s dst c1
132	        dst: dst + 1
133	    ]
134	]
135	s2: string/overwrite-char s dst c2
136	dst: dst + 1
```
line 131, `s` may be expanded, if so an invalid `s` will be passed to `overwrite-char` on line 135.
The fix is simple, just refresh `s` on line 131: `s: string/overwrite-char s dst c1`. Also it's better to refresh `s` on line 102.

runtime/datatypes/string.reds
hiiamboris: after expansion `s` changes thus we need to repeat getting `p:` set to the new `s`
if no expansion - break - no second iteration

hiiamboris:
@qtxie I disagree. If let's say I'm overwriting char 2 in string "abcdef" to "f" -> "afcdef", why should it expand it? I'm moving characters *already present* in the string (so no UCS1->UCS2->UCS4 expansion can happen) to the same place or backwards in index (so size is always enough). Previously expansion erroneously happened in these cases, now it doesn't. That's the whole fix. I was to update `s` I would have to repeat this whole code again:
https://github.com/red/red/blob/89b5f03b4ea74bf1b3b330c34c13a1de6e8ce005/environment/codecs/JSON.red#L138-L142
whereas I know for sure it's not needed in my case and will slow me down.

qtxie:
> If let's say I'm overwriting char 2 in string "abcdef" to "f" -> "afcdef", why should it expand it? 

No. It shouldn't in this case. But even it expends it, should not cause any crashes, just use more memory. That's why I said your PR is a good improvement for `string/overwrite-char`.

What if overwrite at the tail? In this case, `s` will be expanded. It's always safe to do this `s: string/overwrite-char s dst c1` in case `s` was expended.

> If I was to update s I would have to repeat this whole code again:

I don't see why you have to repeat it.

hiiamboris:
> What if overwrite at the tail? In this case, `s` will be expanded. It's always safe to do this `s: string/overwrite-char s dst c1` in case `s` was expended.

I understand that you need that in general case, but not for this one, don't you agree?

> I don't see why you have to repeat it.

OK You're correct, in this code I'm not using any of `unit`, `tail`, `src` or `s` after overwrite-char and before the end the loop.

Will it satisfy you if I change lines 102 and 131 to:
```
				s2: string/overwrite-char s dst c1
				assert s = s2
```
?
(in case some other day someone breaks overwrite-char again)

hiiamboris:
..and unroll the loop for clarity

greggirwin:
So what we need to tell users is that `overwrite-char` _may_ expand the string, e.g. in UCS cases, so it's a good practice to reset your string reference after calls to it in the general case.

Thanks all. Great teamwork.

Issue #4552, opened 2020-06-25
crash on probing context of SELF


9214:
**Describe the bug**
`self` is considered a keyword that objects use to refer to themselves. In some cases, the context to which these words are bound is malformed.

**To reproduce**
Minimal example (paste in console):
```red
context? object [return quote self]
```

To follow along, compile the script below with specified flags and execute it:
```red
Red [Note: "-r -e -d flags"]

dissect: routine [word [word!]][dump4 word/ctx/value]
recycle ; trigger series frame compaction
probe context? dissect object [return quote self]
```

Symbolic values hold a reference to a context to which they are bound. At implementation level such binding is realized as an indirect pointer to a series buffer with two slots:
1. `context!` itself;
1. So-called back-reference, which is an exact copy of an `object!` or `function!` that provides a context to a symbolic value.

Diagrammatically:

<img src="https://user-images.githubusercontent.com/15716466/85756592-3f24f300-b70f-11ea-8318-77460c3a1ccd.png" width="500">

When `context?` is called, it simply follows the binding and fetches the second slot.

```
root: 2755/5464, runs: 0, mem: 1381652 => 1376056, mark: 4.9ms, sweep: 1.0ms

Hex dump from: 02E16784h

02E16784: A0000010 00E63834  00000010 02E16798  ?..?48?.?...?g??
02E16794: 02E167A8 00000020  00E6381C 00000000  ?g?? ...?8?.....
02E167A4: 00000000 BADCAFE0  A0000004 00E63838  ....?????..?88?.
02E167B4: 00000030 02E167C0  02E167C0 BADCAFE0  0...?g???g??????
02E167C4: BADCAFE0 BADCAFE0  BADCAFE0 BADCAFE0  ????????????????
02E167D4: BADCAFE0 BADCAFE0  BADCAFE0 BADCAFE0  ????????????????
02E167E4: BADCAFE0 BADCAFE0  BADCAFE0 BADCAFE0  ????????????????
02E167F4: A0000010 00E6383C  00000080 02E16808  ?..?<8?.?...?h??

*** Runtime Error 1: access violation
*** in file: .../red/runtime/datatypes/string.reds
*** at line: 1079
***
***   stack: red/string/concatenate-literal-part 02663344h 00000000h 0
***   stack: red/datatype/form 02E16950h 02663344h 00000000h 0
***   stack: red/actions/form 02E16950h 02663344h 00000000h 0
***   stack: red/error/reduce 02663324h 02663304h
[...]

PR #4816, opened 2021-01-31
FIX: issue #4803 (Redbin a little too tight?)

### LINKED ISSUE #4803: Redbin a little too tight?
**Describe the bug**
```
Red []
b: []
repeat i 5000 [append b form i]    ;) 4999 does not trigger the error
save/as #{} b 'redbin
```
Output:
```
*** Internal Error: internal limit reached: 20000
*** Where: encode
*** Stack: save
```

Cause:
https://github.com/red/red/blob/de5c05e715291ee13afec3478416c3c6d52e912b/runtime/redbin.reds#L175

**Expected behavior**

No limit.

**Platform version**
```
Red 0.6.4 for Windows built 20-Dec-2020/21:04:56+03:00 commit #7a6d9a9
```

hiiamboris:
Waiting for @9214 review ;)

Summary:
- Fixes #4803 by dynamically adjusting buffers size
- Reduces load time from O(n^2) to O(n) by using hashtable

Before the PR:
```
testing with 5000 strings..
*** Internal Error: internal limit reached: 20000
*** Where: encode
*** Stack: clock save
```
Dynamic but without hashtable:
```
testing with 5000 strings..
60.0 ms	[save/as clear bin b 'redbin]
2.00 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 10000 strings..
219 ms	[save/as clear bin b 'redbin]
0.0 μs	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 15000 strings..
475 ms	[save/as clear bin b 'redbin]
3.00 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 20000 strings..
829 ms	[save/as clear bin b 'redbin]
985 μs	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 25000 strings..
1533 ms	[save/as clear bin b 'redbin]
3.00 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 30000 strings..
2022 ms	[save/as clear bin b 'redbin]
9.99 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 35000 strings..
2910 ms	[save/as clear bin b 'redbin]
6.96 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 40000 strings..
3775 ms	[save/as clear bin b 'redbin]
5.00 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 45000 strings..
4826 ms	[save/as clear bin b 'redbin]
6.99 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 50000 strings..
6127 ms	[save/as clear bin b 'redbin]
9.99 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
```
With hashtable:
```
testing with 5000 strings..
998 μs	[save/as clear bin b 'redbin]
1.00 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 10000 strings..
2.96 ms	[save/as clear bin b 'redbin]
999 μs	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 15000 strings..
5.00 ms	[save/as clear bin b 'redbin]
983 μs	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 20000 strings..
6.00 ms	[save/as clear bin b 'redbin]
1.97 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 25000 strings..
12.0 ms	[save/as clear bin b 'redbin]
1.99 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 30000 strings..
13.6 ms	[save/as clear bin b 'redbin]
2.51 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 35000 strings..
9.96 ms	[save/as clear bin b 'redbin]
2.98 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 40000 strings..
13.0 ms	[save/as clear bin b 'redbin]
3.00 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 45000 strings..
23.0 ms	[save/as clear bin b 'redbin]
4.00 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
testing with 50000 strings..
43.8 ms	[save/as clear bin b 'redbin]
4.02 ms	[c: load/as bin 'redbin]
loaded data valid? =  true
```
150 times faster on 50000 strings

Test code:
```
do https://gitlab.com/hiiamboris/red-mezz-warehouse/-/raw/master/clock.red
b: [] bin: #{}
recycle/off
loop 10 [
	repeat i 5000 [append b form i]
	print ["testing with" length? b "strings.."]
	clock [save/as clear bin b 'redbin]
	clock [c: load/as bin 'redbin]
	print ["loaded data valid? = " b = c]
	; recycle
]
```

### DIFF
--- a/runtime/redbin.reds
+++ b/runtime/redbin.reds
@@ -147,16 +147,16 @@ redbin: context [
 	
 	;-- Reference sub-system --
 	
-	reset: does [path/reset reference/reset]
+	reset: does [path/reset reference/reset]			;-- should be called before & after `encode`
 	
 	path: context [
-		stack: as int-ptr! 0
+		stack: as int-ptr! 0							;-- initialized by 'reset'
 		top:   stack
 		end:   stack
 		
 		push: func [/local newsz [integer!] new [int-ptr!]] [
 			if top + 1 > end [
-				newsz: (as-integer end - stack) * 3 / 2				;-- +50% of current size
+				newsz: (as-integer end - stack) * 3 / 2					;-- +50% of current size
 				new: as int-ptr! realloc as byte-ptr! stack newsz
 				if null? new [reset fire [TO_ERROR(internal no-memory)]]
 				top: new + (top - stack)
@@ -180,7 +180,7 @@ redbin: context [
 				free as byte-ptr! stack
 				stack: as int-ptr! 0
 			]
-			if null? stack [
+			if null? stack [											;-- start with the min. size
 				stack: as int-ptr! allocate min-size * size? integer!
 				end: stack + min-size
 			]
@@ -189,7 +189,9 @@ redbin: context [
 	]
 	
 	reference: context [
-		list: as int-ptr! 0
+		;-- a map of integer! -> red-integer!
+		map:  as node! 0								;-- initialized in 'reset'
+		list: as int-ptr! 0								;-- as well
 		top:  list
 		end:  list
 		
@@ -198,14 +200,11 @@ redbin: context [
 			return: [int-ptr!]
 			/local
 				here [int-ptr!]
+				slot [red-integer!]
 		][
-			here: list
-			while [here <> top][
-				if node = as node! here/value [return here + 1]
-				here: here + here/2 + 2
-			]
-			
-			null
+			slot: as red-integer! _hashtable/get-value map as-integer node
+			if null? slot [return null]
+			list + slot/value
 		]
 		
 		store: func [
@@ -215,10 +214,11 @@ redbin: context [
 				newsz [integer!]
 				reqsz [integer!]
 				new   [int-ptr!]
+				slot  [red-integer!]
 		][
 			size: (as integer! path/top - path/stack) >> log-b size? integer!
-			if top + size + 2 > end [
-				reqsz: (as-integer (top + size + 2) - list) + 8'192		;-- min. required + reserve for later
+			if top + size + 1 > end [									;-- (node=1) + size
+				reqsz: (as-integer (top + size + 1) - list) + 8'192		;-- min. required + reserve for later
 				newsz: (as-integer end - list) * 3 / 2					;-- +50% of current size
 				if newsz < reqsz [newsz: reqsz]
 				new: as int-ptr! realloc as byte-ptr! list newsz
@@ -227,11 +227,13 @@ redbin: context [
 				end: new + (newsz / size? integer!)
 				list: new
 			]
-			top/1: as integer! node
-			top/2: size
-			top: top + 2
-			copy-memory as byte-ptr! top as byte-ptr! path/stack size * size? integer!
-			top: top + size
+			assert not null? map
+			slot: as red-integer! _hashtable/put-key map as-integer node
+			assert not null? slot
+			integer/make-at as cell! slot (as-integer top - list) / size? integer!
+			top/1: size
+			copy-memory as byte-ptr! top + 1 as byte-ptr! path/stack size * size? integer!
+			top: top + size + 1
 		]
 		
 		reset: func [/local min-size] [
@@ -240,11 +242,17 @@ redbin: context [
 				free as byte-ptr! list
 				list: as int-ptr! 0
 			]
-			if null? list [
+			if null? list [												;-- start with the min. size
 				list: as int-ptr! allocate min-size * size? integer!
 				end: list + min-size
 			]
 			top: list
+			either null? map [
+				map: _hashtable/init 1024 null HASH_TABLE_INTEGER 1
+				_hashtable/mark map
+			][
+				_hashtable/clear-map map
+			]
 		]
 	]
 	
@@ -381,7 +389,7 @@ redbin: context [
 		symbols: binary/make-at stack/push* 4
 		table:   binary/make-at stack/push* 4
 		strings: binary/make-at stack/push* 4
-		
+
 		encode-value data payload symbols table strings
 		size: binary/rs-length? payload
 		
@@ -408,6 +416,7 @@ redbin: context [
 		
 		stack/pop 4
 		
+		reset
 		payload
 	]
 	
@@ -439,7 +448,6 @@ redbin: context [
 			p/4 = #"B" p/5 = #"I" p/6 = #"N"
 		][
 			either codec? [
-				reset
 				fire [TO_ERROR(script invalid-data) stack/arguments]
 			][
 				print-line "Error: Not a Redbin file!"
@@ -531,7 +539,7 @@ redbin: context [
 			TYPE_MONEY [(money/get-sign as red-money! data) << 20]
 			default    [0]
 		]
-		
+
 		switch type [
 			TYPE_UNSET
 			TYPE_NONE		[store payload header]
@@ -593,7 +601,7 @@ redbin: context [
 				unless global? [either null? ref [path/pop][encode-reference ref payload]]
 			]
 		]
-		
+
 		offset: offset + 1
 	]
 	
--- a/runtime/redbin.reds
+++ b/runtime/redbin.reds
@@ -235,6 +235,8 @@ redbin: context [
 			copy-memory as byte-ptr! top + 1 as byte-ptr! path/stack size * size? integer!
 			top: top + size + 1
 		]
+
+		on-gc-mark: does [_hashtable/mark map]
 		
 		reset: func [/local min-size] [
 			min-size: 16'384
@@ -249,13 +251,13 @@ redbin: context [
 			top: list
 			either null? map [
 				map: _hashtable/init 1024 null HASH_TABLE_INTEGER 1
-				_hashtable/mark map
+				collector/register as int-ptr! :on-gc-mark
 			][
 				_hashtable/clear-map map
 			]
 		]
 	]
-	
+
 	encode-reference: func [
 		reference [int-ptr!]
 		payload   [red-binary!]
--- a/runtime/redbin.reds
+++ b/runtime/redbin.reds
@@ -189,7 +189,7 @@ redbin: context [
 	]
 	
 	reference: context [
-		;-- a map of integer! -> red-integer!
+		;-- a map of node! -> offset in 'list'
 		map:  as node! 0								;-- initialized in 'reset'
 		list: as int-ptr! 0								;-- as well
 		top:  list
--- a/runtime/redbin.reds
+++ b/runtime/redbin.reds
@@ -150,13 +150,19 @@ redbin: context [
 	reset: does [path/reset reference/reset]
 	
 	path: context [
-		size:  1'000										;-- arbitrary
-		stack: as int-ptr! allocate size * size? integer!	;-- offset
+		stack: as int-ptr! 0
 		top:   stack
-		end:   stack + size
+		end:   stack
 		
-		push: does [
-			if top + 1 >= end [reset fire [TO_ERROR(internal limit-hit) integer/push path/size]]
+		push: func [/local newsz [integer!] new [int-ptr!]] [
+			if top + 1 > end [
+				newsz: (as-integer end - stack) * 3 / 2				;-- +50% of current size
+				new: as int-ptr! realloc as byte-ptr! stack newsz
+				if null? new [reset fire [TO_ERROR(internal no-memory)]]
+				top: new + (top - stack)
+				end: new + (newsz / size? integer!)
+				stack: new
+			]
 			top/value: offset
 			top: top + 1
 			offset: 0
@@ -168,14 +174,24 @@ redbin: context [
 			offset: top/value
 		]
 		
-		reset: does [top: stack]
+		reset: func [/local min-size [integer!]] [
+			min-size: 1024
+			if (as-integer end - stack) > (min-size * size? integer!) [	;-- free the extra RAM
+				free as byte-ptr! stack
+				stack: as int-ptr! 0
+			]
+			if null? stack [
+				stack: as int-ptr! allocate min-size * size? integer!
+				end: stack + min-size
+			]
+			top: stack
+		]
 	]
 	
 	reference: context [
-		size: 20'000										;-- arbitrary
-		list: as int-ptr! allocate size * size? integer!	;-- node, size, offset * size
+		list: as int-ptr! 0
 		top:  list
-		end:  list + size
+		end:  list
 		
 		fetch: func [
 			node    [node!]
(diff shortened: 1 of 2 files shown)


greggirwin:
@hiiamboris `do https://gitlab.com/hiiamboris/red-mezz-warehouse/-/blob/master/clock.red` doesn't work for me. The URL  returns HTML, not the raw script content.

hiiamboris:
Fixed, thanks

greggirwin:
And because I'm evil...
```
>> b: to block! system
== [
    version: 0.6.4 
    build: make object! [
        date: 31-Jan-2021/20:40:54
        git: make object! [
            branch: ""
   ...
>> save/as clear bin: #{} b 'redbin
*** Access Error: cannot decode or encode (no codec): routine ["Internal Use Only"][bool: as red-logic! stack/arguments bool/header: TYPE_LOGIC bool/value: natives/last-lf?]
*** Where: encode
*** Stack: save  
```

hiiamboris:
Yeah. Pity we can't do that.

hiiamboris:
Yeah. Pity we can't do that.

greggirwin:
It's an interesting question, about how to make it most useful, not knowing yet all the ways we might want to. e.g. can we safely save handles and events as none (maybe zero for handle)? We just say they aren't saved, and why. Or we do save them, but note that they are just data values and not portable across the wire. Maybe still useful for debugging. Need to decide if events go across as just a binary, so they could be decoded and seen elsewhere. That leads to thinking about event streams in logs. 

On routines, there aren't many. Some of those are simply optimizations, but as a general case there could be more and for more reasons. But here's the funny thing about them. We can reflect on them. `Help` shows you their source, so we could, if we wanted, serialize their spec and body just like functions. Have to think about how this plays out in real use cases, where compilation on the decoder's side has to be done to reactivate them as their original, active values.

hiiamboris:
```
toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
travis_time:end:08c16828:start=1612294588457285107,finish=1612295005041375888,duration=416584090781,event=install
[0K[31;1mThe command "if [[ ${OS} == "armhf" ]]; then docker run -v ${PWD}:/red -w /red i386/debian:red rebol -qws tests/build-arm-tests.r '-t RPi' && docker run -v ${PWD}:/red -w /red i386/debian:red rebol -qws system/tests/build-arm-tests.r '-t RPi' && printf 'FROM balenalib/raspberry-pi2-debian:latest\n RUN apt-get update && apt-get install -y libgdk-pixbuf2.0-0 ' > Dockerfile && docker build -t balenalib/raspberry-pi2-debian:red . ; fi
  " failed and exited with 1 during .[0m
  ```

**Comment by greggirwin** [comment]
The details page has a notice:

> Please be aware travis-ci.org will be shutting down in several weeks, with all accounts migrating to travis-ci.com. Please stay tuned here for more information.

@x8x are we prepared for this change?

**Comment by greggirwin** [comment]
Are we OK with merging this @dockimbel @qtxie ?

**Comment by qtxie** [comment]
@greggirwin Yes. I think it's OK.

**Comment by dockimbel** [comment]
@hiiamboris 
I'm not OK with this PR. The introduction of an hashtable to handle references optimizes Redbin codec for large amount of series, but is an absolute overkill for small sets of values. We will soon have full networking support and sending/receiving and using Redbin to encode small messages made of Red values should become common. This PR forces the allocation and usage of an hashtable for even the smallest `string!` or `block!` series, making it a resource-heavy (bigger pressure on the GC) and slower.

The implementation should be smarter and only switch to an hashtable if the number of series is beyond a given threshold, so that small messages are processed lighter and faster.

**Comment by greggirwin** [comment]
I chatted with @dockimbel about this a bit, and I still think it's a step forward from where we were. His priority is maximum efficiency for small messages, but I noted that it adds extra complexity to be size aware. He also noted that the mandatory reference handling for every series, introduced by Vladimir, is already a concern for those small messages use-cases. 

We can profile now, but having ports and stressing it that way will tell us more.

**Comment by greggirwin** [comment]
@dockimbel what do you think of as small payloads, and are there datatype distributions we should consider? e.g. numeric, word, block, or string heavy?

**Comment by greggirwin** [comment]
Data
```
loop 5 [
	repeat i 1 [repend b [i 1 * 1.1 form i  to ref! i to word! append copy "i-" i]]
]
```

Original redbin
```
testing with 5 values
Count: 10'000
Time         | Time (Per)   | Memory      | Code
0:00:00.018  | 0:00:00      | 2401076     | [c: load/as bin 'redbin]
0:00:00.042  | 0:00:00      | 6285272     | [save/as clear bin b 'redbin]

testing with 25 values
Count: 10'000
Time         | Time (Per)   | Memory      | Code
0:00:00.033  | 0:00:00      | 9120920     | [c: load/as bin 'redbin]
0:00:00.074  | 0:00:00      | 17001884    | [save/as clear bin b 'redbin]
```

New redbin
```
testing with 5 values
Count: 10'000
Time         | Time (Per)   | Memory      | Code
0:00:00.019  | 0:00:00      | 2400920     | [c: load/as bin 'redbin]
0:00:00.184  | 0:00:00      | 6281372     | [save/as clear bin b 'redbin]

testing with 25 values
Count: 10'000
Time         | Time (Per)   | Memory      | Code
0:00:00.038  | 0:00:00      | 9120920     | [c: load/as bin 'redbin]
0:00:00.233  | 0:00:00      | 17001884    | [save/as clear bin b 'redbin]
```

Encoding is where we take the hit. Given a P2P scenario, each side pays the same price. In a client server setup, there are more clients who take the hit, but server responses may be much bigger than requests, so there may still be a balance of sorts in many cases. If responses are much larger, which is entirely possible, then the server takes the bigger hit, which is not where you want it, but larger responses are also were you gain the most from this approach. With small payloads, but many of them, the pressure is also distributed. With large payloads the original slowdown effectively blocks the encoding process for a very long time.

On a related topic, Boris didn't like my system/options redbin limit idea, but it's really a design preference question. As is often the case, we just push problems around. On the one hand, a declarative limit is central and locked down tighter by default. Auto-expansion means imperative logic everywhere you handle redbin payloads and potentially more work to make a config option. It does offer granular control where a global setting does not. Which do you prefer @dockimbel?

The best real-world use cases I can think of are very small blocks (e.g. dialected IPC), data transfer (smallish to large, varying widely, and probably either string or numeric heavy), and environment images (the `system` question and what do to about values we can't currently encode, which may have a much more consistent size, using `system` as a baseline for comparison).

**Comment by greggirwin** [comment]
I prefer to avoid pathological cases, and will give other things up for that. The original approach became pathological for large values, which we don't want. But we can't rule out how heavy use might be with small messages. I think here of 0MQ-like applications. We should also consider that taking the data to the processor may not always be best. With Red we can bring the logic to the data where that makes more sense.

**Comment by hiiamboris** [comment]
@dockimbel would you prefer it just switches between hashtable and array based on the references count, or can the codec be designed some totally different way?

**Comment by hiiamboris** [comment]
I love data. So here's some.

How to read the plots:
- on X axis is the saved block length (block of unique strings or block of unique refs)
- on Y axis is the milliseconds this operation took
- 2 variants presented: save to a file (overwrite) and save to memory (#{})

**Linear plot**

| strings | refs |
|---|---|
|  |  |

**Logarithmic plot**

| strings | refs |
|---|---|
|  |  |

**Logarithmic plot zoomed in**

| strings | refs |
|---|---|
|  |  |

Conclusion:
- for file ops there is no win in using array codec
- intersection point is at around 30 strings and 60 refs, apparently because string contains 2 nodes per value and ref - only 1 node per value, so I propose that codec should upgrade the array into hashtable at it's 65th reference (60 aligned to 64, next is 65th)


Script I used to collect the data:
```
Red []
recycle/off
repeat size 1000 [
	recycle
	b: clear []
	; repeat i size [append b to ref! i]				;) ref variant
	repeat i size [append b form i]						;) string variant
	count: to 1 100'000 / size
	; t: dt [loop count [save %test.redbin b]]			;) file variant
	t: dt [loop count [save/as clear #{} b 'redbin]]    ;) ram variant
	print rejoin [size "," count "," t "," t/second * 1000 / count]
]
```

greggirwin:
Faaaaantastic. Since @dockimbel mentioned GC pressure, and it should be GC safe now, can we incorporate that metric somehow?

What did you use to generate the plots? I've lost track of all your tools. ;^)

greggirwin:
Ah, nevermind, I see the manual `recycle` call now.

hiiamboris:
Actually I miss the point of "bigger pressure on the GC". How is it bigger if it uses a static global hash table? Gregg's report shows that the number of memory allocated is identical between two codecs.

greggirwin:
Maybe @dockimbel meant CPU pressure?

:^)

9214:
@hiiamboris sorry for such a late response. Do you still need any feedback from me on that PR?

#### Background

For the record, the initial implementation was intentionally made dumb and simple so that I could test if the design of the new reference sub-system makes sense at all; later the release was rushed out of the door and I didn't have a chance to tidy things up and come up with heuristics for amortization (e.g. double or half the size). In that regard, your changes make sense and provide the finishing touches.

#### Ports

In my experience, Redbin is optimized for fast decoding and long-term persistent storage (i.e. archiving data in files and loading it at runtime), and IMO is sub-optimal for communicating over the wire in networking-heavy apps (it's not streamable and random accessible, there's no bit correction and partial-write recovery, reading/writing cannot be shared between multiple threads, there's no RPC protocol, no validation checks ATM, etc). Saying it to merely point to that there might be a case for a different serialization format.

#### Optimizations

Using different reference sub-systems might make sense, but the problem is that codec cannot know in advance how many nodes are there in your data (i.e. should it use a dumb list or hashtable, and how much space to allot for them?) or how deep the tree of values is (i.e. how much to allot for the stack?). OTOH, both of these parameters can be sniffed out in advance with Parse or somesuch.

If you encode a single value that contains a reference (e.g. `series!` full of `scalar!`s), then relying on hashtable is indeed an overkill — you just need to track a single node in such case. But if it is a deeply nested `any-list!`, especially the one that contains many `any-word!`s (which by necessity need to encode `object!` or `function!` to which they are bound, and that usually pulls the entire environment with them, recursively), then hashtable makes a lot of sense even for seemingly "small" values.

#### Unsupported values

The documentation [clearly states](https://github.com/red/docs/blob/master/en/redbin.adoc#unsupported) why they cannot be encoded. In the future, `routine!`s probably can be JIT-compiled during decoding, but even now you can encode their spec and body separately if you want to.

As for `handle!`s and `events!`, in the current implementation both contain references to resources that are both OS-specific *and* session-specific. Not only passing them between different machines doesn't make much sense (different OSes, different versions of the same OS, etc), but also using the same value in a different Red instance on the same PC is meaningless — at best the user will get a weird behavior, and at worst the program will segfault trying to access a non-existent resource.

Any support for them should come from a protocoling level, e.g. a dedicated View encoder that will be smart enough to re-instantiate handlers depending on the `face!` content and the type of backend it runs on — but then you might as well just give it a VID block to process in a platform-independent way.

greggirwin:
Great info and thoughts @9214. Thanks!

My thought on handles and events was that sometimes the values, just as values, tell you something. For example, a 0 handle tells you something. I don't know of other special cases OTTOMH head though.

For events, the most powerful thing would be the ability to replay them. That may not be possible with the current implementation, so then we'd need to look at how to capture them locally to convert to a replayable event stream.

hiiamboris:
> but also using the same value in a different Red instance on the same PC is meaningless

I will remind that in my view test system I'm streaming whole faces (molded) from one Red instance into another and using their handles. This is possible because View handles are HWNDs and those are valid for any process as long as the referenced OS window lives.

9214:
@hiiamboris I doubt it will be any trouble for you to add codec for handles in a separate fork, but when discussing that with @dockimbel last year he agreed that Redbin should not support them in the current form.

In general, the problem of passing around references that can potentially outlive their resource and become dangling is a complex one and cannot be solved without adding extra constraints (like keeping the OS window live for example). As I said, anything that relies on such assumptions should rely on a protocol that will keep them in check. Personally, I don't mind if there will be a power-user mode as a part of Secure dialect, that will permit Redbin (or any other codec) to do all the nasty things.

`event!` and `handle!` are very limited datatypes tied to Red's GUI system, in fact, you cannot even create them yourself because they don't have `make` actions. OTOH, both can be reified as another value for serialization (e.g. an `object!` and an `integer!`), but I don't think you will be able to instantiate them back at runtime after that — this is something that only the View engine can do internally.

```red
reify: function [
	face  [object!]
	event [event!]
][
	accessors: system/catalog/accessors/event!
	event: collect [
		foreach accessor accessors [
			keep accessor
			keep event/:accessor
		]
	]
	
	probe new-line/skip event on 2
]

view [button :reify]
````

Again, nothing that can't be worked around with a bunch of routines.

hiiamboris:
For event, I also created objects mimicking events, but this solution is quite heavy. And standard actors created by VID have this spec: `[face [object!] event [event! none!]]` and won't accept your imitation. That is, I agree that everything can be worked around, but in doing so you will be fighting language design and it will only work on small scale. So this design question is open for me yet.

9214:
@hiiamboris for `face!` serialization in general, if memory serves me well, there was an issue with `on-change*` that prevented me from encoding reactors https://github.com/red/red/issues/4552.

> Originally this investigation started with `on-change*` that has 4 such instances of `self` in its body.

https://github.com/red/red/blob/01874231c40f5aa93c1a6088309c2ba385e3a884/runtime/redbin.reds#L1528-L1532

greggirwin:
Good chat. 

Are Red-level events waiting for a Red-level `struct!` type? Did you start on `struct!` @9214?

For streaming scenarios the answer is probably to use an envelope.

What we have now is a fairly simple implementation. It was even simpler before, and I'd still like @dockimbel to weigh in on the `system/options` idea. I'm OK with the current approach, just want his thoughts. But if we add block/hash duality and sniffing, is it worth the extra complexity, given @hiiamboris' test results? We can always do that later, as it doesn't change the interface in any way (a downside to `system/options` is that it's directly user facing and could change if we go automatic later rather than now). If we keep it using a hash, the worst thing that happens is the first user who starts seeing performance issues brings it up. It can then be addressed quickly enough, if needed, that it shouldn't tarnish Red's reputation in any way. 

We can also safely say, IMO, that there are unsupported value types. But we shouldn't make the user pre-process their data, removing all those values before encoding with redbin in order to avoid errors. Maybe it ties to construction syntax. I don't like the idea of magic placeholder values or hacking in exceptions for these few types. It's a design question we need to solve in a general way.

hiiamboris:
Dunno why you both want to sniff the data first. I think simply copying the references from the array into the hashtable at a given threshold should be enough.

greggirwin:
I don't want to sniff, and agree it shouldn't be necessary.

PR #4843, opened 2021-02-28
Improve JSON codec error information.

greggirwin:


greggirwin:
The failures here aren't regressions, but Travis failing.

PR #4846, opened 2021-03-03
FEAT: define distance? for pairs

hiiamboris:
Why not? I haven't found a single use of `distance?` in Red codebase. But found myself reinventing the pair distance a few times already.
```
>> distance? object [offset: 0x0 size: 0x0] object [offset: 0x0 size: 1x1]
== 0.7071067811865476
>> distance? 0x0 1x1
== 1.414213562373095
```

An alternative would be to define `(length? A - B) -> float!`.

### DIFF
--- a/modules/view/utils.red
+++ b/modules/view/utils.red
@@ -45,7 +45,7 @@ overlap?: function [
 	"Return TRUE if the two faces bounding boxes are overlapping"
 	A		[object!] "First face"
 	B		[object!] "Second face"
-	return: [logic!]  "TRUE if overlapping"
+	; return: [logic!]  "TRUE if overlapping"
 ][
 	A1: A/offset
 	B1: B/offset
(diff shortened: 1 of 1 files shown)


greggirwin:
I agree that it's a good func to have. I have an old one from R2, which supported pairs or blocks:
```
distance: func [
        {Always returns a positive value.}
        a [pair! block!] "If a block, it must contain two number! values."
        b [pair! block!] "If a block, it must contain two number! values."
    ][
        square-root add (square (a/1 - b/1)) (square (a/2 - b/2))
    ]
```

1) Should it be in view/utils, or is there general value in distances outside View?
2) If it's general-purpose, using `x/y` accessors makes it easy to understand.
3) There's some overlap in meaning with `offset?` for series, but I don't know how best to meld the two without some thought. `offset?` for pairs is easy enough, giving a nice name over `subtract`. But where `offset?` is like an inverse of `skip` for series, there's no inverse in this 2D context.

Center distance between faces is a very specific behavior, but can apply to any bounded area. And what does it look like from a "space between" distance, e.g. as spacing is applied in VID. Or distance from anchor points, which might be the UL corner for a box, but the center of a circle. It's an interesting domain.

greggirwin:
And while we don't have a `geo!` coordinate today, that's a very important use case for `distance?`.

hiiamboris:
Yes, 'length?' won't work for coordinates..

PR #4856, opened 2021-03-16
FEAT: auto copy about text to clipboard

hiiamboris:
imagine how often I'm doing this manually ;)

### DIFF
--- a/environment/console/help.red
+++ b/environment/console/help.red
@@ -603,6 +603,7 @@ help-ctx: context [
 	set 'about func [
 		"Print Red version information"
 		/debug "Print full Red and OS version information suitable for submitting issues"
+		/cc "Also Copy the result into Clipboard"
 		/local git plt txt
 	][
 		git: system/build/git
@@ -632,7 +633,7 @@ help-ctx: context [
 			]
 			txt: form txt
 		]
-		write-clipboard txt
+		if cc [write-clipboard txt]
 		print txt
 	]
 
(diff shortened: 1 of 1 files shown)


greggirwin:
My instinct is that it should be done with a refinement (e.g. `about/cc`). It's absolutely a common use case for us, but when apps copy to the clipboard without my knowledge it can be very frustrating. Almost as much as when I overwrite it myself and have already closed what I had copied.

greggirwin (commented):
See comment about maybe using a refinement for this new behavior.

environment/console/help.red
```diff
@@ -603,6 +603,7 @@ help-ctx: context [
 	set 'about func [
 		"Print Red version information"
 		/debug "Print full Red and OS version information suitable for submitting issues"
+		/cc "Also Copy the result into Clipboard"
```
greggirwin, line 606: Small thing but let's lose the caps mid sentence: "Also copy the result into clipboard"

hiiamboris: Then the name cc become a bit cryptic

greggirwin: Ah, I see how you intended it. Being old, I interpret it as Carbon Copy, as it's used with email.

greggirwin: If we do `/cc "Also copy to clipboard"` it puts the 2 c's closer together. Then the case also matches the refinement case.

greggirwin: And, No, `/c2c` is not acceptable. ;^)

hiiamboris: :D

greggirwin:
Good work. Thanks!

greggirwin:
Please announce this in Gitter, and we should update the issue template to note it.

PR #4857, opened 2021-03-16
FIX: Issue #4165

### LINKED ISSUE #4165: `find/same` doesn't properly find words
**Describe the bug**

`find/same` is supposed to be using `same?` comparator, but when looking for a word it doesn't.
```
>> a: object [x: 1]
>> b: object [x: 2]
>> (in a 'x) == (in b 'x)
== true
>> (in a 'x) =? (in b 'x)
== false                    ;) words aren't "same" if bound to different contexts
>> xs: reduce ['x in a 'x in b 'x]
== [x x x]
>> find/same xs in a 'x
== [x x x]                  ;) should be [x x]
>> find/same xs in b 'x
== [x x x]                  ;) should be [x]
>> find/same xs next xs
== [x x x]                  ;) should be [x x]
>> select/same xs in b 'x
== x                        ;) should be none
```

**To reproduce**
```
a: object [x: 1]
b: object [x: 2]
(in a 'x) == (in b 'x)
(in a 'x) =? (in b 'x)
xs: reduce ['x in a 'x in b 'x]
find/same xs in a 'x
find/same xs in b 'x
find/same xs next xs
select/same xs in b 'x
```

```
Red 0.6.4 for Windows built 27-Nov-2019/10:17:45+03:00 commit #b2aafe6
```

hiiamboris:
fixes #4165

### DIFF
--- a/runtime/datatypes/bitset.reds
+++ b/runtime/datatypes/bitset.reds
@@ -721,6 +721,7 @@ bitset: context [
 		#if debug? = yes [if verbose > 0 [print-line "bitset/compare"]]
 
 		if TYPE_OF(bs2) <> TYPE_BITSET [RETURN_COMPARE_OTHER]
+		if op = COMP_SAME [return as-integer bs1/node <> bs2/node]
 		
 		s: 	  GET_BUFFER(bs1)
 		head: as byte-ptr! s/offset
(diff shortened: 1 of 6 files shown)


hiiamboris:
Of note: `event!` is still silly - every event equals every other event.

greggirwin:
I imagine event comparisons are TBD, since they're OS based. Comes back to making that aspect opaque and wrapping it, but that hasn't been done because there's no efficient solution yet. Need structs in Red. It's kind of like the NaN comparison in the ticket. Technically all NaNs are equal if we look at the bits, but the semantics everybody agrees on is that no two are equal. Dec64 raises this question as well, because there are a number of bit patterns that can represent zero.

Issue #4855, opened 2021-03-16
Dates ignore sub-seconds in comparison


hiiamboris:
**Describe the bug**
```
>> t: now/precise
>> t': t + 0:0:0.01
>> t' <= t
== true
>> t' >= t
== true
```

**Expected behavior**

`t' <= t` to be false

**Platform version**
```
Red 0.6.4 for Windows built 2-Mar-2021/15:01:01+03:00 commit #6de9b93
```

greggirwin:
```
>> t' = t
== true
>> t' <> t
== false
>> t' < t
== false
>> t' > t
== false
```

greggirwin:
Date compare intentionally rounds (%date.reds @Line:1093). @dockimbel will have to say why.
```
		t1: floor value1/time + 0.5						;-- in UTC already, round to integer
		t2: floor value2/time + 0.5
```

hiiamboris:
R2 didn't do that:
```
>> t: now/precise
>> t + 0:0:0.0000000001 = t
== true
>> t + 0:0:0.000000001 = t
== false
```

qtxie:
Pushed a fix in PR #4858.

PR #4858, opened 2021-03-18
FIX: issue #4855 (Dates ignore sub-seconds in comparison)

qtxie:


dockimbel:
@qtxie Is it passing the tests? Because from above report, it seems not.

qtxie:
@dockimbel Add an additional fix.

PR #4868, opened 2021-04-01
Help - add doc string for word arg.

greggirwin:


greggirwin:
Bah! Sorry for the duped JSON commit from the other PR.

Issue #4876, opened 2021-04-13
Dynamically update face/pane crashes view


qtxie:
**Describe the bug**
Run the following script, click on the window repeatedly.
```
Red []
getpanel: function [ i ] [ 
    compose/deep [ panel 200x55 0.30.30 [ ] draw []]
]
view [
    aa: panel 400x600 [ text "click to add panel" ] on-down [ 
        clear face/pane
        repeat n 10 [
            append face/pane layout/only getpanel n 
        ]
    ]
]
```
Crashes:
```
*** Runtime Error 1: access violation
*** at: 004341B7h
```

**Expected behavior**
No crash.

**Platform version**
Red 0.6.4 for Windows built 13-Apr-2021/6:52:09

qtxie:
Related to GC.  No crash with `recycle/off`. No crash in debug mode.

qtxie:
Not related to view backend. This also crashes:
```
loop 1000 [layout/only compose/deep [panel [ ] draw []]]
```

qtxie:
It crashes inside `parse` when parsing empty draw block. It looks like the GC corrupted some states in parse. The parse is happened inside another `parse`, which makes it's even harder to find out which value was corrupted.

dockimbel:
Note: the crash can only be reproduced when run from the GUI console. When compiled, even if the example code is wrapped into a `do [...]` section to force interpretation, no crash occurs.

PR #4878, opened 2021-04-14
FIX: issue #4876 (Dynamically update face/pane crashes view)

qtxie:


qtxie:
`block/rs-head` always returns a value even the block is empty. https://github.com/red/red/blob/master/runtime/parse.reds#L1614

runtime/datatypes/block.reds
```diff
@@ -48,6 +48,7 @@ block: context [
 			s	[series!]
 	][
 		s: GET_BUFFER(blk)
+		assert s/tail > (s/offset + blk/head)
```
dockimbel, line 51: Should be `>=`, as tail can be equal to offset + head if the block is empty.

PR #4884, opened 2021-04-24
FEAT: transparent round/to x 0

hiiamboris:
Following the discussion :point_up: [April 22, 2021 9:27 PM](https://gitter.im/red/help?at=6081c02db6a4714a29ded202) and TG chat.

**Summary**

Rationale: to not create special cases where they aren't required

Example:
```
>> round/to pi 1
== 3
>> round/to pi 1e-10
== 3.1415926536
>> round/to pi 1e-100
== 3.141592653589793
>> round/to pi 1e-500
*** Math Error: math or number overflow
>> round/to pi 0
*** Math Error: math or number overflow
```
As `scale` approaches zero we should expect the result to come closer and closer to the original value.
With the `lim (round/to value scale) = value` when `scale -> 0`.
The zero can be reached in practice as mentioned in the above discussion, so it makes sense to handle it automatically, leaving programmer with less headaches.

This PR implements `x = round/to x 0` invariant, even for special floats:
```
>> round/to pi 0
== 3.141592653589793
>> round/to now/time 0
== 21:10:46
>> round/to 1.#inf 0
== 1.#INF
>> round/to -1.#inf 0
== -1.#INF
>> round/to 1.#nan 0
== 1.#NaN
```

### DIFF
--- a/runtime/datatypes/bitset.reds
+++ b/runtime/datatypes/bitset.reds
@@ -721,6 +721,7 @@ bitset: context [
 		#if debug? = yes [if verbose > 0 [print-line "bitset/compare"]]
 
 		if TYPE_OF(bs2) <> TYPE_BITSET [RETURN_COMPARE_OTHER]
+		if op = COMP_SAME [return as-integer bs1/node <> bs2/node]
 		
 		s: 	  GET_BUFFER(bs1)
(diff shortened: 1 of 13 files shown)


hiiamboris:
hmm that's too many commits...

PR #4885, opened 2021-04-24
FEAT: transparent round/to x 0

hiiamboris:
Following the discussion :point_up: [April 22, 2021 9:27 PM](https://gitter.im/red/help?at=6081c02db6a4714a29ded202) and TG chat.

**Summary**

Rationale: to not create special cases where they aren't required

Example:
```
>> round/to pi 1
== 3
>> round/to pi 1e-10
== 3.1415926536
>> round/to pi 1e-100
== 3.141592653589793
>> round/to pi 1e-500
*** Math Error: math or number overflow
>> round/to pi 0
*** Math Error: math or number overflow
```
As `scale` approaches zero we should expect the result to come closer and closer to the original value.
With the `lim (round/to value scale) = value` when `scale -> 0`.
The zero can be reached in practice as mentioned in the above discussion, so it makes sense to handle it automatically, leaving programmer with less headaches.

This PR implements `x = round/to x 0` invariant, even for special floats:
```
>> round/to pi 0
== 3.141592653589793
>> round/to now/time 0
== 21:10:46
>> round/to 1.#inf 0
== 1.#INF
>> round/to -1.#inf 0
== -1.#INF
>> round/to 1.#nan 0
== 1.#NaN
```

### DIFF
--- a/environment/actions.red
+++ b/environment/actions.red
@@ -148,7 +148,7 @@ round: make action! [[
 		"Returns the nearest integer. Halves round up (away from zero) by default"
 		n		[number! money! time! pair!]
 		/to		"Return the nearest multiple of the scale parameter"
-		scale	[number! money! time!] "Must be a non-zero value"
+		scale	[number! money! time!] "If zero, returns N unchanged"
 		/even		"Halves round toward even results"
 		/down		"Round toward zero, ignoring discarded digits. (truncate)"
 		/half-down	"Halves round toward zero"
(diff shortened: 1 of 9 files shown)


dockimbel:
This PR is changing the meaning of the `/to` refinement for `round`. The current meaning is:

```
REFINEMENTS:
     /to          => Return the nearest multiple of the scale parameter.
        scale        [number! money! time!] "Must be a non-zero value."
```

This PR should at least update the doc string accordingly.

hiiamboris:
Thanks! Fixed!

greggirwin (approved):
All OK by me.

dockimbel:
@qtxie Are you OK with the changes made by this PR?

qtxie:
@dockimbel Yes. I'm OK.

qtxie (approved):

Issue #4889, opened 2021-05-06
Error on Mac version red-05may21-3723789c6


rsheehan:
**Before submitting**
Prior to opening this ticket, please share your finding in our [dedicated chat](https://gitter.im/red/bugs), so we can confirm it and help in reproducing it, including on other platforms.

**Describe the bug**
When red runs, programs work as normal, but on termination the following message appears.
*** Driver Internal Error: Script Error : status needs a value 
*** Where: run-console 
*** Near: [status: either all [Windows? gui?] [ gui-sys-call exe any [all [file form-args file] ""] ]]

**To reproduce**
1. Run the recent Mac version.

**Expected behavior**
No error should be reported after running.

**Screenshots**

**Platform version**
I am running on Mac OS 10.13.6. The red version is red-05may21-3723789c6.
Still a problem in "Red 0.6.4 for macOS built 10-May-2021/14:11:11+12:00  commit #9bae458".

loziniak:
probably caused by #4869 .

loziniak:
https://gitter.im/red/bugs?at=609393d375aeb43be72d64f0 :

> looks like `sys-call` routine is defined without an integer return value
> in a few places even

PR #4902, opened 2021-05-14
FIX: issue #4889 (Driver Internal Error: Script Error : status needs …

loziniak:
…a value)

Not tested locally, as it would require a Rebol licence, if I'm not mistaken. So, hope for the best.

### DIFF
--- a/red.r
+++ b/red.r
@@ -32,7 +32,7 @@ redc: context [
 		temp-dir: switch/default system/version/4 [
 			2 [											;-- MacOS X
 				libc: load/library %libc.dylib
-				sys-call: make routine! [cmd [string!]] libc "system"
+				sys-call: make routine! [cmd [string!] return: [integer!]] libc "system"
 				join any [attempt [to-rebol-file get-env "HOME"] %/tmp] %/.red/
 			]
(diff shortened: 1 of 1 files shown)


red.r
```diff
@@ -83,7 +83,7 @@ redc: context [
 					
 					IsProcessorFeaturePresent: make routine! [feat [integer!] return: [integer!]] kernel32 "IsProcessorFeaturePresent"
 
-					gui-sys-call: func [cmd [string!] args [string!]][
+					gui-sys-call: func [cmd [string!] args [string!] return: [integer!]][
```
hiiamboris, line 86: this I think will error out for `return:`

hiiamboris: and the other `func`

loziniak: Seems probable:
```
% echo "Rebol [] a: func [return: [integer\!]] []" | ./rebol
...
** Script Error: Invalid argument: return
** Where: throw-on-error
** Near: func [return: [integer!]] []
```

loziniak: I deleted `return:`s in `func`s, as they were needed only in `routine`s. Force-pushed the changes.

Issue #4906, opened 2021-05-19
File! does not escape semicolon on MOLD/SAVE


rebolek:
If you have a `file!` with a semicolon in name and `save` it, semicolon is not escaped and such file can’t be loaded back properly as everything after semicolon is ignored.

**To reproduce**

```
>> save %test reduce [to file! "semi;colon" 'ignored]
>> read %test
== "%semi;colon ignored"
>> load %test
== %semi
```

Note: `save/all` or `mold/all` works same.

**Expected behavior**

When semicolon is present in a `file!`, either use percent encoding so `;` will be encoded as `%3B`, or enclose whole filename in quotes, e.g.: `%"semi;colon"`.

**Platform version**

```
Red 0.6.4 for Linux built 13-May-2021/15:09:32
```

PR #4907, opened 2021-05-25
FIX: issue #4906 (File! does not escape semicolon on MOLD/SAVE)

qtxie:


greggirwin (commented):
It would be nice if the magic tables were commented. They are completely opaque to humans, how they're used, and how to maintain them.

dockimbel:
I will add the needed comments.

PR #4916, opened 2021-06-27
Addition of base to create-focus-events

gltewalt:
so code such as this will function:

```red
view [
    size 300x300
    toggle red on-created [set-focus face] on-key [
        switch event/key [
            up    [face/offset/y: face/offset/y - 5]
            down  [face/offset/y: face/offset/y + 5]
            left  [face/offset/x: face/offset/x - 5]
            right [face/offset/x: face/offset/x + 5]
        ]
    ]
]
```


gltewalt:
Mistake

PR #4920, opened 2021-07-02
FIX: reset system/thrown before capturing call stack in error!.

qtxie:
error/capture may trigger GC. free-virtual will check system/thrown.
In this case, `system/thrown` is RED_THROWN_ERROR when free-virtual is called.

PR #4921, opened 2021-07-03
FIX: temporarily fix a GC issue in parse.

qtxie:
In function parser/process, `input` points to the buffer of `series` block. When `series` block expanded, `input` still points to the old slot, which is OK. But if GC triggered later, the memory may be overwrote by the GC.

### DIFF
--- a/runtime/parse.reds
+++ b/runtime/parse.reds
@@ -1886,7 +1886,7 @@ parser: context [
 	]
 
 	init: does [
-		series: block/make-in root 8
+		series: block/make-in root 40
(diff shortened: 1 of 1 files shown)


qtxie:
Fix issue: https://github.com/red/red/pull/4920#issuecomment-872939846

dockimbel:
@qtxie  Was the related GC issue fixed by another commit, so that we can close this PR?

qtxie:
Yes. It's fixed by commit e03c49abe2e03ee50fd031be018c2fa5fd4a0421.

Issue #4863, opened 2021-03-24
PARSE binary includes spaces into datatype match


hiiamboris:
**Describe the bug**
```
>> parse to #{} "   word" [word!]
== true
>> parse to #{} "   word" [copy w word! (probe to "" w)]
"   word"
== true
>> parse to #{} "   123" [copy i integer! (probe to "" i)]
"   123"
== true
>> to integer! "   123"
*** Script Error: cannot MAKE/TO integer! from: "   123"
*** Where: to
*** Stack:  
```

**Expected behavior**

Should skip leading spaces.
```
>> parse to #{} "   word" [word!]
== false
>> parse to #{} "word" [word!]
== true
```

**Platform version**
```
Red 0.6.4 for Windows built 18-Mar-2021/22:28:01+03:00  commit #73d9bd2
```

meijeru:
The expected behaviour is explicitly stated in section 2.3 of the Parse doc:
"binary!: matching by datatype or type set is supported for UTF-8 encoded values; such match succeeds if matched portion of the input represents one of the datatype’s literal forms. _Blank characters before tokens are automatically skipped_."

hiiamboris:
I think this is:
1. A source of confusion: you expect `integer!`, whereas the result is not directly convertible 
2. Less practically valuable than extracting the exact datatype match

I understand this is a quirk of current design, but I don't know if it can't be improved.
I'm okay with marking this a wish, or closing with an explanation why the design can't be better.
Or maybe `to` converters should be more forgiving and skip spaces?

greggirwin:
@hiiamboris do we have a `to/load/transcode` table which compares them? In the case of `to` the behavior can be specific in how each type converts to another. We have https://github.com/red/red/wiki/%5BDOC%5D-%60to%60-vs-%60make%60, which only talks about logic values right now, which is a subtlety people can easily trip over.

Why are you not in favor of `to integer!` ignoring spaces in strings? If we argue that it should be a valid integer based on the optimized `tokenizer/scan-integer` implementation, what about this?
```
>> to integer! "123.45"
== 123
```

greggirwin:
And we have other cases, which makes me think we need to address this in general. e.g. why are leading spaces allowed for pairs, but not trailing spaces?
```
>> to pair! " 1x2"
== 1x2
>> to pair! " 1x2 "
*** Script Error: cannot MAKE/TO pair! from: " 1x2 "
*** Where: to
*** Stack:  
```

[...]

PR #4923, opened 2021-07-03
FIX: 4900 (fmod platform discrepancy leads to modulo ambiguity)

### LINKED ISSUE #4900: `fmod` platform discrepancy leads to modulo ambiguity
**Describe the bug**
```
Red/System []
probe fmod 123.0 1.0 / 0.0
```
Result on Windows:
```
-1.#IND.0
```
On Linux:
```
123.0
```

So it gets carried into Red code:
```
>> 123 % 1.#inf
== 1.#NaN        ;) Windows
== 123.0         ;) Linux
```

**Expected behavior**

I think Linux result makes more sense. Whatever the dividend is, infinite divisor is still bigger than it, so the remainder should be the dividend.

**Platform version**
```
Red 0.6.4 for Windows built 12-May-2021/6:12:03+03:00  commit #cebd5b9
Red 0.6.4 for Linux built 11-May-2021/6:20:25+03:00  commit #81d0416
```

hiiamboris:
Fixes #4900 by following GCC semantics for x % infinity.

Also removes dead code.

### DIFF
--- a/runtime/datatypes/float.reds
+++ b/runtime/datatypes/float.reds
@@ -241,12 +241,11 @@ float: context [
 				]
 			]
 			OP_REM [
-				if any [left = -INF left = +INF right = 0.0][return QNaN]	;-- issue #4574
-				either all [0.0 = right not NaN? right][
-					fire [TO_ERROR(math zero-divide)]
-					0.0									;-- pass the compiler's type-checking
-				][
-					fmod left right
+				case [
+					0.0 * right = 0.0 [fmod left right]	;-- finite right part - no special case
+					NaN? right [QNaN]
+					0.0 * left = 0.0 [left]				;-- finite % +-infinity - special case for #4900
+					true [QNaN]							;-- +-infinity % +-infinity = NaN
 				]
 			]
 			default [
--- a/runtime/datatypes/float.reds
+++ b/runtime/datatypes/float.reds
@@ -245,7 +245,7 @@ float: context [
 					0.0 * right = 0.0 [fmod left right]	;-- finite right part - no special case
 					NaN? right [QNaN]
 					0.0 * left = 0.0 [left]				;-- finite % +-infinity - special case for #4900
-					true [QNaN]							;-- +-infinity % +-infinity = NaN
+					true [QNaN]							;-- (+-infinity or NaN) % +-infinity = NaN
 				]
 			]
 			default [
--- a/tests/source/units/regression-test-red.red
+++ b/tests/source/units/regression-test-red.red
@@ -3009,6 +3009,11 @@ comment {
 		--assert not all-equal?4205
 		unset [anded4205 last-random4205 all-equal?4205]
 	
+	--test-- "#4260"
+		catch [throw 1]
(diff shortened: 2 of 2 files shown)


dockimbel:
@qtxie Please review this PR.

qtxie (approved):

dockimbel:
I get an error in the tests suite after merging this PR in master:
```
Compiling C:\Dev\Red\tests\source\units\auto-tests\run-all-comp2.red ...
...using libRedRT built on 22-Jul-2021/17:44:39
...compilation time : 2681 ms

Target: MSDOS 

Compiling to native code...
*** Compilation Error: argument type mismatch on calling: red/float/push64 
*** expected: [integer!], found: [struct! [
        header [integer!] 
        data1 [integer!] 
        data2 [integer!] 
        data3 [integer!]
    ]] 
*** in file: %/C/Dev/Red/tests/source/units/auto-tests/run-all/regression-test-red.red 
*** at line: 363 
*** near: [
    stack/unwind 
    natives/nan?* true 
    stack/unwind 
    f_--assert 
    stack/unwind 
    stack/reset 
    stack/mark-func
]
```
@hiiamboris Can you reproduce it?

hiiamboris:
Found it:
```
Red [] -1.#nan
```
`red -c 1.red`:
```
Compiling to native code...
*** Compilation Error: red/float/push64 is missing an argument
*** in file: %/D/devel/red/red-src/red/1.red
*** at line: 353
*** near: [#user-code]
```

qtxie:
`-1.#nan`, I think the lexer should report it as a syntax error.

dockimbel:
I've let both compiler and runtime lexer accept `-1.#nan` and normalize it as `1.#NaN`. Does any negative NaN exists in IEEE-754 standard? Do other mainstream languages have a concept of negative NaN?

hiiamboris:
https://stackoverflow.com/questions/21349847/positive-vs-negative-nans
We should just ignore the sign and treat all nans equally, and mold as just `1.#NaN`.

dockimbel:
Then it's how it works now (after my yesterday's fix).

greggirwin:
I agree with @qtxie that it should be an error. We're talking about the lexical form here, not the internal representation of NaN, right? What value does it add to let people write it as negative, if it has no meaning?

hiiamboris:
[Postel's Law?](https://github.com/red/red/issues/4863#issuecomment-820621919) ;)

greggirwin:
It has its costs. Everything we allow is something we can't disallow later. Postel's Law is nice when you're talking about some things, absolutely (e.g. variants of protocols or formats). In the other ticket, note that I added "That doesn't mean being too liberal of course.", just like the old axiom "Be open minded, but not so open that your brains fall out."

So I'll ask my question again: what value does it add, when it *looks* like it has meaning, but doesn't?

hiiamboris:
I'm not against forbidding it either.

dockimbel:
So if all are in agreement to forbid that lexical form, I'll do the change.

Issue #2125, opened 2016-07-19
switch chokes on datatypes


WiseGenius:
When checking a value equal to `block!`, `switch` errors with a strange message:
```
red>> switch/default block! [
[        #[block!] [print "yes"]
[    ][print "no"]
*** Script Error: author has no value
*** Where: switch
```

When checking a value equal to any other datatype except `datatype!`, `switch` fails to match it:
```
red>> switch/default word! [
[        #[word!] [print "yes"]
[    ][print "no"]
no
```

But hang on. What's this?
```
red>> switch/default integer! [
[        #[integer!] [print "yes"]
[        1 [print "one"]
[        2 [print "two"]
[    ][print "no"]
one
```

I could maybe understand the other way around, but:
```
red>> switch/default 1 [
[        #[integer!] [print "yes"]
[    ][print "no"]
no
```

And then there's this:
```
red>> switch/default string! [
[        #[string!] [print "yes"]
[        1 [print "one"]
[        2 [print "two"]
[        "Surely not!" [print "Oh dear!"]
[    ][print "no"]
Oh dear!
```

And even this:
```
red>> switch/default block! [
[        [a] [print "No way!"]
[    ][print "no"]
No way!
```

dockimbel:
SWITCH is using FIND internally, so the issue is FIND not matching the datatype! properly.

WiseGenius:
Should this result be expected?
```
red>> switch/default integer! [
[        1 [print "one"]
[        2 [print "two"]
[        #[integer!] [print "yes"]
[    ][print "no"]
one
```
Again, I could maybe understand the other way around, but still:
```
red>> switch/default 1 [
[        #[integer!] [print "yes"]
[    ][print "no"]
no
```
And then there's this:
```
red>> switch/default string! [
[        1 [print "one"]
[        2 [print "two"]
[        "Surely not!" [print "Oh dear!"]
[        #[string!] [print "yes"]
[    ][print "no"]
Oh dear!
```
And still this:
```
red>> switch/default block! [
[        [a] [print "No way!"]
[    ][print "no"]
No way!
```
And should this be happening?
```
red>> switch/default [print "a"] [
[        [a] [print "a"]
[        [b] [print "b"]
[    ][print "no"]
*** Script Error: b has no value
*** Where: switch
```
And therefore this strange value again?
```
red>> switch/default [print "b"] [
[        [a] [print "a"]
[        [b] [print "b"]
[    ][print "no"]
*** Script Error: author has no value
[...]

PR #4924, opened 2021-07-05
FIX: Issue 4911 (FIND type! vs FIND typeset! inconsistency)

### LINKED ISSUE #4911: FIND type! vs FIND typeset! inconsistency
**Describe the bug**

`find/only` disables special treatment of types, but not of typesets:
```
>> find/only reduce [1 integer! 2] number!
== [1 integer! 2]            ;) expected: `== none`
>> find/only reduce [1 integer! 2] integer!
== [integer! 2]
```

**Expected behavior**

R2 returns 1st result in both cases. But I would prefer Red returned `none` in 1st case. Apart from `/only` there is no other way I know of to turn off this special treatment, and obviously there are cases where you wanna just find a value without tons of workarounds. `/only` also should leverage hashtable's fast lookups for this case.

It is also [proposed](https://gitter.im/red/bugs?at=60c1282481b45810b894342f) that docstring for `/only` is changed to "Treat series, datatype, and typeset search values as singular.", and docstring for `value` be provided.

**Platform version**
```
Red 0.6.4 for Windows built 12-May-2021/6:12:03+03:00  commit #cebd5b9
```

hiiamboris:
Fixes #4911
Adds tests for #2125
Removes some `switch` kludges made obsolete by commit https://github.com/red/red/commit/32d353efea0fcb53660db88a702a92b315a5168b
`find [#[integer!] 1] integer!` now matches `1` only, not the datatype (for less surprises and consistency with typesets)
`find/only` also enables the use of hash lookups for datatypes and typesets

### DIFF
--- a/environment/actions.red
+++ b/environment/actions.red
@@ -291,10 +291,10 @@ copy: make action! [[
 find: make action! [[
 		"Returns the series where a value is found, or NONE"
 		series	 [series! bitset! typeset! port! map! none!]
-		value 	 [any-type!]
+		value 	 [any-type!] "Typesets and datatypes can be used to search by datatype"
 		/part "Limit the length of the search"
 			length [number! series!]
-		/only "Treat a series search value as a single value"
+		/only "Treat series and typeset value arguments as single values"
 		/case "Perform a case-sensitive search"
 		/same {Use "same?" as comparator}
 		/any  "TBD: Use * and ? wildcards in string searches"
--- a/runtime/datatypes/block.reds
+++ b/runtime/datatypes/block.reds
@@ -858,6 +858,7 @@ block: context [
 			op		[integer!]
 			type	[integer!]
 			stype	[integer!]
+			dt?		[logic!]
 			ts?		[logic!]
 			found?	[logic!]
 			hash?	[logic!]
@@ -917,9 +918,11 @@ block: context [
 		]
 
 		type: TYPE_OF(value)
+		dt?: all [not only? type = TYPE_DATATYPE]		;-- /only disables special treatment
+		ts?: all [not only? type = TYPE_TYPESET]
 		if any [
-			type = TYPE_DATATYPE
-			type = TYPE_TYPESET
+			dt?
+			ts?
 			all [not same? type = TYPE_OBJECT]
 		][hash?: no]									;-- use block search
 		any-blk?: ANY_BLOCK?(type)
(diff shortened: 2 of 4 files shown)


greggirwin:
Good work. Let's see what @dockimbel thinks about `/same`, per my comment above.

greggirwin (commented):
Thanks for adding `switch` tests as well. :+1: Once we hear Doc's thoughts on `/same`, the rest looks good to me.

hiiamboris:
Point is, `1 = 1`, `1 == 1`, `1 =? 1` are all true. `1 =? integer!` is not a valid analogy.

hiiamboris:
Failed tests are not from this PR btw.

greggirwin:
First we have the `/same` docstring: "Use "same?" as comparator." Then we have the behavior.
```
>> same? 1 integer!
== false
>> same? integer! integer!
== true
```
To have `/same` _not_ match those is completely unintuitive.

Without any refinement, `find` is in its least strict mode. `/only` makes it a bit more strict, `/case` stricter still, and `/same` makes it the most strict. Once we pass a point of maximum strictness for a given type, you just get the same result. At least until a Red Analyzer says "I notice you're `find`ing in a literal, flat block of integers. Consider removing the `/same/case/only refinements from your call."

hiiamboris:
So following this logic let's disable type/typeset matching completely, as:
```
>> strict-equal? 1 integer!
== false
>> strict-equal? integer! integer!
== true
>> equal? 1 integer!
== false
>> equal? integer! integer!
== true
```

greggirwin:
> So following this logic let's disable type/typeset matching completely, as:

If we argue this way, we'll remove all useful behavior, next being type matching in `parse` in this context.

> I think we don't want yet another special case to deal with, that's all.

What is the specific special case you're talking about?

What we're talking about here is _`find`'s_ behavior. @meijeru can weigh in on the correctness and consistency side, but the goal is useful, unsurprising, and correct behavior. Consistency is great too, because that eases explanations.

`/only` is an added constraint, which conveniently (for my note there) qualified as more strictness, and _is_ also type related.

hiiamboris:
> What is the specific special case you're talking about?

Making /same different in behavior from /case and normal equality test is a special case.
I think you're reasoning from a wrong mental model.
It should be:
- /same, /case, and (no /same or /case) are on par and determine the comparison operator
- /only disables special cases: block in block, string in string, type in block, typeset in block

And that's all there is to it.
Docstrings align with this:
```
     value        [any-type!] {Typesets and datatypes can be used to search by datatype.}

REFINEMENTS:
     /only        => Treat series and typeset value arguments as single values.
     /case        => Perform a case-sensitive search.
     /same        => Use "same?" as comparator.
```

greggirwin:
For some reason I can't see/find my original comments in github (must have been a review comment, but nothing's there now), which is inconvenient as it set the context for everything that came after. I have to imagine is what that `/same`  should work like `/only` for types and typesets. @hiiamboris prefers it to have no effect when the search value is a type or typeset. As he notes `/same` determines the comparison operator. To me that implies `same?` and should match that behavior.

Ultimately, it's about which way we lean, and what matches the user's intent. If `/same` works like only, they are redundant and lean us toward 2 ways of matching type/typset exactly. If `/same` is a no-op in those cases, we lean the other way and have 2 ways to find by matching types and typesets. Either way, there is overlap. If `/same` becomes a no-op, we have to change its doc string too.

dockimbel:
This PR and related ticket are so confusing that I've merged them expecting the changes to do the opposite of what I thought they were doing...

greggirwin:
This is merged now, but @dockimbel would you please add your thoughts on `/same`?

greggirwin:
@dockimbel ping.

PR #4959, opened 2021-09-09
FIX: (Update docstring on `open`.)

RayMPerry:
(Feel free to absorb this into the next `port!` commit instead of merging now.)


dockimbel:
Why do you think it was a typo?

PR #4965, opened 2021-09-16
Remove unrelevant comment left from early days.

loziniak:


dockimbel:
You've misunderstood the context of that comment. It's Rebol code, the `sys-call` is mapped to the C `system()` function to avoid using the Rebol native `call` that is quite buggy in Rebol 2.7.x.

loziniak:
Apparently, it's mapped elsewhere, not in commented line.

I tracked the comment down to 2013 and it seems it originally referred to something else and had more sense:
https://github.com/red/red/blame/874b9a8a21f2ab3823b35ba3770aec42cd052959/red.r#L148

PR #4968, opened 2021-09-18
Improve view

qtxie:
Using hardware acceleration causes problems on some systems. For example, the Edge browser and the Red gui-console are both flickering on my Win10. Turn off the hardware acceleration fix it.


dockimbel:
> For example, the Edge browser and the Red gui-console are both flickering on my Win10. 

Could it be just an issue in the graphic card driver?

dockimbel:
The changes you've made in %style.red seems to be Windows-specific, while those definitions should be cross-platform...

hiiamboris:
Shouldn't default height be inferred from the system font height?

qtxie:
> 
> 
> > For example, the Edge browser and the Red gui-console are both flickering on my Win10.
> 
> Could it be just an issue in the graphic card driver?

Yes. I think so. I have updated to the latest driver and installed all the updates for Windows, but no help.

dockimbel:
> Shouldn't default height be inferred from the system font height?

We need to switch to a dynamic coordinate system, but that's a pretty big design and refactoring effort that I was planning to do along the work on Android backend.

qtxie:
> Using hardware acceleration causes problems on some systems. For example, the Edge browser and the Red gui-console are both flickering on my Win10. Turn off the hardware acceleration fix it.

Close it now. Looks like only my desktop has this issue, no other users complained about it.

PR #4979, opened 2021-10-03
Removed symbols for command shells. 

gltewalt:
The symbols `$` are pasted to a users command line if the auto-copy icon on the web page is clicked.  Will cause an error.

### DIFF
--- a/environment/console/GUI/core.red
+++ b/environment/console/GUI/core.red
@@ -177,7 +177,7 @@ object [
 		either s [
 			cnt: 0
 			if all [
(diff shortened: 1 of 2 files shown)


gltewalt:
Consider the novice clicking on the "copy line" icon, pasting to the console or to the terminal, and then seeing an error.
It isn't greatly important that it be changed, I was just thinking of the newbie. Admittedly, I don't know how often that icon is clicked vs. clicking and dragging with the mouse to highlight and copy the text.

PR #4995, opened 2021-11-12
FIX: Issue #4994 (CRASH during transcode of a long string)

### LINKED ISSUE #4994: CRASH during transcode of a long string
**Describe the bug**
```
*** in file: /D/devel/red/common/unicode.reds
*** at line: 159
***
***   stack: red/unicode/cp-to-utf8 167 078FAFFFh
***   stack: red/unicode/to-utf8-buffer 04989D9Ch 078FAFFFh 1000002
***   stack: red/lexer/scan-alt 04989DACh 04989D9Ch 1000002 false true true false 00FFF2C8h 00000000h 00000000h
***   stack: red/natives/transcode* false -1 -1 -1 -1 -1 -1 -1
***   stack: red/interpreter/eval-arguments 04989D8Ch 0569FB88h 0569FB88h 00000000h 00000000h 04AB1418h
***   stack: red/interpreter/eval-code 04AB1418h 0569FB68h 0569FB88h false 00000000h 00000000h 04AB1418h
***   stack: red/interpreter/eval-expression 0569FB68h 0569FB88h false false false
***   stack: red/interpreter/eval 04989D8Ch true
***   stack: red/natives/do* false 1 -1 -1
(...)
```

**To reproduce**

Run this code in console:
```
save %123 append/dup "" "§" 1000000
transcode read %123
```

**Platform version**
```
Red 0.6.4 for Windows built 6-Nov-2021/19:36:18+03:00  commit #30e5312
```

hiiamboris:
Fixes #4994

Upper size estimate for UTF-8 strings was invalid:
 https://github.com/red/red/blob/28b3c83a890eeed86ae90447fe8bf314584ad021/runtime/lexer.reds#L2447

Here's a map of char widths:

| range | bytes in UTF-8 | size in Latin1 | size in UCS-2 | size in UTF-16 |
|-|-|-|-|-|
| 0-7F | 1 | 1 | 2 | 2 |
| 7F-FF | 2 | 1 | 2 | 2 |
| 100-7FF | 2 | N/A | 2 | 2 |
| 800-FFFF | 3 | N/A | 2 | 2 |
| 10000-10FFFF | 4 | N/A | N/A | 4 |

From: https://unicode-org.github.io/icu/userguide/icu/unicode.html

Also, there's a recurring pattern `size << (unit >> 1)` in the runtime code. No idea where it comes from, but it's wrong. Maybe `unit` was a power of two at some point.

### DIFF
--- a/runtime/allocator.reds
+++ b/runtime/allocator.reds
@@ -1059,7 +1059,8 @@ alloc-codepoints: func [
 	unit	[integer!]
 	return: [int-ptr!]						;-- return a new node pointer (pointing to the newly allocated series buffer)
 ][
-	if zero? size [size: 16 >> (log-b unit)]
+	assert unit <= 4
+	if zero? size [size: 16 >> (unit >> 1)]
 	alloc-series size unit 0				;-- optimize by default for tail insertion
 ]
 
--- a/runtime/allocator.reds
+++ b/runtime/allocator.reds
@@ -1059,7 +1059,7 @@ alloc-codepoints: func [
 	unit	[integer!]
 	return: [int-ptr!]						;-- return a new node pointer (pointing to the newly allocated series buffer)
 ][
-	if zero? size [size: 16 >> (unit >> 1)]
+	if zero? size [size: 16 >> (log-b unit)]
 	alloc-series size unit 0				;-- optimize by default for tail insertion
 ]
 
--- a/runtime/datatypes/char.reds
+++ b/runtime/datatypes/char.reds
@@ -112,7 +112,7 @@ char: context [
 				ser: as red-series! spec
 				s: GET_BUFFER(ser)
 				unit: GET_UNIT(s)
-				p: (as byte-ptr! s/offset) + (ser/head << (unit >> 1))
+				p: (as byte-ptr! s/offset) + (ser/head << (log-b unit))
 				if p >= as byte-ptr! s/tail [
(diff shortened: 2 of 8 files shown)


dockimbel:
> Also, there's a recurring pattern size << (unit >> 1) in the runtime code. No idea where it comes from, but it's wrong. Maybe unit was a power of two at some point.

`unit` value range is: 1, 2, 4, 8, 16. It represents the atomic size of items in a Red series!.

hiiamboris:
e.g. size = 100 items, unit = 16 bytes
total bytes: 100 << (16 >> 1) = 100 << 8 = 25600

qtxie:
The `unit` value range is `1, 2, 4` for string related functions. So the code `size << (unit >> 1)` is correct in such cases. I think put an assert in those functions is better. `assert unit <= 4`.

hiiamboris:
Thanks! Indeed, for the [1 2 4] set it just magically works. In such case a clever trick ;)
I've rolled log-b's back.

greggirwin:
As a pattern, without comments to note `[1 2 4]` limitation, this could easily happen again. e.g. people look at string code and recreate the pattern. A few of us know about it now, but since this can cause serious issues, should that pattern be a macro itself, in one place, with an assert and comment? Maybe even _why_ it magically works for that subset. :^)

PR #5015, opened 2021-12-22
FEAT: CLI support

hiiamboris:
docs: https://github.com/red/docs/pull/266


hiiamboris:
By the way, there are some [test scripts](https://gitlab.com/hiiamboris/red-cli/-/tree/master/mockups/tester) I use to detect regressions. They creates random subsets of a given entry function, then a lot of random calls to each subset. And then batch files test if the output changed from the previous version of the library. I didn't commit these. Should I?

dockimbel:
Thanks for this PR! I will review it in details and test that new features in the next days.

modules/cli.red
```diff
@@ -0,0 +1,947 @@
...
+	]
+
+	unary?: function [
+		"Check if option R is unary in F's spec"
+		f [function!] r [string!]
+	][
+		r: str-to-ref r
+		spec: find-refinement prep-spec :f r
+		all [
+			word? pick spec 4
```
greggirwin, line 172: Can these magic numbers be named?

modules/cli.red
```diff
@@ -0,0 +1,947 @@
...
+		value [string! none!]	;-- none allowed if argument is nullary
+		/options				"(placeholder for future expansion)"
+			opts	[block! map! none!]
+	][
+		ref: str-to-ref name
+		spec: find-refinement prep-spec :prog ref
+		refs: first spec
+		#assert [not empty? refs]
+		
+		either pos: find/tail call/3 ref [				;-- are refinements already added?
```
greggirwin, line 323: Another magic number. There are a few more as well, which I won't call out in detail. :^)

greggirwin: That is, `call/3` seems pretty important, used a lot, but I have no idea what it refers to.

modules/cli.red
```diff
@@ -0,0 +1,947 @@
...
+							default?: do is-help-version?					;-- try default arguments too
+							complain [ER_OPT "Unsupported option:" arg]
+						]
+					])
+					[	if (all [
+							not default?								;-- disable unary clause for internal help/version
+							unary? get/any program arg-name				;-- read the value (if required)
+						])
+						[	dlms											;-- require a delimiter
+						|	pos: (complain [ER_CHAR "Unsupported option character at" pos])
```
greggirwin, line 541: I like `complain`. Good word.

modules/cli.red
```diff
@@ -0,0 +1,947 @@
...
+
+
+	;-- used to signal an error in supplied command line
+	complain: func [about [block!]] [
+		;@@ TODO: print to stderr instead (not supported by Red yet)
+		#assert [word? first about]
+		throw/name reduce about 'complaint
+	]
+
+	else: make op! func [truth e [string!]] [unless truth [do make error! e]]
```
greggirwin, line 62: This makes the call sites a bit clunky IMO. Most use `form rejoin`, and the `form reduce` call could as well, just adding spaces to the literal strings. Otherwise there's one literal string used. This could easily hide that detail, so all calls just pass either a string or a block, knowing it's `rejoined`. 

If this were a simple `err-if` func, rather than an op, it would a) make it clear that those calls are potential errors right away, b) eliminate the need for parens around the left arg. It does reverse the logic with that name, but I think that will make things clearer overall. e.g. `pos else ...` becomes `err-if none? pos ...`

hiiamboris: IDK why there's a `form rejoin` :) Should be just `rejoin`... Will fix

modules/cli.red
hiiamboris: IIRC we have already raised this topic:
https://github.com/red/red/blob/7a3e634a30984e3e9ac02409ed8088e8e38a1763/modules/cli.red#L301-L306

modules/cli.red
hiiamboris: This too is described:
https://github.com/red/red/blob/7a3e634a30984e3e9ac02409ed8088e8e38a1763/modules/cli.red#L115-L117

I guess I could do away with macros as I did in `reactivity` PR, but then [you were against macros](https://github.com/red/red/pull/4529#discussion_r641719121)

modules/cli.red
greggirwin: Good, so there's a precedent. :^)

modules/cli.red
greggirwin: 1) What do macros have to do with it? 
2) Having the structure described is critical. +1 for that.
3) They're still magic numbers. So at the call site I have to have the offsets into those structures in my head and mentally map them. If they're not in my head, I have to leave the call site to understand what the call is touching.

But we can come at it from the other direction.

What is the argument _for_ using magic numbers?

hiiamboris: macros are the only constants other than numbers

hiiamboris: my only argument is laziness
you have advocated magic number based XML format all along and now nitpicking about a few indexes that you were okay with during the previous review :)

greggirwin: I'm not saying to change the format. That's fine, as it is with XML. What I'm saying is that `call/3`  is far less meaningful than `call/:TYPE`. I thought this was clear in item 3, as clearly talking about the call site.

Every review is done independently, and the reviewer will see something different every time. It would be great if we had real code reviews, and I felt I could take time to do them in depth. As it is, I do the best I can, and try to make sure people know I'm at least _trying_ to help. :^)

modules/cli.red
hiiamboris: `else` isn't always called with `rejoin`, so `rejoin` stays

modules/cli.red
hiiamboris: updated

modules/cli.red
greggirwin: If `rejoin` is always used with a block, otherwise it's already a string. the helper can just decide to use the string it's given directly, or `rejoin`.

Any thoughts on the readability aspect?

greggirwin (approved):
Thanks for changing magic numbers to names. :+1:

modules/cli.red
hiiamboris: Hmm IIRC I intentionally left `rejoin` explicit, because I'll forget what it's doing with blocks.

environment/console/CLI/console.red
```diff
@@ -8,6 +8,7 @@ Red [
 		Distributed under the Boost Software License, Version 1.0.
 		See https://github.com/red/red/blob/master/BSL-License.txt
 	}
+	Needs:  CLI
```
dockimbel, line 11: Why is this inclusion needed in Red consoles? What is the use-case for the CLI module there?

hiiamboris: And why is `View` module needed there? So users can use all modules without compiling the console themselves.

hiiamboris: OTOH why not? Why should every contributor keep multiple copies of console? E.g. I never use `console.red` and I have to keep `console-view.red` (the one I'm always using but it's not in the repository). If we remove CLI from here, we should remove CLI from gui-console too for consistency, so that means I will have both `console-view-cli.red` and `gui-console-cli.red` scripts. What's the use of those "bare" consoles? Why should they be by default different than the one `red.exe` builds?

greggirwin: As one who likes to play in the console, because it's the most interactive way to work with Red right now, I can see mocking up CLI inputs there and using `/args` to experiment and assist in creating docs for apps with real samples. The consoles should be consistent in that regard.

dockimbel: > and assist in creating docs for apps with real samples

I'm not sure to understand what you mean by that.

> E.g. I never use console.red and I have to keep console-view.red (the one I'm always using but it's not in the repository)
>  What's the use of those "bare" consoles? Why should they be by default different than the one red.exe builds?

I don't understand what you are talking about there. We have a CLI and a GUI console. You can compile the GUI console to run from the shell on Windows by compiling using `-d` instead of `-t Windows`. Here are the command-lines I'm using for compiling those consoles manually on a daily basis:
```
1) do/args %red.r "-r -d %environment/console/CLI/console.red"                ; shell only, debug mode
2) do/args %red.r "-r -d %environment/console/GUI/gui-console.red"            ; shell + GUI, debug mode
3) do/args %red.r "-r -t Windows %environment/console/GUI/gui-console.red"    ; GUI only
```
The ones that `red.exe` builds are 1) (without the `-d`) and 3).

I also keep some extra versions (just cloning `console.red`), like the one I use for C3 development that includes the `eth://` port by default.

--- 
I'm sorry but I'm not convinced at all. This CLI module is for compiled executables only. Compilation is not a feature of the console(s), unless I have missed something in how the CLI module should be used. That is why I'm questioning the default inclusion of that module.

Personal convenience of the contributor is not a criterion for deciding on which module should be available by default. General user benefit is. As I see it, I think few people will use this CLI module and even less will bother experimenting from the console instead of including it in their script and compiling it directly for testing. Moreover, it just takes a `do %<path>/cli.red` in order to load it in the console for eventual experimentation. So I really don't see the value of including it by default, while it will just slower every console compilation (1000 LOC of Red code compiled is not unsignificant) and expand its size for nothing. 

For the needs of static/dynamic code library inclusion in Red, we need to design the module system first.

hiiamboris: I propose:
1. Add `console-view.red` of the following contents:
```
Red [
	Title:	"Red console"
	Author: ["Nenad Rakocevic" "Kaj de Vos"]
	File: 	%console-view.red
	Rights: "Copyright (C) 2012-2018 Red Foundation. All rights reserved."
	License: {
		Distributed under the Boost Software License, Version 1.0.
		See https://github.com/red/red/blob/master/BSL-License.txt
	}
    Needs: [View CLI]
]

#include %console.red
```
2. Rename `gui-console.red` to `gui-console-min.red` so it compiles a few seconds faster.
3. Create `gui-console.red` of the following content:
```
Red [
	Title:	 "Red GUI Console"
	Author:	 "Qingtian Xie"
	File:	 %gui-console.red
	Icon:	 %app.ico
	Needs:	 [View CLI]
	Config:	 [gui-console?: yes red-help?: yes]
	Rights:  "Copyright (C) 2014-2018 Red Foundation. All rights reserved."
	License: {
		Distributed under the Boost Software License, Version 1.0.
		See https://github.com/red/red/blob/master/BSL-License.txt
	}
]

#include %gui-console-min.red
```

qtxie: The modules included in the console are varied when working on different features. I usually use the cli console without any module when doing red/core development. When I work on a task need some modules, I copy the `console.red` and add the modules, it takes less than 30s and I only need to do that once per task. I think it's not a pain for contributors.

hiiamboris: > It takes less than 30s and I only need to do that once per task. I think it's not a pain for contributors.

If memory serves, I had to explain the process of building consol-view to more than five people, and I think some of them gave up the moment they heard they need to mess with the source files.

qtxie:
> some of them gave up the moment they heard they need to mess with the source files.

I don't think those _contributors_ can do any serious contribution for this project if they gave up for such a simple modification.

> but when you work across a few projects at a time, it's a mess.

Just keep your versions of console.red in the CLI folder, don't commit them, git won't complain when switching branches. For example, I have `console-view.red` and `console-audio.red` in the CLI folder, I don't have any issue when switch branches.
image

hiiamboris:
> I don't think those _contributors_ can do any serious contribution for this project if they gave up for such a simple modification.

No these were just regular users trying to work around some bugs, like introduced a few months ago `forall` crash that ruined a lot of my code, or trying to add compiled stuff to the console so they don't have to compile their program every time. I don't remember all the cases already.

> > but when you work across a few projects at a time, it's a mess.
> 
> Just keep your versions of console.red in the CLI folder, don't commit them, git won't complain when switching branches. For example, I have `console-view.red` and `console-audio.red` in the CLI folder, I don't have any issue when switch branches. 

That's what I'm also doing yes. I'll have to start using multiple build scripts then. `build-cli` and `build-gui` I'm using, now I will have to add `build-cli-no-cli` and `build-gui-no-cli` for Audio branch and older builds that do not have CLI. Anyway, everything can be solved easily, I'm only saying you guys are making it messier than it has to be and neglect that you're doing it. That's all :)

hiiamboris:
Anyway, we've spent more time arguing that anyone of us would spend working around whatever is in this PR, so @dockimbel just state your final call and I'll commit it.

qtxie:
> I'm only saying you guys are making it messier than it has to be and neglect that you're doing it. That's all :)

It's not a mess at all. It's just not the way you like. :-) We'll have more and more modules, you cannot just include all of them in the console by default when you need to compile the console dozens of times each day.

greggirwin:
I wasn't going to say anymore, as it's not an urgent issue IMO. We can think more over time. However, in the big picture...

> We'll have more and more modules, you cannot just include all of them in the console by default when you need to compile the console dozens of times each day.

1) Having a %user.red file that is in their hands solves a lot of this (I do this with a ` --catch user.red` shortcut today). The fact that Red's console is open, where R2's wasn't, means people can and want to build more into their custom consoles. New people sometimes dive too deep, too fast, and think it will be magically easy. 

2) Modular compilation, or modular runtime construction, is the future. The latter also means we need to think deeply about how to keep it as safe as possible, which is the central tension with easy extensibility. What code do you trust?

hiiamboris:
Biggest questions in this PR are:
1. Whether it should be `cli/process-into function-name` or `cli/process-into 'function-name`
   I tend to prefer the 1st option, but compiler doesn't, and since not every script is compiled with `-e`, often we have to write `cli/process-into ('function-name)` which I clearly do not prefer.
2. Never liked `process-into` as a name myself, but a better one never occurred to me.

greggirwin:
Consistency in how we call things makes option 2 seem like the better choice. It's a tradeoff, but there's no perfect solution, simply because of how things have to work technically.

On  `process-into`: `process-using` or `process-with` ?

Another name consideration: `post-scriptum` could be `postscript`.

Since I'm glancing at it, what do you see as the use cases for `/no-version /no-help`? If the user defines their own, the override the built-in versions, correct?

hiiamboris:
> Since I'm glancing at it, what do you see as the use cases for `/no-version /no-help`? If the user defines their own, the override the built-in versions, correct?

Mainly if they want `-h` for their own use, or disable `--version`, `-h`, or `--help` (less likely).

greggirwin:
It doesn't look like you can override this way:
```
main: func [
	/help
	/h "alias /help"
	/version
	/v "alias /version"
][
	if help [print "HELP!"]
	if version [print [now/precise]]
	print ["Thanks for playing!"]
	halt
]

cli/process-into 'main
```

But my old compiled tests just fail right now, and compiling without the lit-word for the func also fails. I'm synced up, but must have something else going on. If the overrides, like above, just worked, We could do away with `/no-version /no-help`, but it's not urgent. We can see what people think once it's merged.

greggirwin:
@hiiamboris @dockimbel are there any other changes you want made before merging this?

hiiamboris:
I'm waiting for Nenad to say final word how we handle the module inclusion.

pekr:
> Biggest questions in this PR are:
> 
>     1. Whether it should be `cli/process-into function-name` or `cli/process-into 'function-name`
>        I tend to prefer the 1st option, but compiler doesn't, and since not every script is compiled with `-e`, often we have to write `cli/process-into ('function-name)` which I clearly do not prefer.
> 
>     2. Never liked `process-into` as a name myself, but a better one never occurred to me.

What about `/output`? It is used by `call`. I would not mind even using just `/into`, if there is not a clash with something else.

greggirwin:
@hiiamboris something else for us to think about WRT lit-args, is how they affect other code by requiring other funcs to use them for internal propagation. I'm guessing that's why other funcs in CLI use them as well, not just the entry point.

hiiamboris:
Those CLI funcs that are using lit-arg are user-facing.

hiiamboris:
I tried to remove CLI module from both consoles and create two separate files `console-full.red` and `gui-console-full.red`, but this didn't go well. View module can be turned off with some options (`--no-view`, `view?: off`, and some more, it's a mess I can't wrap my head around). In this case do we need a third file that has CLI but no View? Or do we include and exclude CLI together with View? Both options look dumb. I'm open to leaving both consoles as they are, but then we need to process not only `console.red` source during `red.r` deployment but also `gui-console.red`.

Even bigger issue is libRedRT. It makes most sense to include everything into it so we won't have to recompile it. But I cannot include CLI into it, because it simply will error during compilation. Looks like compiler tries to compile the Needs value? I have no idea. It tries to pass `exec/f_ <anon431>` whatever it is to `undecorate` function in `libRedRT.r` is all I could find out. I don't know how to debug it.

hiiamboris:
I suggest we merge it as is, then later someone (maybe me) will raise an issue of libRedRT not having CLI included.

greggirwin:
Let's see what @dockimbel. I'm all for putting features out in "good enough" shape to get feedback on them. 

Nenad asked why it was needed for consoles, and @hiiamboris noted that it can work for interpreted scripts. I will note that the idea of building custom consoles for this is too much work for users (not talking about contributors, but average community members). We either include it by default, or we say "If you run interpreted scripts and want to use CLI use `do %../cli.red` in your script.

This leads to Rebol's %user.r feature: http://www.rebol.com/docs/core23/setup.html#section-4  I hack this in Red using `--catch` in my console shortcuts. There are a number of issues with this approach, but the general idea of making it easy for users to configure an "environment" is a good one. It also leads down the long dark path of `include`, modules, a build system, and more. 

For now, if the core team wants to leave out things like CLI and XML, what's easiest for users that doesn't require deep changes to Red?

qtxie:
> Nenad asked why it was needed for consoles

I think he was asking why modify `console.red` and `gui-console.red` directly.

For the end users (those who use red.exe), we can inject the CLI and XML modules in `red.r`, in this way they have all the modules by default.

hiiamboris:
@qtxie it currently inserts only `View` module and only into CLI console: 
https://github.com/red/red/blob/1a04975b0858cfe97647005aceed17c2de1e3620/red.r#L465-L485
`view?` is controlled by `--no-view`:
https://github.com/red/red/blob/1a04975b0858cfe97647005aceed17c2de1e3620/red.r#L748

What do you propose:
- `--no-view` will exclude both View and all other modules?
- we rename `--no-view` to `--no-modules` so it will add/remove all supported modules?
- `red.r` always injects all other modules, but `--no-view` still controls View module?

I never used this flag, so it's not up to me to decide.

qtxie:
> red.r always injects all other modules, but --no-view still controls View module?

This option is OK to me.

hiiamboris:
Done

greggirwin:
I want to get CLI out. AFAIK it just has to be set up for the new, interim, module approach. And for @dockimbel to say whether he wants it in the main repo or in red/code.

hiiamboris:
> has to be set up for the new, interim, module approach

Anything more concrete? I know nothing of this approach.

> main repo or in red/code

I had hoped we have resolved the batteries excluded argument already.

PR #5019, opened 2021-12-31
FEAT: high level tracer

hiiamboris:
cleaner ver with operators

### DIFF
--- a/environment/tools.red
+++ b/environment/tools.red
@@ -488,6 +488,7 @@ system/tools: context [
 					][
 						clear change skip tail r -4 "..."
 					]
+					clear skip r part					;-- when part < 3-4
 				]
 				r
 			]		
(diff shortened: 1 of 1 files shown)


hiiamboris:
For whatever reason, compiler hates the `<<<` word.
It compiles the console and console works, but attempt to compile libRedRT with that word inside fails with:
```
Compiling to native code...
*** Driver Internal Error: Script Error : undecorate expected sym argument of type: word path

*** Where: repend
*** Near:  [undecorate def]
```
As a result, most of the test suite crashes, then just hangs on the last script:
```
** - unicode******************************************0 / 1 **
** - preprocessor tests*******************************0 / 42 **
** - run-all-comp1*red********************************0 / 1 **
** - run-all-comp2*red********************************0 / 1 **
** - Red print****************************************1 / 18 **
running  run-time-error-test.r
```
Should I open an issue about it?

hiiamboris:
I take it as a yes then.

PR #5022, opened 2022-01-03
FEAT: high level tracer

hiiamboris:
3rd variant, proposed by Gregg

### DIFF
--- a/environment/tools.red
+++ b/environment/tools.red
@@ -534,11 +534,20 @@ system/tools: context [
 				level:   level % 10 + 1 * 2				;-- cap at 20 as we don't want indent to occupy whole column
 				
 				either data/isubex? [
-					expr: p: last data/stkexs
+					expr: last data/stkexs
+					isop?: all [
+						3 = length? expr
+						word? :expr/1
+						op? get/any :expr/1
+					]
 					either event = 'push [
 						expr: back tail expr
 					][
 						while [set-word? :expr/-1] [expr: back expr]
+					]
+					if isop? [							;-- locally reorder op calls for more readability
+						expr: copy expr
+						move p: skip tail expr -3 next p
 					]
 				][
 					p: tail data/topexs
--- a/environment/tools.red
+++ b/environment/tools.red
@@ -275,32 +275,35 @@ system/tools: context [
 		;; context for trace data collected by 'collector' tracer and it's options
 		data: context [
 			;; input of collector:
-			debug?:   no								;-- /debug refinement (raw events output)
-			inspect:  none								;-- inspect function to call
-			ievents:  none								;-- events accepted by this inspect function (none = unfiltered)
-			iscopes:  none								;-- list of scopes accepted by this inspect fn (none = unfiltered)
-			isubex?:  none								;-- whether to call inspect on subexpressions
+			debug?:        no							;-- /debug refinement (raw events output)
+			inspect:       none							;-- inspect function to call
+			event-filter:  none							;-- events accepted by this inspect function (none = unfiltered)
+			scope-filter:  none							;-- list of scopes accepted by this inspect fn (none = unfiltered)
+			inspect-sub-exprs?: none					;-- whether to call inspect on subexpressions
 			;; entered blocks/parens/paths stack:
-			blocks:   []								;-- copied
-			orgblk:   []								;-- original
+			copied-blocks: []							;-- copied
+			orig-blocks:   []							;-- original
 			;; depth tracking:
-			fdepth:   0									;-- function call depth (prologs/epilogs only)
-			level:    []								;-- nesting level of expressions in each block (last 0 = top lvl)
-			path:     []								;-- path of refs up to current scope (starts empty)
+			func-depth:    0							;-- function call depth (prologs/epilogs only)
+			expr-levels:   []							;-- nesting level of expressions in each block (last 0 = top lvl)
+			path:          []							;-- path of refs up to current scope (starts empty)
 			;; mirrors of Red stack (of values):
-			stack:    []								;-- closer to code (words left alone, series copied)
-			evstack:  []								;-- after evaluation, raw values
+			stack:         []							;-- closer to code (words left alone, series copied)
+			eval'd-stack:  []							;-- after evaluation, raw values
 			;; entered expression lists:
-			topexs:   []								;-- top-only, inside code copy
-			subexs:   []								;-- all exprs, inside code copy
-			orgexs:   []								;-- all exprs, inside original code
-			stkexs:   []								;-- all exprs, inside the stack (partially evaluated)
+			copied-top-exprs: []						;-- top-only, inside code copy
+			copied-exprs:  []							;-- all exprs, inside code copy
+			orig-exprs:    []							;-- all exprs, inside original code
+			stack-exprs:   []							;-- all exprs, inside the stack (partially evaluated)
 	
 			reset: function ["Reset collector's data"] [
-				; set [debug? inspect iscopes ievents isubex?] none
-				blks: [blocks orgblk level path stack evstack topexs subexs orgexs stkexs]
+				; set [debug? inspect scope-filter event-filter inspect-sub-exprs?] none
+				blks: [
+					copied-blocks orig-blocks expr-levels path stack eval'd-stack
+					copied-top-exprs copied-exprs orig-exprs stack-exprs
+				]
 				foreach b blks [clear get b]
-				self/fdepth: 0
+				self/func-depth: 0
 			]
 
 			collector: function [
@@ -320,7 +323,7 @@ system/tools: context [
 							not tail? p: skip copy code offset
 						][
 							p: head change p as tag! uppercase form p/1
-							if s: pick tail topexs -2 [p: at p index? s]
+							if s: pick tail copied-top-exprs -2 [p: at p index? s]
 							p
 						]
 						code
@@ -330,15 +333,19 @@ system/tools: context [
 						pad :ref 10
 						pad mold/flat/part :value 20 22
 						pad mold/flat/part code2 60 62
-						pad level 8
+						pad expr-levels 8
 					]
 				]
 				
 				call: [
 					all [								;-- filtering by events, scope, expression level:
-						any [none? ievents  find ievents event]
-						any [none? iscopes  none? code  find/same/only iscopes code]
-						any [isubex?  0 = last level  all [1 = last level  find [open call return] event]]
+						any [none? event-filter  find event-filter event]
+						any [none? scope-filter  none? code  find/same/only scope-filter code]
+						any [
+							inspect-sub-exprs?
+							0 = last expr-levels
+							all [1 = last expr-levels  find [open call return] event]
+						]
 						inspect system/tools/tracers/data event code offset :value :ref frame
 					]
 				]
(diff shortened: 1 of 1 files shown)


hiiamboris:
For some reason, `log` in `trace-9` test became twice as big in this PR and the test is failing.
For some other reason, failure is only detected in run-all-comp, but not in run-all-interp, even though the test clearly fails when run from the console.

environment/tools.red
```diff
@@ -244,49 +244,341 @@ system/tools: context [
...
+			]
 		]
-	]
+		
+		;; context for trace data collected by 'collector' tracer and it's options
+		data: context [
+			;; input of collector:
+			debug?:   no								;-- /debug refinement (raw events output)
+			inspect:  none								;-- inspect function to call
+			ievents:  none								;-- events accepted by this inspect function (none = unfiltered)
```
greggirwin, line 280: Please consider more meaningful names for the current smashed words, as I suggested in my example on Telegram.

hiiamboris: `i` stands for `inspect-function-related`
it's meaningful :)

greggirwin: You have comments, describing the names, which is great. Because otherwise they'd be opaque. :^) This applies to all the smashed words in the context. Saying "it's meaningful" doesn't make it so, and is not a valid defense. If I felt they were clear, I wouldn't have suggested other names on Telegram, and then commented here. In the larger context of teamwork, when you get feedback suggesting changes, please consider if your _response_ to that feedback may be read as "It's fine the way it is.", because that is how I often interpret your responses to my suggestions.

Also, please use your best judgement when implementing suggestions, which are often worded quickly as guidelines. For example, if you change the name to `inspect-function-related-events` I will see through your ploy to make the alternative look worse, thereby defending your position. :^)

Let's see what @dockimbel says.

hiiamboris: No need to be mean :)
I do find them fine as they are, but only because when I expand them they uglify the code with their verbosity. I agree and also want to expand them but it's all tradeoffs.

hiiamboris: Since you're better than me at naming, propose names for the rest of the data, not only `orig-exprs` and `copied-top-exprs`. For example, shorter `inspect-events` looks like an action, because it focuses on `inspect` rather than on `events`. IIRC that's why I called it `ievents` in the first place.

hiiamboris: Draft (please improve):
```
ievents:  ???								;-- events accepted by this inspect function (none = unfiltered)
iscopes:  ???								;-- list of scopes accepted by this inspect fn (none = unfiltered)
isubex?:  call-on-subexprs?					;-- whether to call inspect on subexpressions
orgblk:   orig-blocks						;-- original
fdepth:   func-depth						;-- function call depth (prologs/epilogs only)
level:    expr-levels						;-- nesting level of expressions in each block (last 0 = top lvl)
evstack:  evaluated-stack ??				;-- after evaluation, raw values
topexs:   copied-top-exprs					;-- top-only, inside code copy
subexs:   copied-exprs						;-- all exprs, inside code copy
orgexs:   orig-exprs						;-- all exprs, inside original code
stkexs:   stack-exprs						;-- all exprs, inside the stack (partially evaluated)
```

hiiamboris: Draft (please improve):
```
ievents:  ???								;-- events accepted by this inspect function (none = unfiltered)
iscopes:  ???								;-- list of scopes accepted by this inspect fn (none = unfiltered)
isubex?:  call-on-subexprs?					;-- whether to call inspect on subexpressions
orgblk:   orig-blocks						;-- original
fdepth:   func-depth						;-- function call depth (prologs/epilogs only)
level:    expr-levels						;-- nesting level of expressions in each block (last 0 = top lvl)
evstack:  evaluated-stack ??				;-- after evaluation, raw values
topexs:   copied-top-exprs					;-- top-only, inside code copy
subexs:   copied-exprs						;-- all exprs, inside code copy
orgexs:   orig-exprs						;-- all exprs, inside original code
stkexs:   stack-exprs						;-- all exprs, inside the stack (partially evaluated)
```

greggirwin: Will review today and see what names come to me. :+1:

One right now is `evaluated-stack` could be `eval'd-stack`. Having `'` in names is a little on the edge, but good for human abbreviations sometimes. OK, one more (going off just names, and not total code context here). `call-on-subexprs?` could be `inspect-sub-exprs?`

greggirwin: For `ievents`, what about `event-filter`? That makes the `none?` check for it clearer as meaning "no event filter".

greggirwin: Same for `iscopes` becoming `scope-filter`.

hiiamboris: it's ok I guess, the only issue is that those filters do not make it clear that they belong to `inspect` func

greggirwin: Is there something else they could be related to?

hiiamboris: Well, there's an event filter in every tracer. Initially I had one filter in `collector` and one in `inspect` (both are tracers), but then removed the former to allow every event to be acted upon. That partially lifts off the ambiguity.

greggirwin: I leave the final call to you. :^)

hiiamboris: Committed

environment/tools.red
```diff
@@ -244,49 +244,348 @@ system/tools: context [
...
+		]
+		
+		;; helpers to keep code readable, unlike `change/only back back tail series last series`
+		push: func [s [series!] i [any-type!] /dup n [integer!]] [append/only/dup s i any [n 1]]
+		drop: func [s [series!] n [integer!]] [clear skip tail s negate n]
+		pop:  func [s [series!]] [take/last s]
+		top:  func [s [series!]] [back tail s]
 
-			if event = 'open [indent: indent + 1]
+		incr: function [x [word! series!] /by o] [		;@@ remove this once we have it generally available
```
dockimbel, line 266: That function is really not necessary here (or in general). It obsfuscates the code more than it helps. 

`incr/by 'offset -1` vs `offset: offset - 1`.

This later is clear and simple, the former is ugly and not clear. Don't count on it to be generally available, at least not in this form and name.

environment/tools.red
```diff
@@ -244,49 +244,348 @@ system/tools: context [
...
+			expr-levels:   []							;-- nesting level of expressions in each block (last 0 = top lvl)
+			path:          []							;-- path of refs up to current scope (starts empty)
+			;; mirrors of Red stack (of values):
+			stack:         []							;-- closer to code (words left alone, series copied)
+			eval'd-stack:  []							;-- after evaluation, raw values
+			;; entered expression lists:
+			copied-top-exprs: []						;-- top-only, inside code copy
+			copied-exprs:  []							;-- all exprs, inside code copy
+			orig-exprs:    []							;-- all exprs, inside original code
+			stack-exprs:   []							;-- all exprs, inside the stack (partially evaluated)
```
dockimbel, line 297: You should use `make block! <size>` to initialize literal blocks. Not only it would avoid bloating the Redbin payload with many empty blocks, but also would help reduce the memory pressure at runtime by preallocating enough slots for the common use-cases.

environment/tools.red
```diff
@@ -244,49 +244,348 @@ system/tools: context [
...
+			];; collector function
+		];; data context
 	
+		guided-trace: function [
+			"Trace a block of code, providing 'inspect' tracer with collected data"
+			inspect [function!] "func [data [object!] event code offset value ref frame]"
+			code    [any-type!]
+			all?    [logic!]    "Trace all sub-expressions of each expression"
+			deep?   [logic!]    "Enter functions and natives"
+			debug?  [logic!]    "Dump all events encountered"
```
dockimbel, line 458: We don't put docstrings in functions that are not meant for end-users to use directly. This bloats the Redbin payload unnecessarily. Just use comments instead.

dockimbel:
I have made some comments about some code parts, but you don't have to make those changes, I will merge this PR and proceed with the changes myself.

environment/tools.red
hiiamboris: Well, if I felt it obfuscates the code I wouldn't use it in the first place:
```
incr top expr-levels
```
vs
```
change top expr-levels 1 + last expr-levels
```

FYI the other main use for it (not used in this PR) is:
```
incr 'face1/face2/extra/obj1/obj2/x
```
vs
```
face1/face2/extra/obj1/obj2/x: face1/face2/extra/obj1/obj2/x + 1
```

Though I recall my first reaction to it was exactly like yours :)

environment/tools.red
hiiamboris: I actually intended this to be used by deeper hackers ;)

hiiamboris: Good to know about the general commenting guidelines though :+1:

hiiamboris:
OK! Leave it up to you then :)

dockimbel:
I gave it a try on `%tests/demo.red`:

```
>> trace %tests/demo.red
    data: [[0 7 20] [2 2 6 7 20] [1 ...] => [[0 7 20] [2 2 6 7 20] [1 2 6 7 20] [...]  
    pattern: "Red"                       => "Red"                                      
    prin newline                         => unset                                      
  ````FOREACH line data [cursor: 1 + line/1 line: next line gap-start: line/1 gap-e...]
      cursor: 1 + line/1                 => 1                                          
      line: next line                    => [7 20]                                     
      gap-start: line/1                  => 7                                          
      gap-end: line/2                    => 20                                         
	      prin tab                           => unset                                      
	      prin tab                           => unset                                      
  ````````FOREACH/REPEAT/EITHER prin either all [gap-start <= i i <= gap-end] [#" "...]
          pattern/:cursor                => #"R"                                       
R  ``````FOREACH/REPEAT i 21 [prin either all [gap-start <= i i <= gap-end] [#" "] [...]
        prin either all [gap-start <...] => unset                                      
        if i > gap-end [unless tail?...] => none                                       
        cursor: cursor + 1               => 2                                          
        if cursor = 4 [cursor: 1]        => none
```

What is `FOREACH/REPEAT/EITHER` and where does it come from? No `repeat` or `either` in the tracing above...

hiiamboris:
> No `/here` refinement. I'll fix that too.

Yeah. Leftover.

hiiamboris:
> I gave it a try on `%tests/demo.red`:
> 
> ```
> >> trace %tests/demo.red
>     data: [[0 7 20] [2 2 6 7 20] [1 ...] => [[0 7 20] [2 2 6 7 20] [1 2 6 7 20] [...]  
>     pattern: "Red"                       => "Red"                                      
>     prin newline                         => unset                                      
>   ````FOREACH line data [cursor: 1 + line/1 line: next line gap-start: line/1 gap-e...]
>       cursor: 1 + line/1                 => 1                                          
>       line: next line                    => [7 20]                                     
>       gap-start: line/1                  => 7                                          
>       gap-end: line/2                    => 20                                         
> 	      prin tab                           => unset                                      
> 	      prin tab                           => unset                                      
>   ````````FOREACH/REPEAT/EITHER prin either all [gap-start <= i i <= gap-end] [#" "...]
>           pattern/:cursor                => #"R"                                       
> R  ``````FOREACH/REPEAT i 21 [prin either all [gap-start <= i i <= gap-end] [#" "] [...]
>         prin either all [gap-start <...] => unset                                      
>         if i > gap-end [unless tail?...] => none                                       
>         cursor: cursor + 1               => 2                                          
>         if cursor = 4 [cursor: 1]        => none
> ```
> 
> What is `FOREACH/REPEAT/EITHER` and where does it come from? No `repeat` or `either` in the tracing before that, so it looks confusing to me...

I guess for some reason `all`-expressions are not detected as top-level, but `either`-expressions are...
```
>> trace [either true [0][1]]
  ````EITHER true [0] [1]```````````````````````````````````````````````````````````````````````````````````````````````````````````````````````````````````````
      0                                                      => 0                                                                                               
  ``````````````````````````````````````````````````````````````````````````````````````````````````````````````````````````````````````````````````````````````
    either true [0] [1]                                      => 0                                                                                               
== 0
>> trace [all [0 1]]
    all [0 1]                                                => 1                                                                                               
== 1
```

hiiamboris:
Let's add `any` and `all` to our `[reduce do/next]` list of natives that need to report their expressions.

environment/tools.red
greggirwin: If we have `incr` we should have `decr`.

greggirwin: And I do think `incr` adds meaning, as a name.

environment/tools.red
greggirwin: Those allocations only hit us once though. I agree that preallocating makes sense sometimes, but `[]` is more declarative and visible to the reader as a block.

environment/tools.red
greggirwin: Ummmm, aren't we all end users? :^) If this is useful information, and we don't know what tools might exist in the future, I don't want to be the guy that said "Use 2 digits for years." Func specs support doc strings, so I'd like to encourage people to use them. Of course, we can ballpark some numbers about how much bloat this might lead to, but then have to guess quite a bit about how much time is saved, or bugs reduced, if that information lives only in comments.

PR #5025, opened 2022-01-05
Add no-trace attr to func spec parser.

greggirwin:


hiiamboris:
and `[trace]`?

greggirwin:
Ah, yes. Not used in any yet, but will trip `help` as soon as it is. Good catch.

toomasv:
https://github.com/red/docs/blob/master/en/datatypes/function.adoc should also be synchronized?

greggirwin:
Yes, docs should be updated as well.

dockimbel:
@greggirwin You should rebase your PR, as there are unrelated changes in `%environment/tools.red` in it.

greggirwin:
Grrr. Git.

PR #5026, opened 2022-01-06
FEAT: Added XML codec

rebolek:


hiiamboris:
This bug still persists: https://gitter.im/red/codecs?at=61b346418853d316403de86f

rebolek:
Gee, I thought it was fixed already, thanks, I’ll fix it again.

rebolek:
Found the problem, it happened only with empty tags.

hiiamboris:
This PR also needs tests.

rebolek:
> This PR also needs tests.

Of course, they will be added.

greggirwin:
@dockimbel we're still finalizing this, so don't merge just yet.

greggirwin:
Talking now about how best to handle options.

dockimbel:
Is there some design document or just a documentation on how the Red datatypes are mapped to XML elements?

greggirwin:
https://gitlab.com/rebolek/markup/-/blob/main/README.md docs the output formats. There is no schema type system beyond that (though I built one for Prolific in the past). All tags map to words, unless you say to use strings as keys, and all other data is just kept as strings from being parsed.

environment/codecs/XML.red
```diff
@@ -0,0 +1,940 @@
...
+	verinfo: encinfo: namespace: nl?:
+		none
+	cont-val: ""
+	value?: false
+	attributes: none
+
+	target: []
+	target-stack: []
+
+	set 'load-xml func [
```
dockimbel, line 86: There is a general problem with how the codecs are implemented. They don't respect the codec API (load/save), but provide adhoc global functions that are circumventing the `load/as` and `save/as` ones instead of just been simple handy wrappers on top of them. Though, as this is a global implementation issue in most codecs, I would not veto this PR based on that. I'm just mentioning it here, so that people gets aware of that.

environment/codecs/XML.red
```diff
@@ -0,0 +1,940 @@
...
+		element
+	]
+
+	; -- Character Range
+	Char: {[#x1-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]}
+	RestrictedChar: {[#x1-#x8] | [#xB-#xC] | [#xE-#x1F] | [#x7F-#x84] | [#x86-#x9F]}
+
+	Char: charset [
+		#"^(01)" -#"^(D7FF)" #"^(E000)" - #"^(FFFD)" #"^(10000)" - #"^(10FFFF)"
+	]
```
dockimbel, line 158: This charset eats 136 KB of memory.

environment/codecs/XML.red
```diff
@@ -0,0 +1,940 @@
...
+
+	; -- Names and Tokens
+	NameStartChar:
+	name-start-char: charset [
+		":_" #"a" - #"z" #"A" - #"Z" #"0" #"^(C0)" - #"^(D6)" #"^(D8)" - #"^(F6)" 
+		#"^(F8)" - #"^(02FF)" #"^(0370)" - #"^(037D)" #"^(037F)" - #"^(1FFF)"
+		#"^(200C)" - #"^(200D)" #"^(2070)" - #"^(218F)" #"^(2C00)" - #"^(2FEF)"
+		#"^(3001)" - #"^(D7FF)" #"^(F900)" - #"^(FDCF)" #"^(FDF0)" - #"^(FFFD)"
+		#"^(010000)" - #"^(0EFFFF)"
+	]
```
dockimbel, line 180: This charset eats 120 KB of memory.

environment/codecs/XML.red
```diff
@@ -0,0 +1,940 @@
...
+	NameChar:
+	name-char: union name-start-char charset [
+		"-." #"0" - #"9" #"^(B7)" #"^(0300)" - #"^(036F)" #"^(203F)" - #"^(2040)"
+	]
```
dockimbel, line 184: This charset eats 120 KB of memory.

dockimbel:
My point about the charsets size is that such memory will be allocated on loading the XML codec, whether or not he user code calls the codec or not. We should be careful to not make such allocations unless needed. I suggest moving all the significant allocations to a `init` function that is called the first time the codec is invoked.

Such `init` callback function should be part of the codec infrastructure and called automatically when needed without having each codec to implement the logic for that.

dockimbel:
As another global comment, I have always been in favor of a streaming approach for XML (and other formats like JSON and CSV) rather than static mappings to a predefined datastructure. XML for Red is just a serialization format to communicate with third-party software, it is not a native format for us. We should then treat it as that and not try to insert a potentially heavy datastructure between the raw XML data and the user's datastructures. The streaming approach (beyond all performance and resource usage benefits) allows the user to map directly the XML raw data to the final datastructure and is used in the industry since two decades. I haven't looked the the generative counterpart to streaming parser, but we can certainly come up with a good design for that part too.

Again, I won't veto this PR based on that comment. I plan to implement in the future (in my "20% fun time with Red") at least one streaming parser in the future to show how it can be done in Red efficiently.

environment/codecs/XML.red
greggirwin: For codecs that need options, the `load/save` interface needs to change. `/as` doesn't give you enough control. That means adding something completely general like `/with`, which has been a deep discussion for this codec, and how it compares to using refinements. Then `encode/decode` in each codec need to support a general `/with options` refinement, whether they need them or not. 

I'm open to all suggestions, since we may not add many more standard codecs, but users will write them and we want those to use a consistent approach.

environment/codecs/XML.red
```diff
@@ -0,0 +1,940 @@
...
+		element
+	]
+	loose-document: [
+		opt prolog
+		element
+	]
+
+	; -- Character Range
+	Char: {[#x1-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]}
+	RestrictedChar: {[#x1-#x8] | [#xB-#xC] | [#xE-#x1F] | [#x7F-#x84] | [#x86-#x9F]}
```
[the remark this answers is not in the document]
greggirwin: I'll hazard a guess that @rebolek meant to comment those out.

greggirwin:
> My point about the charsets size is that such memory will be allocated on loading the XML codec, whether or not he user code calls the codec or not. We should be careful to not waste memory space unless needed. I suggest moving all the significant allocations to a `init` function that is called the first time the codec API is invoked.
> 
> Such `init` callback function should be part of the codec infrastructure and called automatically when needed without having each codec to implement the logic for that.

I agree with not wasting space and time, but are we better off making `init` a best practice (even beyond codecs), or should codecs be listed in `needs`? Codecs that just call OS APIs (e.g. images) come basically for free, and `redbin` we want there all the time. The rest can be optimized out if someone cares. Nice when batteries are included, so it's almost an `exclude` feature, kind of like R2's various kernel choices.

greggirwin:
I'll be very anxious to see how streaming works, and I think @rebolek has done a SAX parser in the past. I will say that I think the static model is easy to reason about, and we should see what _existing_ systems use streaming JSON or XML that we can test with. Formats _designed_ for streaming are a different story. Formats _not_ designed for streaming pose issues like "I did a whole bunch of work and then hit a malformed bit." Now I have to track state, make sure if I've taken any actions that need to be undone they _can_ be undone with information I have, etc.

OTOH, eventing systems and aggregation are going to be key components, so thinking about it is important.

greggirwin:
Messaging in general is key, and how you envelope things to pass them around, etc. It's also a wide world of needs and approaches. Everything from MQTT, fire and forget, to 0MQ brokering systems.

environment/codecs/XML.red
greggirwin: Today, the `load-*/save-*` funcs are for when you need more control. We _could_ eliminate some options from the codecs, so they work just as the defaults wrapped by `encode/decode` (so the calls go `load -> decode -> load-*` today). 

The tradeoff in removing features is more work for users. e.g. we can always output `/pretty` for JSON, and let users `trim/lines` if they want. I believe we added `/ascii` after research on compatibility, but we can also say "forget people stuck with that" and always use UTF-8. For XML, we can always emit one format and post-process to get the others if people want them. That eliminates `/as` . We can also always include metadata and let users remove or ignore it if it's there. That eliminates `/meta`. I think `/str-keys` was there because XML names can be invalid words. We can tell people who hit that problem that those aren't supported, or we can always load them as strings and post-process to words or deal with them as strings; neither of which is great. We can leave out `/str-keys` for now, and see who complains. Adding it later won't be a breaking change. For those issues, we can normalize the interface and push the problem around. Is that a win overall?

Then there's CSV. XML is a pain, but a hard spec. CSV is a very soft spec, with lots of variants in the wild. Wrapping those details in the codec save the user enormous pain and possible errors. If we limit the options, it becomes much less useful. So it's a great, concrete example to use when we ask "How should codecs with options be handled?"

environment/codecs/XML.red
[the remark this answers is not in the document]
rebolek: > My point about the charsets size is that such memory will be allocated on loading the XML codec, whether or not he user code calls the codec or not. We should be careful to not waste memory space unless needed. I suggest moving all the significant allocations to a `init` function that is called the first time the codec API is invoked.
> 
> Such `init` callback function should be part of the codec infrastructure and called automatically when needed without having each codec to implement the logic for that.

I’ll agree on the `init` function, I’ll update the codec.

environment/codecs/XML.red
rebolek: The `load-*` first appeared in @giesse's JSON codec and I followed that path with CSV and XML. We can get rid of it but they allow for setting options as @greggirwin mentioned. If we can think of a different solution that would work for all codecs generaly (e.g. setting JPEG quality etc.) then I’m for getting rid of `load-*/save-*`.

greggirwin: Cross posting from Gitter:

We are a week away from month end, and my goal is to get XML out, along with Animation. We'll note in the blog post that there may be changes, and that we are reviewing the overall codec interface.

The original JSON interface is a legacy issue for us now. R2 didn't have codecs, and that code has been around in various forms for a very long time. Romano did the first version, I took over maintenance and kept it up with Douglas Crockford on JSON.org, then @giesse, @rebolek, and @hiiamboris contributed to improving it; but the original interface lives on. :^\

I don't want to pull anyone's focus right now, but also don't want to just hack in a different interface without thinking it through. So it will be a pretty big task. It will likely also inform other extension models or how a system catalog works, even for I/O protocols perhaps.

greggirwin:
> My point about the charsets size is that such memory will be allocated on loading the XML codec, whether or not he user code calls the codec or not. We should be careful to not waste memory space unless needed. I suggest moving all the significant allocations to a `init` function that is called the first time the codec API is invoked.
> 
> Such `init` callback function should be part of the codec infrastructure and called automatically when needed without having each codec to implement the logic for that.

@rebolek put the `init` change in, and I'm...I'll just say that while I agree about not wasting time and memory, this change made the code _SO_ much harder to understand (context level vars now in funcs), and _SO_ much more fragile (have to keep two sets of words in sync in different parts of the code), that I want to be very careful about what we prioritize. In order to save space, the ~100 rule names in `load-xml` are tightly packed into 15 lines.

This is not a criticism of Bolek's code, but of how we address issues. We don't have modules today, but we have worse code to live with once we could easily leave XML out. Given this change (and I hate to see the time and effort on `init` wasted), I'd almost say go back to the original and leave XML as something for users to add.

rebolek:
@greggirwin I can unpack the rule names but I agree that it’s much more fragile. Too bad we can’t extend objects as that would solve this problem and we didn’t have to have the words defined twice.

hiiamboris:
You're overengineering and dramatizing IMO. Nenad's point was to only stash charsets into `init`, and maybe to optimize them using `not`.

greggirwin:
Right now I want to get the XML codec out, and make the code the best it can be. I am all for optimizing with `not`, as long as we don't end up obfuscating the rules too much, when compared with the spec. But not right now. :^) Options:

1) Go back to the original version, pre `init`, and pay the price. That means making it optional later may break code, but will be easy for people to fix by including the XML codec. 

2) Go back to the original version, pre `init`, and make people include the XML codec manually. Not as magical for users, where it just works, and we need to decide where %xml.red lives. Probably red/code. 

3) See if there's an easy, general solution for including codecs optionally, e.g. via `needs: [codecs [xml]]`. This is, again, less magical. But we may only add a couple more standard codecs, and all others will be optional, so we want to make it easy. This, unfortunately, leads to a lot of other questions we're not ready to answer. We can stub our way around this for XML by including the source code in Red, adding a little bloat, but not evaluating it unless it's used.

4) Take the time to fix the `init` code approach, documenting it clearly, because people learn from and copy code. 

5) ???

Magical is great, but as long as it's dead easy to include, that's almost as good. It's all tradeoffs. XML, as I've said, straddles the line of what's likely to be useful, and also what we want to encourage people to use. For those who need it (and there is a lot of XML out there), it's an important piece to have available, and good for advertising.

hiiamboris:
(4) will take like what, half an hour tops? ;)

greggirwin:
Go for it, if you want, and we can review.

hiiamboris:
I'll pass, you already have a champion for this quest ;)

greggirwin:
Something like this in %xml.red?
```
	;------------------------------------------------------------------
	; -- Charsets whose creation is deferred until the codec is used, as some are large.
	; !! Names declared here MUST match those used in `init-charsets`.

	Char: RestrictedChar: none
	NameStartChar: name-start-char: none
	NameChar: name-char: none
	
	init-charsets: does [
		; -- Character range
		Char: charset [
			#"^(01)" -#"^(D7FF)" #"^(E000)" - #"^(FFFD)" #"^(10000)" - #"^(10FFFF)"
		]
		RestrictedChar: charset [
			#"^(01)" -#"^(08)" #"^(0B)" - #"^(0C)" #"^(0E)" - #"^(1F)"
			#"^(7F)" - #"^(84)" #"^(86)" - #"^(9F)"
		]

		NameStartChar:
		name-start-char: charset [
			":_" #"a" - #"z" #"A" - #"Z" #"0" #"^(C0)" - #"^(D6)" #"^(D8)" - #"^(F6)"
			#"^(F8)" - #"^(02FF)" #"^(0370)" - #"^(037D)" #"^(037F)" - #"^(1FFF)"
			#"^(200C)" - #"^(200D)" #"^(2070)" - #"^(218F)" #"^(2C00)" - #"^(2FEF)"
			#"^(3001)" - #"^(D7FF)" #"^(F900)" - #"^(FDCF)" #"^(FDF0)" - #"^(FFFD)"
			#"^(010000)" - #"^(0EFFFF)"
		]

		NameChar:
		name-char: union name-start-char charset [
			"-." #"0" - #"9" #"^(B7)" #"^(0300)" - #"^(036F)" #"^(203F)" - #"^(2040)"
		]
		
	]	
	;------------------------------------------------------------------
```
Then in `load-xml`
```
		if none? name-char [init-charsets]
```
All else stays the same.

Thoughts?

greggirwin:
Good to note why names are aliased as well.

greggirwin:
On a related note, I have no love for huge charsets (thank you Unicode), which are human hostile when formed. @toomasv did a nice reformatter for DiaGrammar that recreates an original bitset spec from a bitset.

rebolek:
Ok, I’ll change that.

> Good to note why names are aliased as well.

They’re not anymore, that was cleaned up.

dockimbel:
For initializing charsets, we could use a more Red-ish approach which leverages the code/data equivalence paradigm while keeping the code readable:

```
xml: context [

    Char: [#charset
        #"^(01)" -#"^(D7FF)" #"^(E000)" - #"^(FFFD)" #"^(10000)" - #"^(10FFFF)"
    ]
    RestrictedChar: [#charset
        #"^(01)" -#"^(08)" #"^(0B)" - #"^(0C)" #"^(0E)" - #"^(1F)"
        #"^(7F)" - #"^(84)" #"^(86)" - #"^(9F)"
    ]

    NameStartChar:
    name-start-char: [#charset
        ":_" #"a" - #"z" #"A" - #"Z" #"0" #"^(C0)" - #"^(D6)" #"^(D8)" - #"^(F6)"
        #"^(F8)" - #"^(02FF)" #"^(0370)" - #"^(037D)" #"^(037F)" - #"^(1FFF)"
        #"^(200C)" - #"^(200D)" #"^(2070)" - #"^(218F)" #"^(2C00)" - #"^(2FEF)"
        #"^(3001)" - #"^(D7FF)" #"^(F900)" - #"^(FDCF)" #"^(FDF0)" - #"^(FFFD)"
        #"^(010000)" - #"^(0EFFFF)"
    ]

    NameChar:
    name-char: [#union name-start-char [#charset
        "-." #"0" - #"9" #"^(B7)" #"^(0300)" - #"^(036F)" #"^(203F)" - #"^(2040)"
    ]]
    
    
    build-late: function [spec [block!] return: [bitset!]][
        switch spec/1 [
            #charset [charset next spec]
            #union   [
                union
                    any [all [word? spec/2 get spec/2] all [block? spec/2 build-late spec/2] spec/3]
                    any [all [word? spec/3 get spec/3] all [block? spec/3 build-late spec/3] spec/3]
            ]
        ]
    ]
    
    init: has [word][
        foreach word words-of self [
            all [
                block? set/any 'spec get/any word
                issue? spec/1
                set word build-late spec
            ]
        ]
    ]
]
```

greggirwin:
The `#charset/#union` and `#delayed` examples make my Spidey-sense tingle. 

1) They look like macros, which they sort of are.

2) If generalized, to work anywhere in Red, they add a new layer to understanding and reasoning about when evaluation occurs. I'm not saying _not_ to generalize solutions to this need, that's probably the way to go; but that lazy or deferred evaluation comes with costs. e.g. `#delay` something with side effects and other code may be affected. Has limitations, like not being able to include routines, which must be understood.

3) If not generalized, it's quite complicated to do locally, compared to the `decl+does` approach, though it does solve the dual decl problem. Unfortunately, the new code is harder to get right, and quite abstract, so it's going to be copy-pasted for sure. 

Full Boris Example:
```
ctx: context [
    name-char: [#delayed charset [#"A" - #"F"]]

    init: func [word][
        all [
            block? get/any word
            #delayed = first get word
            set word do next get word
        ]
    ]
        
    fn: does [
        init 'name-char
        print mold name-char
    ]
]
ctx/fn
print mold ctx
name-char
```

Charsets and images are the only datatypes that would benefit from this, unless you have an object or block that generates a bunch of data when it's created. Is it worth getting advanced, or better to provide simple best practice examples?

greggirwin:
And if people are writing custom parsers that deal with large unicode charsets, they're probably expecting to use them every time in their app. That makes deferred charsets even more specialized, as they only apply to standard parsers/codecs. There we can ask if it's better to be clever here, or to make it easy to include modules easily, maybe even lazily since they should be entirely self contained (aside from dependencies which may also be lazily included).

rebolek:
@dockimbel @hiiamboris nice solution, I’ll adopt it. I’m so lost in XML right now I can’t think about cool optimizations anymore :-)

pekr:
I am not a measure for an opition being considered here (deep topic), but I somehow do agree with Gregg here. I also don't find Doc's nor Boris' code more readable or red-ish, not liking the #issues in a block. Looks a bit heavy, overloaded, but the overal code also more abstracted.  Why not to store a real code there and just 'do the block for an evalution? Not secure? Anyway, most average users are going to be a codec users, not caring about its internals probably, so ....

hiiamboris:
Another option: `name-char: do [does [name-char: union name-start-char charset [...]]]` that should be compiler-friendly ;)

rebolek:
> Another option: `name-char: do [does [name-char: union name-start-char charset [...]]]` that should be compiler-friendly ;)

It may be compiler friendly but otherwise looks terrible :)

dockimbel:
@greggirwin That code example you provided is not transcribing the approach properly and misses the whole point of _marking_ definitions (using `#delayed`) to evaluate lazily. You either use `init 'name-char`, directly pointing to the word to evaluate _or_ you use a marker, but not both, that does not make sense. Here are the two options:

Option 1: using names explicitly

```
ctx: context [
    name-char: [charset [#"A" - #"F"]]

    init: func [word][
            set word do get word
    ]
        
    fn: does [
        init 'name-char
        print mold name-char
    ]
]
ctx/fn
print mold ctx
```
Option 2: using markers

```
ctx: context [
    name-char: [#delayed charset [#"A" - #"F"]]

    init: has [word][
        foreach word words-of self [
            all [
                block? get/any word
                #delayed = first get word
                set word do next get word
            ]
        ]
    ]
        
    fn: does [
        init
        print mold name-char
    ]
]
ctx/fn
print mold ctx
```

The first approach is a bit simpler, but the second approach is more generic and avoids maintaining a list of words to lazy evaluate.

Lazy evaluation is fundamental in Redbol languages, it is used in many places in R2 mezz definitions, especially for VID (and maybe View too, can't remember).

greggirwin:
> @dockimbel @hiiamboris nice solution, I’ll adopt it. I’m so lost in XML right now I can’t think about cool optimizations anymore :-)

NO NO NO!!!! Don't change it to adopt this @rebolek! We're almost done and I posted a number of concerns about it.

greggirwin:
> That code example you provided is not transcribing the approach properly

Which maybe points out that it's not directly easy to understand. :^) 

> Lazy evaluation is fundamental in Redbol languages

Yes, what I said was that is adds a _new_ layer to it.

Option 1 is simpler than option 2 (by far IMO), but the original `init-charsets` is simpler still. Option 2 makes you look in multiple places to understand the code: `init` is step one when you see it called. That's very abstract code which doesn't tell you much of anything. You don't even know it applies to `#delayed` until the middle of the body. Once you grok that, now you have to go find all the places `#delayed` occurs to see what it applies to. Either that or you read the code and see `#delayed` a few times, which looks like a macro, so you may think it's a standard Red feature, then have to learn it isn't.

While the original, dual decl approach isn't great, it is at least very easy to understand.

`#delayed` is a very cool idea, and might be worth considering as a general feature, or in the context of a very large module that uses is 50 times. Here I think it's a misapplication of advanced code.

dockimbel:
> While the original, dual decl approach isn't great, it is at least very easy to understand.

So we need to write all standard libs for Red in a way a newbie can understand at the expense of maintainability? That makes no sense to me.

hiiamboris:
> > @dockimbel @hiiamboris nice solution, I’ll adopt it. I’m so lost in XML right now I can’t think about cool optimizations anymore :-)
> 
> NO NO NO!!!! Don't change it to adopt this @rebolek! We're almost done and I posted a number of concerns about it.

Why don't we start betting over who wins this round... Three lines of code are at stake! ;)

dockimbel:
> Why don't we start betting over who wins this round... Three lines of code are at stake! ;)

We can argue about a single expression for weeks when we are in a good mood! ;-)

dockimbel:
@pekr 
> Why not to store a real code there and just 'do the block for an evalution?

That's _exactly_ what we are doing... It just relies on `do next` instead of just `do`, to skip the marker (`next` is not even needed as the marker is a passive value...).

Also "real code" is very subjective in an homoiconic language. ;-)

rebolek:
> NO NO NO!!!! Don't change it to adopt this @rebolek! We're almost done and I posted a number of concerns about it.

But I already implemented it and it’s working fine :-(

greggirwin:
> So we need to write all standard libs for Red in a way a newbie can understand at the expense of maintainability?

What's important to me is maintainability, and the more people who can safely maintain things the better. 

But we can't say _which_ version is more maintainable, and each may be more maintainable by a different set of people. We can compare them side by side, and ask people which they prefer, and maybe we should. What we want to know is what _really_ makes code more maintainable, right? Once we know that (and we do know some reasons, based on research others have done), then we have a target to aim for.

I don't think the simple version is less maintainable, so it's not "at the expense of". But given the choice of simple or advanced code, including for standard libs, I will almost always vote for simplicity. It's not about newbies, it's about all of us, and some newbies may be very experience programmers in other langs. Some will like the simple version, and some will prefer the advanced version. I said exactly why I didn't like the `#delayed` version, and I also noted the main flaw with the simple version. What are the strengths of `#delayed` that make it more maintainable?

We can also ask, "What is our motto?" :^)

greggirwin:
> But I already implemented it and it’s working fine :-(

But it was also working fine before. And it was also working fine before any `init` changes were made. This has cost us time, and delayed getting docs done (because code is more fun, I know. :^) I'll take this topic to another channel though.

greggirwin:
I'll add a few code review comment here

- There is now a func called `init-charsets` which is really `init-delayed`, because it's not specific to charsets in any way. That is, if you read that func, it has nothing to do with charsets. Should we change the name? Being specific is better, but being accurate is important. Suggestions?
- `if block? Char [init-charsets]`  presents a subtle change. Where I had `if none? name-char [init-charsets]` we now have `block?` instead of `none?`. When thinking of "uninitialized" values, which is clearer? I also chose a mid-length charset name, because `char` is potentially ambiguous in more than a local scope. 
- `#delayed` is also used on `lower-letter` and a half-dozen other small ASCII charsets. Why is it used there? Does it have an impact? Not criticizing @rebolek's use of them there, though I think they don't buy us anything. What I'm pointing out is that now we ask "Why are some charsets delayed and others aren't?". When the charset defs themselves are called out, as in the simple version, it's easy to comment them as a group. When they are spread out, as they are now, that's harder. Maybe why the comment isn't included. 

Like all of you, I want the best code out there, representing Red. I also, importantly in this context, want the code _out there_. 

To that end, for those who want `#delayed`, please address the issues and concerns above and make a strong case defending it. ...before Monday because that's the end of the month. ;^)

hiiamboris:
Personally I'm for not spawning new complexity where it's not needed. In this case iterating over more than a hundred words is much more complex than dumbly setting three charsets. I'm also for not nitpicking over the PR code as long as it is acceptable and working, because we have more pressing concerns (like crashing forall, GC unable to collect OS resources, lack of `apply`, of `assert`, and a lot more.. just look around), even though I understand it's fun and relaxing to theorize about small things sometimes ;)

pekr:
OK, my last remarks too. Please note my input as an average user point of view, the decision is fully upon the Red team, of course:

1. There is imo nothing wrong, if code is more aproachable / readable even by average users. This is not a create-the-craziest-one-liner or Codegolf competition example, please.
2. To Doc's initial code - not sure why `issue!` is used, but mostly - does not the code following the opening bracket of multiline code section `[#delayed`, is agains our own code formatting rules? I mean - at least myself, I always leave the rest of the line empty, in the past I have used it for an opening comment, but not doing so anymore.
3. What I find even more redbolish, is the data / code paradigm. And for this, a block storage should be just enough. In the past I did some reporting / mapping system, to which I aplied (bind) various code sections. It feels more readable to me, unless I am deeply missing something, which might be the case, but let me try:

```
;-- charset store
charsets: [
   char            [char: charset [#"a" - #"z"]]
   restricted-char [restricted-char: charset [#"0" - #"9"]]
]

;-- access examples
print mold charsets/char
print mold select charsets 'char

;-- iteration example
foreach [section code] charsets  [
    print ["Section: " section]
    do code
] 
```

hiiamboris:
@rebolek shouldn't the codec be able to remove `&#x`-shite from attributes like this: https://github.com/unicode-org/cldr/blob/main/common/supplemental/numberingSystems.xml ?

greggirwin:
@pekr thanks for your thoughts. :+1: Good example, though it won't work in this case, because the tricky is that the context has to capture the words.

@hiiamboris agreed on prioritizing how our time is spent. Recently, when I hammer on getting things out, it's not just for that one thing, but to build the overall team mindset of focusing on deliverables.

greggirwin:
@hiiamboris the madness never ends (in CLDR too). 

1) We can TBD that feature for now (and note it in docs). It's a good test case to use for exploring options. 

2) It raises the question of how to mold chars in escaped form. I imagine they did this because there's a good chance any given rendering system will show `�/𑓙`, so we could just translate those directly, keeping them as numeric character references.

hiiamboris:
Well, yes, I actually wanted to add smth like "madness never ends" myself. Everyone around seems to go bananas these days ;)

Anyway, why would they chose to encode some chars and not the others? Maybe because the editor who wrote the file didn't have their in his own font at the time :) I don't have a better explanation. Regardless these seems valid for XML and HTML along with the absurdly long list in the link, so I think it must be supported. But then XML codec most likely will become a module or bloat the executable. IDK, no easy answer.

hiiamboris:
By my quick estimate, list is about 70kB of text.

dockimbel:
> Option 1 is simpler than option 2 (by far IMO), but the original init-charsets is simpler still.
> While the original, dual decl approach isn't great, it is at least very easy to understand.
> But given the choice of simple or advanced code, including for standard libs, I will almost always vote for simplicity.

Simplicity is not the only concern, resource usage is also a big concern when working on a library used by others and even more when it's about a language standard library.

There are also other concern that should be high on the list when it's about a standard library, like performance or code size. Though, I agreed to give up on those two for now for this (and other) codec(s). We'll address those concerns in the future, when we'll find time or get stuck in some app because of bad performances...

> But we can't say which version is more maintainable,[...]
> I don't think the simple version is less maintainable, so it's not "at the expense of".
> What are the strengths of #delayed that make it more maintainable?

One is respecting the DRY principle, the other is not. That's an objective metric for maintainability.

I'm personally fine with the Option 1 solution I've described as there are only a few bitsets concerned, so only a handful of names to maintain in sync (still Option 2 remains an objectively cleaner solution). Making this codec a module does not solve the issue, it just hides it under the carpet (unless we don't want to include this codec by default in the consoles, in which case, I'm fine with the modularization).

> We can also ask, "What is our motto?" :^)

Making every user apps more bloated, slower and resource hungry, even for the simpler apps, will not help us reach our goals of providing a better tool for programmers. We should care about the quality of the libraries we are providing. Focusing exclusively on code simplicity in the standard library (which is anyway subjective) will not bring us to a nice place.

greggirwin:
@hiiamboris I vote strongly that numeric character references be a TBD feature. See who complains. @dockimbel invoked DRY, so I'll invoke YAGNI. :^) As much as I like the idea of perfect spec support, the reality is that we have to make pragmatic choices, and if a design has largely unused features, we need a good reason to support them. It doesn't hurt anything to defer this feature, does it?

greggirwin:
@dockimbel, agreed on DRY. But can you answer, concretely, this question:

> What are the strengths of #delayed that make it more maintainable?

You say "remains an objectively cleaner solution", but I think it's only __subjectively__ cleaner. i.e. it's cleaner to you and Boris, but not to Pekr and I.

We will always have a mix of simple and advanced code, to be sure. As with the push and pull we have on performance objectives, it will make Red stronger overall. Tailor the solution to the need.

dockimbel:
>  agreed on DRY. But can you answer, concretely, this question:
> What are the strengths of #delayed that make it more maintainable?

I already answered it several times. The solution I've proposed is the only DRY one proposed here.

> You say "remains an objectively cleaner solution", but I think it's only subjectively cleaner.

DRY is objectively cleaner than non-DRY.

dockimbel:
I think I have past the point where this discussion makes anymore sense to me. I retract all my comments and propositions. You can accept this PR or tell me when to hit the "Merge" button. 

I'll just build my own consoles that are matching my needs, like everybody else it seems, so we can all be happy. ;-)

greggirwin:
OK. @rebolek please go back to the non-`#delayed` version, finish docs, and we should be good to go.

rebolek:
@greggirwin ok, reverted. I went thru all the discussions and I believe that everything should be implemented as we agreed. I need to add more tests, but that can be done separately, we can merge this now, I believe.

rebolek:
@greggirwin everything should be ready for merging.

hiiamboris:
Looks like compiler didn't swallow that :)

rebolek:
> Looks like compiler didn't swallow that :)

Can you show me the error? I tried compiling it without a problem (after fixing a few lines).

hiiamboris:
I can't reproduce it either.

greggirwin:
Is AppVeyor back to being reliable? It had issues for a while. If you can both compile, and I will do that today as I haven't, that's good enough for me.

@dockimbel I think we're ready.

greggirwin:
@rebolek please don't forget to doc how to separate metadata from the document.

rebolek:
@greggirwin Added a function `get-meta` to `%xml-tools.red` that will grab just the metadata from a document and added info about it to the docs. Also, one question:

There is a little inconsistency in the `key-val` format. Attributes are stored as `[word value]` but XMLdecl uses `[set-word: value]` format. I'm not sure why, I probably copied it from the `triples` format and nobody noticed it. Are you fine with changing it to a `word!` to be consistent with attributes?

greggirwin:
On key-val using words instead of set-words, yes, please make that change. Good catch.

greggirwin:
@rebolek %xml-tools.red is missing a `formats` definition. I added it here to test, but will let you make the change in the PR.
```
>> meta: get-meta res
*** Script Error: formats has no value
*** Where: find
*** Near : formats format [do make error! "Unknown f"] 
*** Stack: get-meta  
```

rebolek:
@greggirwin I must've deleted it by accident and it was still present in the console session. Thanks for finding it out!

hiiamboris:
Check for leaks ;)
```
Red []

leak-check: function [code] [
	print ["checking" mold/flat/part code 70]
	old: values-of system/words
	also do code (
		new: values-of system/words
		repeat i length? words: words-of system/words [
			unless any [
				:old/:i =? :new/:i
				all [									;-- newly loaded words
					i > length? old
					unset? :new/:i
				]
				all [									;@@ workaround for #5065
					any-function? :old/:i
					:old/:i = :new/:i
				]
			][
				w: pad words/:i 15
				s1: mold/flat/part :old/:i 30
				s2: mold/flat/part :new/:i 30
				print rejoin ["leak: "w"^-= "s1" -> "s2]
			]
		]
	)
]
do %xml.red
```

```
>> leak-check [load-xml ""]
checking [load-xml ""]
leak: key               = unset -> pretty?:
leak: pretty?           = unset -> false
leak: store-doctype     = unset -> (if include-meta? [repend tar
[]
>> leak-check [to-xml [a "b" #[none]]]
checking [to-xml [a "b" none]]
leak: key               = unset -> pretty?:
leak: pretty?           = unset -> false
>> leak-check [to-xml []]
checking [to-xml []]
*** User Error: "Invalid input data"
*** Where: do
*** Near : error-invalid-data
*** Stack: leak-check also to-xml encode
```

rebolek:
@hiiamboris thanks! fixed. I guess there’s nothing more left to do for now.

greggirwin:
Ping @dockimbel. Ready for merging here.

environment/codecs/XML.red
```diff
@@ -0,0 +1,1092 @@
...
+	|	sq any [att-schars | reference] sq
+	]
+	SystemLiteral: [
+		dq any [#"^^" | dq] dq
+	|	sq any [#"^^" | sq] sq
+	]
+	SystemLiteral: [
+		dq thru dq
+	|	sq thru sq
+	]
```
hiiamboris, line 219: dupe

environment/codecs/XML.red
```diff
@@ -0,0 +1,1092 @@
...
+		opt [
+			S+ not [any Char "?>" any Char] any Char
+		]
+		"?>"
+	]
+	PI: [
+		"<?" copy att-name PITarget S+ copy value to "?>"
+		2 skip
+		store-pi
+	]
```
hiiamboris, line 266: another dupe

greggirwin:
@dockimbel it's been a while now. Do you remember if this merge was reverted because we weren't sure about whether it should be standard or in red/code? Or because we still had questions about the options interface? Something else?

greggirwin:
I was looking at my draft blog post and realized XML wasn't there like I thought it was.

greggirwin:
Agreed, if it goes in red/code, we need a standard solution for loading it. Nenad may also have been concerned about the effect on compile times. But we could wrap it in `do` to avoid that. I don't remember if we tested it, or if there are R/S bits. The performance hit shouldn't be bad in that case. Oh, wait, I have some test numbers in the blog draft. I will look at that.

greggirwin:
No R/S in the code that I see at a glance.

rebolek:
Right, there’s no R/S.

PR #5069, opened 2022-02-14
FEAT: initial format module

hiiamboris:
This is roughly what Format module will look like. Code uglified a bit to accomodate for the lack of higher level constructs, but still can only be tested interpreted because requires L10N.

Should this module's context be `system/formatting` or just `formatting` or ...?


pekr:
Please call file properly - format.red, instead of module.red, which is aligned aproach to the View module. :-)

hiiamboris:
That name is already used.

PR #5072, opened 2022-02-14
FEAT: ODBC scheme

gurzgri:
This is a draft of what I can image as an ODBC scheme for Red.

The PR in it's current state is in no way acceptable for a bunch of reasons - this PR is solely to figure out if there is interest in the contribtion of such a scheme and if due to public interest it might be worth to go on and put some work into finalizing this beyond a "works for me" and "does what I need" state.

Documentation for the scheme is available at https://github.com/gurzgri/red-odbc (not to be confused with https://github.com/gurzgri/red).

- Code is unlicensed for now, this is of course subject to change to the Boost Software License together with an approriate copyright notice once the PR has chances of being accepted.
- An agreement has to be reached on the method of integration with Red: For now, it's treated as a module with needs to be required in the scripts header as `Needs: 'ODBC` and is required such in `console.red` and `gui-console.red`. Yet the code itself is put into `environment/schemes` path instead of `modules/odbc` path. 
- Code size, especially wrt odbc.reds (way to many defines not used in odbc.red) and wrt to the connection/state/info map. Many of the entries there aren't too useful to the average use case of just accessing a given known datasource.
- A fair number of tests is already supplied, but these are in no way unit test but sort of integration tests. Hard to test w/o requiring specific data sources and driver installed, I could use some guidance here on how to proceed with this.
- For now, code works best *when it works* but might not work so well yet when running into ODBC failure/error conditons (I can image possible stack related problems and with under certain circumstances not properly freed/unreleased buffers).
- I have no chance to this on macOS. I might once be able to test this on Linux, but it's not a priority for me for now.
- Tests for now haven't been done with anything but Postgres, Intersystens Caché/Iris and MS Text Driver (depending on my own default use cases)
- Caveat emptor: I'm doing this to the best of my knowledge, but I'm not a professional developer and am doing this in my spare time only.

Any feedback is welcome.

### DIFF
--- a/compiler.r
+++ b/compiler.r
@@ -105,8 +105,9 @@ red: context [
 	iterators: [loop until while repeat foreach forall forever remove-each]
 	
 	standard-modules: [
-	;-- Name ------ Entry file -------------- OS availability -----
-		View		%modules/view/view.red	  [Windows macOS Linux]
+	;-- Name ------ Entry file -------------------- OS availability -----
+		View		%modules/view/view.red	  		[Windows macOS Linux]
+		ODBC        %environment/schemes/odbc.red 	[Windows            ]
 	]
 
 	func-constructors: [
--- a/environment/console/CLI/console.red
+++ b/environment/console/CLI/console.red
@@ -8,6 +8,7 @@ Red [
 		Distributed under the Boost Software License, Version 1.0.
 		See https://github.com/red/red/blob/master/BSL-License.txt
 	}
+	Needs:  'ODBC
 ]
 
 #include %input.red
--- a/environment/console/GUI/gui-console.red
+++ b/environment/console/GUI/gui-console.red
@@ -5,7 +5,7 @@ Red [
 	Tabs:	 4
 	Icon:	 %app.ico
 	Version: 0.0.1
-	Needs:	 View
+	Needs:	 [View ODBC]
 	Config:	 [gui-console?: yes red-help?: yes]
 	Rights:  "Copyright (C) 2014-2018 Red Foundation. All rights reserved."
 	License: {
--- a/environment/schemes/odbc.red
+++ b/environment/schemes/odbc.red
@@ -1,0 +1,4239 @@
+Red [
+	Title:   "Red ODBC Scheme"
+	Author:  "Christian Ensel"
+	File: 	 %odbc.red
+	Tabs:    4
+	Rights:  "Copyright 2022 Christian Ensel. All rights reserved."
+	License: 'Unlicensed
+]
+
+
+;========================================= Red/System ==
+;
+;  ██████  ███████ ██████      ██  ██████ ██    ██  ██████
+;  ██   ██ ██      ██   ██    ██  ██       ██  ██  ██
+;  ██████  █████   ██   ██   ██    █████    ████    █████
+;  ██   ██ ██      ██   ██  ██         ██    ██         ██
+;  ██   ██ ███████ ██████  ██     ██████     ██    ██████
+;
+
+#system [
+
+#include %odbc.reds
+
+#enum odbc_common_field! [
+	ODBC_COMMON_FIELD_TYPE:             0
+	ODBC_COMMON_FIELD_HANDLE
+	ODBC_COMMON_FIELD_ERRORS
+]
+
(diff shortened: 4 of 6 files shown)


greggirwin:
Thanks @gurzgri! My next week is whacked, so I won't have time to review. It's good timing for the discussion on integration, as we're dealing with that right now for other features (XML and more).

rebolek:
Wow, that’s a lot of work certainly! I’m going to try it once I have some time.

hiiamboris:
wow man you love blank lines :)

GiuseppeChillemi:
Very high interest here.

dockimbel:
A few comments:

> An agreement has to be reached on the method of integration with Red: For now, it's treated as a module with needs to be required in the scripts header as Needs: 'ODBC and is required such in console.red and gui-console.red. Yet the code itself is put into environment/schemes path instead of modules/odbc path.

We're working right now on defining how to properly handler third-party libs until we get a full module system and a package manager. We should propose a new organization in the next days.

> Code size, especially wrt odbc.reds (way to many defines not used in odbc.red) and wrt to the connection/state/info map. Many of the entries there aren't too useful to the average use case of just accessing a given known datasource.

FYI, defines are matched linearly in parse rules by the preprocessor, so increasing defines number has a direct impact on compilation time (among my wishes for future Parse features: a matching mode backed by a hashtable). So, you should comment all defines that are not used and switch to enums for most of the used ones. Also enums are context-aware as an additional benefit.

> A fair number of tests is already supplied, but these are in no way unit test but sort of integration tests. Hard to test w/o requiring specific data sources and driver installed, I could use some guidance here on how to proceed with this.

There is a CSV engine mode in ODBC which could be used for testing purposes, that should make it a bit easier. Anyway, we need to also define how third-party modules and libs should be tested in the next days.

Last, a remark about your R/S code. I see you are often using `or` and `and` in conditional expressions, but R/S supports `any` and `all` like in Red:

```
if (<expr1>) or (expr2) [...]
```
can be written as:
```
if any [<expr1> <expr2>][...]
```

greggirwin:
Doesn't seem like too many blank lines to me, at least for humans. :^)

gurzgri:
Thanks for your comments and tips, @dockimbel

The defines have been replaced by enum labels, all unused ones have been commented out (leaving only a ridiculous tiny number which are actually in use).

With the enum labels in place plus only a handful of own ODBC_* macros, the conditional expressions too have been overhauled, improved readability quite a bit.

GiuseppeChillemi:
> There is a CSV engine mode in ODBC which could be used for testing purposes, that should make it a bit easier. Anyway, we need to also define how third-party modules and libs should be tested in the next days.

Using MUNGE with SQL Server I have to convert every CR of long text fields of a table because there is no way to understand if the CR related to the text or means "end of row".  This situation could happen with the ODBC CSV engine. (Just my 2 cents)

gurzgri:
Finally, I (think I) grasped the concept of Red/System contexts and envision how I should have organized all this from the very beginning ... That will take some serious refactoring and will mostly be boring diligent work. Slow but steady wins the race, I guess.

Anyways, I did some mayor improvements over the last days, so anyone who wants to try this please make sure you're up to date with latest commits. 

And rewriting this for Red/System contexts will be totaly opaque from a user's perspective, as it won't break the current scheme api at all.

gurzgri:
Refactored to use #enum in favour of #define, no more polluting the R/S global context now.
Only a handful of ODBC_-prefixed macros and two R/S global words SQL and ODBC remain.
Makes code look more organized and friendlier to the eye.

bubnenkoff:
Is there any chance that this pull request will be merged in main branch?

greggirwin:
> Is there any chance that this pull request will be merged in main branch?

Not at this time. The best place for it right now is red/code, so people can pull it from there. We're also considering a new interim system to help with that, so you could use `needs` in the header.

bubnenkoff:
@greggirwin thanks for response!
Could you explain what stop from merging? 
- Red need some improvement before it can be done?
- Red do not need in out of the box?
- odbc driver need some improving?
- ... 

I am asking to better understand situation to be able to plan the job.

Issue #5009, opened 2021-12-17
Issues in paths are not lexed


hiiamboris:
**Describe the bug**

Sometimes it's useful to have #issue instead of word in the data (came up in XML codec). But currently one needs parens for that:
```
>> b: [#x 1 #y [#z 2]]
>> b/#y/#z
*** Syntax Error: (line 1) invalid issue at #y/#z
*** Where: case
*** Stack: load 
>> b/#x: 3
*** Syntax Error: (line 1) invalid issue at #x: 3
*** Where: set
*** Stack: load 
>> b/(#y)/#z
== 2
>> b/(#x): 3
== 3
```
I don't think parens are a necessity, since `/` and `:` chars are invalid in issues anyway, and path access is more useful than the error behavior. R2 allows `b/#y/#z`.

**Expected behavior**

Issues work inside paths without parens.

**Platform version**
```
Red 0.6.4 for Windows built 6-Dec-2021/19:27:05+03:00  commit #252d78d
```

hiiamboris:
See also #5007 on compiler problems

PR #5085, opened 2022-02-21
FEAT: String interpolation

hiiamboris:
This PR is technically a part of our [Format effort](https://github.com/hiiamboris/red-formatting/discussions), but:
- it's not tied to anything exported by Format module
- it's widely useful outside of Format scope and should not require one to include the latter

## Intro

[String interpolation is used in most languages](https://en.wikipedia.org/wiki/String_interpolation#Examples) as a readable way of inserting values into a string.

Though the first question everyone of us may have is *"what, another `rejoin`?"*, if `rejoin` was good enough for me or Gregg, this design would never have been born.

Here are just a few common examples written using `rejoin` function: 
```
avcmd: rejoin [{"} player {" "} vfile {" --audio-file "} afile {"}]
print rejoin ["Download OK: '" remote "' [" info/Content-Type ", " info/Content-Length " bytes]"]
do make error! rejoin [token1 " cannot follow " token2 " in the " name " part of " mold .mask]
```
Try to look at these expressions and visualize how the resulting string will look like, and if I've got all spaces and quotes right.
Not a human task, eh? Even telling the strings from code requires syntax highlighting or quote counting or guessing.

<details>
<summary>These are equivalent expressions written via string interpolation:</summary>

```
avcmd: #rejoin {"(player)" "(vfile)" --audio-file "(afile)"}
#print "Download OK: '(remote)' [(info/Content-Type), (info/Content-Length) bytes]"
#error "(token1) cannot follow (token2) in the (name) part of (mold .mask)"
```
Pretty clear, right?

</details>

Another, even more solid reason for this design is [**translation**](https://gitlab.com/hiiamboris/red-mezz-warehouse/-/blob/master/composite.md#localization). Whatever way we choose to translate our software, we have to give *whole* messages to the translator. 

For more **background info** you can read [related Format discussion](https://github.com/hiiamboris/red-formatting/discussions/20), [preliminary design document](https://gitlab.com/hiiamboris/red-mezz-warehouse/-/blob/master/composite.md) and the [original sad-emoji dialect design](https://github.com/greggirwin/red-formatting/blob/master/formatting-functions.adoc#composite), but I'll try to explain the main points here.

List of so far **registered use cases** can be [found here](https://github.com/hiiamboris/red-formatting/discussions/20#discussioncomment-2206751) but those are only mine. Maybe Gregg can find his own somewhere. [Below that](https://github.com/hiiamboris/red-formatting/discussions/20#discussioncomment-2210811) is a description of design implemented in this PR - a result of my experience with interpolation.

**Why a macro?**

Since input is a string, we don't get `word!`s from it. Without words, nothing is bound during context and function creation. Then the moment we extract words from the string, we can only bind them to the global scope, or to an explicitly provided list of contexts. [Gabriele expressed the concern very clearly](https://i.gyazo.com/ef113c3532289f0acf999c8fd5fe122a.png)

Macro helps us avoid this trap. It gets expanded and produces words before those words are bound, resulting in code that *just works*.

We also get the benefit of speed: if program is compiled, all macros are expanded at compile-time.

Function approach is still provided for advanced users as it has it's merit: macro cannot work on message that is generated at run time (it can, but it has no knowledge of contexts where to bind words).

**Syntax**

Every paren inside the string becomes code:
`#rejoin "(x) + (y) = (x + y)"` -> `rejoin ["" (x) " + " (y) " = " (x + y)]`
Why parenthesis? Because as everywhere else in Red, it visually hints at evaluation.

To treat a paren literally, it is escaped *inside* by a backslash:
`#rejoin "total (n) hits (\ratio: (100% * n / total))"` -> `rejoin ["total " (n) " hits (ratio: " (100% * n / total) ")"]`
This relies on backslash being invalid in Red, so if there's a plan to leverage `\` then another escape sigil must be found.

Original string type is preserved (leading `""` also serves this purpose):
`#rejoin %"file-(n).(ext)"` -> `rejoin [%"file-" (n) "." (ext)]`
`#rejoin <x (y)=(z)>` -> `rejoin [<x > (y) "=" (z)]`

**Extensions**

1. Format module will also include a `#format` macro, backward-compatible with `#rejoin` but with two more features:
   - it will convert `(expression as mask)` into `(format (expression) mask)`
   - it will support per-message locale specification:
     `#format/in "you've spent (spent as {$0.00}) of (spent + left as {$0.00})" locale`
     will be preprocessed into
     `#rejoin "you've spent (format/in (spent) {$0.00} locale) of (format/in (spent + left) {$0.00} locale)"`
     This will require #5009 to be solved.
  
2. It is expected that users will write their own macros for common cases based on `#rejoin`, e.g.:
   ```
   #macro [#log any-string!] func [[manual] s e] [insert remove s [log #rejoin] s]

   #log "(now) test message: (1) + (2) = (1 + 2)"		;) `#log` now is synonymous with `log #rejoin`
   #log "(now) test message: (2) * (3) = (2 * 3)"
   ```
3. Another extension idea concerns `#error` macro. Right now the only fully custom error we have is the User Error. I propose adding another fully custom error into every error category, so we could produce Script, Math, other errors with custom messages. `#error` should then default to Script error, with `#error/math`, `#error/syntax`, `#error/user` etc forms changing the error type.
4. For cases when template is only known at runtime (e.g. report generation with user-defined template), `rejoin` function is extended to accept `any-string!` argument of exactly the same syntax as the macro:
   - `rejoin ` -> ``
   - to overcome the binding issue, `rejoin/with` accepts one or more contexts to bind produced expressions before reducing them
   - as a feature, `rejoin/trap` allows one to replace evaluation errors with some text (per original Gregg's design)
   - `/with` and `/trap` only apply to string case and have no effect on block argument

5. https://github.com/red/REP/issues/112 should be considered to bring URL support to this design.

P.S. I need some help reducing the docstrings :)

### DIFF
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -1093,7 +1093,7 @@ transcode-trace: func [
 ]
 
 with: function [
-	"Bind code to a given context or list of contexts (in order from first to last)"
+	"Bind block to a given context or list of contexts (in order from first to last)"
 	ctx [any-object! function! any-word! block!]
 		"Block [x: ...] becomes a context, [x 'x ...] is reduced and passed sequentially to BIND native"
 	block [block!] "Is not evaluated"
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -1092,15 +1092,72 @@ transcode-trace: func [
 	transcode/trace src :system/lexer/tracer
 ]
 
-;--- Temporary definition, use at your own risks! ---
-rejoin: function [
-	"Reduces and joins a block of values."
-	block [block!] "Values to reduce and join"
+with: function [
+	"Bind code to a given context or list of contexts (in order from first to last)"
+	ctx [any-object! function! any-word! block!]
+		"Block [x: ...] becomes a context, [x 'x ...] is reduced and passed sequentially to BIND native"
+	block [block!] "Is not evaluated"
 ][
-	if empty? block: reduce block [return block]
-	append either series? first block [copy first block] [
-		form first block
-	] next block
+	case [
+		not block? :ctx  [bind block :ctx]
+		set-word? :ctx/1 [bind block context ctx]
+		'otherwise [
+			foreach item reduce/into ctx clear [] [
+				bind block :item
+			]
+			block
+		]
+	]
+]
+
+context [
+	non-paren: charset [not #"("]
+
+	trap-error: function [on-error [function! string!] :code [paren!]] [
+		error: try [return do code]
+		either function? :on-error [do [on-error error]] [on-error]
+	]
+
+	set 'rejoin function [
+		"Reduces and joins a block of values or interpolates a template string"
+		values [block! any-string!] "Values to reduce and join or string with parenthesized expressions"
+		/with "Bind expressions loaded from string to a list of contexts"
+			ctx [block!] "List in format accepted by WITH function"
+		/trap "Trap string evaluation (but not loading) errors and replace with text"
+			on-error [function! string!] "Replacement text, or func [error [error!] return: [string!]]"
+	][
+		unless block? values [
+			string: as string! values
+			block: clear []
+			parse string [
+				collect into block [
+					keep (as values "")					;-- ensures the resulting type
+					any [
+						keep copy some non-paren		;-- any text
+					|	keep [#"(" ahead #"\"] skip		;-- escaped opening paren
+					|	pos: (set [value: end:] transcode/next pos) :end	;-- paren expression
+						keep (:value)
+					]
+				]
+			]
+			
+			if with [system/words/with ctx block] 			
+			if trap [									;-- each result has to be evaluated separately
+				forall block [
+					if paren? block/1 [
+						block: insert block [trap-error :on-error]
+					]
+				]
+			]
+			
+			values: block
+		]
+		
+		if tail? block: reduce values [return block]
+		append either series? first block [copy first block] [
+			form first block
+		] next block
+	]
 ]
 
 sum: func [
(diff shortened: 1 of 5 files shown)


hiiamboris:
How it could in theory work without a macro:

1. Special lexer string-like syntax that gets transcoded as a block, e.g. `(x) + (y) = (x + y)` -> `["" (x) " + " (y) " = " (x + y)]

   Seems to me much more complex solution than a macro.

2. Introduce lexical scoping to the language, and let strings infer it so they can be expanded properly.

   Extremely complex solution with uncertain outcomes. But may resolve the loops leaking words issue?

3. Let string expansion routine access the stack and automatically bind the result to contexts it finds in the stack. It will only bind to entered functions and `make object` contexts, ignoring arguments pushed to the stack.

   Seems simple enough. But contradicts our definitional scoping model where words are bound at entity creation time. Resulting block will be bound not where the string appears, but where it's expanded. Which arguably for strings may be a desired outcome, e.g. we define a set of template strings somewhere, then fetch them by words and expand in different places automatically binding to different contexts. But still an inconsistency: e.g. if a function defined in an object uses such expansion, it will have access to function words but not to object words, because we most likely have left the `make object` scope by that time.
   
   Also this option will need a function flag, akin to `[no-trace]`, e.g. `[no-bind]` to tell the expand function to skip itself. E.g. a `log` wrapper that calls `expand` would want to hide its own context from being bound to. A possible wrapper around `log` wrapper as well, and so on.

loziniak:
Hi! One remark – in Red an escape sign is usually a caret: `^`. What's the downside of using (more idiomatic in my opinion) `^(` instead of `(\` to escape parens?

hiiamboris:
```
>> "^("
*** Syntax Error: (line 1) invalid string at ("
*** Where: set
*** Near : transcode/trace buffer :delimiter-lex
*** Stack: error? load
```
For more details please read https://codeberg.org/hiiamboris/red-common/src/branch/master/composite.md#syntax

PR #5142, opened 2022-05-18
FEAT: faster/simpler empty? function

hiiamboris:
benchmark source:
```
Red [] 

do %../common/clock.red

empty1: func [
    {Returns true if a series is at its tail or a map! is empty} 
    series [series! none! map!] 
    return: [logic!]
][
    case [
        series? series [tail? series] 
        map? series [series = #()] 
        none? series [true]
    ]
]

empty2: func [
    {Returns true if a series is at its tail or a map! is empty} 
    series [series! none! map!] 
    return: [logic!]
][
	either series [zero? length? series][true]
]

do [loop 5 [
clock/times [empty1 none   ] 1000000
clock/times [empty1 []     ] 1000000
clock/times [empty1 [a b]  ] 1000000
clock/times [empty1 #()    ] 1000000
clock/times [empty1 #(a: b)] 1000000
print ""
clock/times [empty2 none   ] 1000000
clock/times [empty2 []     ] 1000000
clock/times [empty2 [a b]  ] 1000000
clock/times [empty2 #()    ] 1000000
clock/times [empty2 #(a: b)] 1000000
print "---"
]] 
```

### DIFF
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -64,11 +64,11 @@ quit: func [
 ]
 
 empty?: func [
-	"Returns true if a series is at its tail or a map! is empty"
-	series	[series! none! map!]
+	"Returns true if data is a series at its tail or an empty map"
+	data	[series! none! map!]
 	return:	[logic!]
 ][
-	either series [zero? length? series][true]
+	either data [zero? length? data][true]
 ]
 
 ??: func [
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -68,11 +68,7 @@ empty?: func [
 	series	[series! none! map!]
 	return:	[logic!]
 ][
-	case [
-		series? series [tail? series]
-		map? series [series = #()]
-		none? series [true]
-	]
+	either series [zero? length? series][true]
 ]
 
 ??: func [


greggirwin:
If we're going to do this, and it's a nice optimization, we need to either rename the `series` arg or use `none? series`. It took me longer than I care to admit to parse it, as I kept reading `either series` as `either series?`. Even the original (which is nice and clear) suffers from this at `map? series`.

Here's another variation, which ekes out a hair more performance for block/map, at the expense of none.
```
empty3: func [
    {Returns true if a series is at its tail or a map! is empty} 
    data [series! none! map!] 
    return: [logic!]
][
	; 0 is the none case, because `none = length? none`
	zero? any [length? data  0]
]
```

Runs are very consistent here.
```
1.12 μs [empty1 none]
0.73 μs [empty1 []]
0.73 μs [empty1 [a b]]
0.95 μs [empty1 #()]
0.96 μs [empty1 #(a: b)]

0.43 μs [empty2 none]
0.65 μs [empty2 []]
0.64 μs [empty2 [a b]]
0.63 μs [empty2 #()]
0.63 μs [empty2 #(a: b)]

0.63 μs [empty3 none]
0.57 μs [empty3 []]
0.57 μs [empty3 [a b]]
0.56 μs [empty3 #()]
0.56 μs [empty3 #(a: b)]
```

I do like the complete clarity of the original, but am OK with this optimization as well. `Empty?` is a high level call, but may be used a lot. Your call @dockimbel or @qtxie.

hiiamboris:
Nice solution as well :+1:

hiiamboris:
I also had the `series` moment when reading my own code :)

hiiamboris:
In Rebols `empty?` is a native. Red has funny natives like `positive?` which are very rarely used, but not `empty?`.

GiuseppeChillemi:
> In Rebols `empty?` is a native. Red has funny natives like `positive?` which are very rarely used, but not `empty?`.

Just a note, Rebol `positive?` has it too and I use it to check the boundaries of integers passed as args for series index. It makes the reading the language more natural, even if `value >0` could be used.

Issue #5108, opened 2022-03-18
Extra block copy in make map!


hiiamboris:
**Describe the bug**

Looks like map is copying it's own spec on `make`:
```
>> recycle/off
>> prof/each/times [copy #(x: 1 y: 2) make map! [x: 1 y: 2] copy [x: 1 y: 2]] 10000
<10000> 20%      .00076ms         272 B [copy #(x: 1 y: 2)]
<10000> 40%      .00152ms         364 B [make map! [x: 1 y: 2]]
<10000> 40%      .00151ms          92 B [copy [x: 1 y: 2]]
>> 364 - 92
== 272
```

**To reproduce**

`s: stats loop n: 10000 [make map! [x: 1 y: 2]] stats - s / n`

**Expected behavior**

272 bytes from `make` as from `copy`

**Platform version**
```
Red 0.6.4 for Windows built 11-Mar-2022/2:40:53+03:00  commit #27f71f9
```

qtxie:
If insert a `recycle` before `stats - s / n`, I got this:
```
>> s: stats loop n: 10000 [make map! [x: 1 y: 2]] recycle stats - s / n
root: 3768/6544, runs: 0, mem: 3284144 => 1808236, mark: 2.0ms, sweep: 1.0ms
root: 3768/6544, runs: 1, mem: 3984004 => 1808376, mark: 1.0ms, sweep: 2.0ms
== -1.084
```

hiiamboris:
What would be the point? Measuring memory that's not freed by recycle?

qtxie:
Sorry, I misunderstood the issue. So it's about the implementation detail. You want that `make` takes the same amount of memory as `copy`?

hiiamboris:
Yeah. Extra block is just unnecessary GC pressure in this case.

PR #5143, opened 2022-05-18
FIX: issue #5108 (Extra block copy in make map!)

ALANVF:
Seemed simple enough, though I'm not super experienced with R/S so I might've missed something

### DIFF
--- a/runtime/datatypes/map.reds
+++ b/runtime/datatypes/map.reds
@@ -234,8 +234,11 @@ map: context [
 		]
 
 		if zero? size [size: 1]
-		blk: block/make-at as red-block! stack/push* size
-		if blk? [block/copy as red-block! spec blk null no null]
+		either blk? [
+			blk: block/clone as red-block! spec no no
+		][
+			blk: block/make-at as red-block! stack/push* size
+		]
 		make-at as red-value! blk blk size
 	]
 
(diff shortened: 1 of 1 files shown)


dockimbel:
Your fix seems fine, `block/copy` relies on `_series/copy` which are parts of `copy` action implementation, so higher-level than `block/clone`, but for this use-case, I don't see any difference in behavior.

I only reviewed the code I didn't test it. Did you run the tests from #5108 to check the fix efficiency?

ALANVF:
I ran the test used for looking at the allocated memory, and it appeared to be fixed by my changes

hiiamboris:
We also usually add regression tests (see any other PR).

ALANVF:
I know, but this was small enough that I didn't think it was necessary

greggirwin:
A lot of changes seem that way, but they all add up. It's also possible that someone will come along later and remove the copy ("Hey, I optimized something!"), unless it's clearly documented that it's needed. Since there are no comments, which are also good to add, because that informs future work that may be based on code, tests seem like a good idea.

dockimbel:
In general, we require regression tests for PRs that are fixing bugs. Though, we can skip the test for this kind of code improvement, which is not a bug, just inefficient implementation. Remember this is v0.6.4 and still many design decisions are not yet made or are subject to change until 1.0. So let's not waste time on writing tests for such cases, on code that might be gone in the future.

Issue #4429, opened 2020-05-05
[View] Rich-text `para` facet resets default settings when present


hiiamboris:
**Describe the bug**

Notice how specifying `para` immediately centers the text:

| Code | Layout |
| --- | --- |
| `view [rich-text {123^/234^/345^/456^/567^/678^/789}]` | ![](https://i.gyazo.com/fcac1d47d3cc19878c5c817874db1fd0.png) |
| `view [rich-text {123^/234^/345^/456^/567^/678^/789} para []]` | ![](https://i.gyazo.com/9729cc869062e7b40487e343ac364ffe.png) |
| `view [rich-text top {123^/234^/345^/456^/567^/678^/789}]` | ![](https://i.gyazo.com/6f55c5c853bc5e906624816220e96714.png) |
| `view [rich-text right {123^/234^/345^/456^/567^/678^/789}]` | ![](https://i.gyazo.com/0fa73c79cc98098bfa3cd994a5f359e3.png) |

**To reproduce**
```
view [rich-text {123^/234^/345^/456^/567^/678^/789}]
view [rich-text {123^/234^/345^/456^/567^/678^/789} para []]
view [rich-text top {123^/234^/345^/456^/567^/678^/789}]
view [rich-text right {123^/234^/345^/456^/567^/678^/789}]
```

**Expected behavior**

1 & 2 should look the same. As for what should be the default alignment, not sure. Likely top-left.

**Platform version (please complete the following information)**
```
Red 0.6.4 for Windows built 18-Mar-2020/14:57:39+03:00 commit #101a64f
```

9214:
https://github.com/red/red/pull/4385

Issue #4280, opened 2020-02-04
[View] custom font resets alignment for BUTTON, CHECK and RADIO


9214:
**Describe the bug**

On Windows, changing any font settings of the `button`, `check` or `radio` face resets its text alignment.

**To reproduce**

```red
view [button "THIS" right button "THAT" right font-color red]
```

On a quick glance, it looks like `process-custom-draw` [applies](https://github.com/red/red/blob/master/modules/view/backends/windows/events.reds#L828) DrawText transforms without honoring `para!` settings.

**Expected behavior**

I expect text alignment to remain unaltered.

**Screenshots**

Without custom font:

![image](https://user-images.githubusercontent.com/15716466/73770328-815b9100-477c-11ea-83d1-a6226d651011.png)

With custom font:

![image](https://user-images.githubusercontent.com/15716466/73770235-596c2d80-477c-11ea-9bb7-c421042e5527.png)

**Platform version**

```
Red 0.6.4 for Windows built 4-Feb-2020/16:11:59+01:00 commit #811db99
```

hiiamboris:
I suggest reopening as while the provided snippet works, the default (unspecified) alignment is still overridden:
![](https://i.gyazo.com/dfc7d7969601489c65eee4187099b713.png)
`view [button 80x50 "THIS" button 80x50 "THAT" font-color yello]`

```

PR #5152, opened 2022-06-15
FIX: para breaks `toggle` alignment

ALANVF:
The text alignment for `toggle` objects is ignored/reset when its `para` is configured.

Short example:
```red
view [ button 25 bold "A" toggle 25 bold "B" ]
```
Output:
image


hiiamboris:
#4280 #4429

Issue #5123, opened 2022-04-15
stack overflow on unique 


bubnenkoff:
**Describe the bug**
```
out: copy []
loop 10000 [collect/into [keep [[a: 1]] ] out] 
 [[a: 1] [a: 1] [a: 1] [a: 1] [a: 1] [a: 1] [a: 1] [a: 1] [a: 1] [a: 1] [a: 1] [a: 1] [a: 1] [a: 1] [a: 1...
unique out
```
error:
```
*** Internal Error: stack overflow
*** Where: unique
*** Near : out
*** Stack:
```

**Expected behavior**
Getting unique list of blocks. In this case is: `[a: 1]`

More examples: https://gitter.im/red/bugs?at=625919878db2b95f0ab88c99

**Platform version**
```
>> about
Red 0.6.4 for Windows built 3-Feb-2022/20:16:35+03:00  commit #fbdf84d
```

GiuseppeChillemi:
I confirm this bug, I can't use `unique` and `difference` on data coming from the server of my company because I am always hit by a stack overflow.

Thread here: https://gitter.im/red/help?at=62c871ea9a314a3ec44f658c
(Note it starts with `unique set1 set2` but should be `difference`)

PR #5166, opened 2022-07-20
FIX: issue #5123 (stack overflow on unique)

qtxie:


qtxie:
I tried to remove the `stack/push` and use just the `result: as red-value! blk`. It didn't work. Seems there are some places where rely on this temp result stay on the stack.

PR #5192, opened 2022-08-31
FEAT: clock function for profiling in the small

hiiamboris:


environment/functions.red
```diff
@@ -1116,13 +1116,37 @@ dt: function [
 	body	[block!]
 	return: [time!]
 ][
-	t0: now/precise
+	t0: now/precise/utc
```
hiiamboris, line 1119: Note: with /utc it's 2x faster than without (and `now` call is significant in profiling)

dockimbel:
That's a nice PR to merge, thanks!

greggirwin:
From #4446, here are some `clock` results.
```
0.99 μs	[find/same/only bs pick bs 9]
1.98 μs	[find/same/only bs pick bs 90]
17.0 μs	[find/same/only bs pick bs 900]
170 μs	[find/same/only bs pick bs 9000]
1.74 ms	[find/same/only bs pick bs 90000]

1.00 μs [find/same/only bs pick bs 9]     
4.00 μs [find/same/only bs pick bs 90]    
26.0 μs [find/same/only bs pick bs 900]   
245 μs  [find/same/only bs pick bs 9000]  
2.43 ms [find/same/only bs pick bs 90000] 
```
I pulled these because, to me, the formatting and use of units makes it harder to see times relative to each other, as scale goes. Having the result in ascending order helps dramatically of course, so you know the times are getting larger. But the optional floating decimal and units both require attention. The new `profile` func uses a format like the old one I did, and @giesse improved upon.

[mine](https://gist.github.com/greggirwin/908d44dc069ed84cf69f053e1308390d)
[Gabriele's](https://gist.github.com/giesse/1232d7f71a15a3a8417ec6f091398811)

Naming wise, we have `dt`. That's too short and meaningless for a global word, the rationale being that it's a console func, like `cd`. To that I say "meh". :^) It's aliased as `time-it`, which isn't great either, but at least has meaning. `Clock` adds unit forming and code output, and reverses the logic of show vs return with `/delta` which I don't like at all as a name for what it does, or that it returns a float you have to know is in `ms`. `/times` also seems confusing as related to `time!` use in Red.

As a side note, while playing in the console to write this comment, I used `clock` in a `profile` call, which resulted in console output, because I didn't think to use `/delta`. I finally got this result:
```
...0.0 μs	[]
0.0 μs	[]
#1   loop             | 100001     | 0:01:45.3599
#2   +                | 100000     | 0:00:00.214308
#3   append           | 100000     | 0:00:00.328473
#4   set/any          | 100000     | 0:00:00.186415
#5   clear            | 100000     | 0:00:00.205348
#6   head             | 100000     | 0:00:00.17718
#7   clock            | 100000     | 0:01:42.8304
```

While `clock` adds human info over `time-it` (always welcome), I think this was a hasty merge and the function should be reconsidered with some thought given to consistency and purpose of various profiling/measurement needs. e.g. mine was written to compare alternative implementations or related pieces of code at a higher level. The new instrumentation based `profile` shows detailed evaluation results at the lowest level. I also made a "non-local timing markers" system which proved helpful in some real world situations, and which is a different approach entirely. We also want to consider future instrumentation possibilities, both at the low level, like DTrace works, but also at the high level, in what is now called Observability. It ties to aggregators too.

Performance measurement and monitoring is a domain. Let's tackle it that way, rather than taking an ad hoc approach.

hiiamboris:
@greggirwin do you have any *specific* improvement in mind?

greggirwin:
My suggestion is the last line in my comment. _Design_ these related function so they're cohesive.

Cherry picking a single example ignores the aspects I pointed out. Play fair, and be constructive. My point was that this format doesn't make the scale visible at a glance. You have to look both for the decimal point and at the unit, then scale in your head. This example gains from timing exactly the same code at 10x for each pass, but you could be off by an order of magnitude and not see it easily. We can also argue for automatic separators in numbers here, for the loop counts in the code. That will also help.
```
1.00 μs [find/same/only bs pick bs 9]     
4.00 μs [find/same/only bs pick bs 90]    
26.0 μs [find/same/only bs pick bs 900]   
245 μs  [find/same/only bs pick bs 9000]  
2.43 ms [find/same/only bs pick bs 90000] 
```
Here's the same data as times. Is this best? Maybe not for micro profiling, because that use case should never leverage hours or minutes, and often not even seconds, so they are extra noise. `Profile` makes great use of that info though.
```
0:00:00.00000100
0:00:00.00000400
0:00:00.00002600
0:00:00.00024500
0:00:00.00243000
```
I'm 95% sure we don't want separators after the decimal point, but you can still see the leading number shifting by column. 

For this level of profiling, do fractional microseconds help? Can we round and still be accurate enough? Is it confusing if we omit hours (still valid times, lexically).
```
00:00.000001
00:00.000004
00:00.000026
00:00.000245
00:00.002430
```
Still a bit noisy, if we're almost always talking sub-second results. Leave out everything we don't need for consistent output up to 10 seconds.
```
0.000001
0.000004
0.000026
0.000245
0.002430
```
From here we ask what kind of profiling do we do, when and where they're used, and how best to present results for each case. To start, profiling should probably always return raw data, with an option to format it for display. That may also be separate from printing it, so you can gather formatted results for reports and such. Given raw data, various formatters can be plugged in. Do we go that far or just provide standard formatters? Even if we do the latter, a plugin model makes it easier to play with alternatives. And we can provide standard versions, like with `parse-trace`. Maybe we want to keep them simple, but somebody writes a smart one that looks at the resulting time ranges and adjusts accordingly, or even graphs the result.

This is an enormously valuable domain IMO, where (I think but haven't researched) Red could stand out against other langs. Also important when we get to packages and modules, because the system can then benchmark them automatically and consistently. This leads to thinking about test definitions not just spec'd for correctness, but for applicability at different scales. e.g., if you don't need massive scale, you can choose a lib that's easier for you to understand and maintain should the need arise.

hiiamboris:
I think you misunderstand the scope of this function, which is: 80% for playing with small snippets in console (and that's why having `/times` is not an issue - you will just have it again when you press Up key), 20% for measuring very fast parts of code (by repeating them many many times). So your arguments simply don't apply here, and it has nothing to do with `profile` which is a completely different scope.

I have a much more advanced profiler than your one or the one in the runtime. With the features you describe. But I never wanted any of that for `clock` mezz (and I use it a lot), even RAM measurement is not important in it. It's only requirements are: don't introduce ANY overhead, be precise, keep simple, keep readable. Think of it as a quick humans interface to `dt`, which is the programmatic profiling primitive.

Precision of `clock` is able to "see" the baseline of interpreted `loop` itself, and each token's evaluation time:
```
>> clock/times [] 1e8
0.02 μs	[]
>> clock/times [1] 1e7
0.05 μs	[1]
== 1
>> clock/times [1 2] 1e7
0.08 μs	[1 2]
== 2
```

greggirwin:
> I think you misunderstand the scope of this function

I understood it, I just don't think it's a great func, but we need to look at the big picture. I'm not saying we shouldn't have a func _like_ this, but that it's part of that bigger picture which we want to be thought out.

hiiamboris:
It is thought out and battle proven. If you wanna improve it, please give *concrete* proposals ;)

greggirwin:
I do care about the battle, but I care more about the war. Up to @dockimbel now about how to move forward here.

PR #5211, opened 2022-09-16
FIX: #2586

hiiamboris:


hiiamboris:
I think I didn't push this previously because of https://github.com/red/red/commit/99ef4039321591b08119d8cb0dfab920fee4ad72 (lots of other funcs that are either buggy or not in line with float acceptance)

hiiamboris:
Buggy only these:
```
>> pick "abc" 1.5
== none
>> poke "abc" 1.5 x
*** Script Error: value out of range: 0
*** Where: poke
*** Near : x
*** Stack:  
```

PR #5219, opened 2022-09-29
FIX: increased tolerance for antialiased text color

hiiamboris:


greggirwin:
What about `/within` rather than `/tol` ?

hiiamboris:
Does it matter?

greggirwin:
Naming always matters, unless you're the only one who ever has to look at it.

hiiamboris:
Whole test system needs redesign from scratch. And this test's lifecycle will end once we get to View improvements. I'd rather not put lipstick on a pig.

greggirwin:
Please don't make this someone else's fault. Every choice can make things better or worse. Should we not always strive to write the best code we can? What kind of code and projects do you appreciate and respect? What do we want people to think about Red?

hiiamboris:
I just prefer to spend my time where it adds value is all :)

greggirwin:
So always give things good names to start with.

hiiamboris:
Indeed. I took note of your critique and agree with it.

PR #5243, opened 2022-11-23
FEAT: Win: adds read-only flag for AREA face

qtxie:
view [area 100x100 read-only "1some text some text^/2some text some text^/3some text some text^/4some text some text^/5some text some text^/6some text some text^/7some text some text^/8some text some text^/9some text some text"]


dockimbel:
What about the other GUI backends?

dockimbel:
I'm not in favor of this platform-specific option leaking up to VID level. I would prefer a generic platform-independent way in VID/View to disable keyboard keys input on widget where such inputs produce changes.

hiiamboris:
One can edit an area with a pointing device (by selecting things and right click menu).

qtxie:
> What about the other GUI backends?

The `area` control has a `read-only` mode on other GUI backends too. 

> I would prefer a generic platform-independent way in VID/View to disable keyboard keys input on widget where such inputs produce changes.

Yes. That's a better way.

PR #5266, opened 2023-01-13
Win resize update

luce80:
This change only inverts execution of a few lines so that win face/size is updated **before** creating a `resizing` event.
To see before and after the fix:
```
print "" ; open console fore debug
view/flags/options [t: box "hello" red 200x200 ] [resize] [
	actors: object [
		on-create: func [face][print ["create" face/data: face/size]]  ; init value
		on-focus: func [face event][print "focus" face/data: face/size]  ; store old size
		on-resize: func [face event][print "resize" face/actors/on-resizing face event] ; forward
		on-resizing: func [face event /local siz][prin ["resizing" face/size event/face/size t/size]
			siz: face/size - face/data    ; compute size difference
			face/data: face/size          ; store new size
			
			t/size: t/size + (siz * 1x1)
			print [" resizing" t/size]
			t/color: random 255.255.255 ; show that the problem is only with size
		]
	]
]
```
After the window has opened press `Alt+space` then `d` then `arrow right` one or more times.


qtxie:
You can use `event/offset` if you need the new size.

luce80:
Even in a reaction code ? 
`event/offset` contains a `size` ? It should be documented hilighting the "anomaly" . 
So is it "by design" and not meant to be changed ?
Can you please show me the corrected version of the above code ?

qtxie:
@luce80 Is this correct for you?
```
view/flags/options [t: box "hello" red 200x200 ] [resize] [
	actors: object [
		on-resizing: func [face event /local siz][prin ["resizing" face/size event/face/size t/size]
			siz: event/offset - face/size    ; compute size difference
			
			t/size: t/size + (siz * 1x1)
			print [" resizing" t/size]
			t/color: random 255.255.255 ; show that the problem is only with size
		]
	]
]
```

luce80:
Ok, thanks!
Can you "correct" also this script ?
```
win: layout [t: box "hello" red 200x200 ]
react compose [t/size: win/size - (win/size - t/size)]
view/flags win 'resize
```
And my humble little modification is backward compatible and let do things in more ways.
Anyway I will not insist anymore.

qtxie:
@luce80 What's wrong with this script?

Issue #5265, opened 2023-01-05
Local words break SORT/COMPARE


dsunanda:
**Describe the bug**
A SORT/COMPARE witt /LOCALs behaves badly

**To reproduce**
```
o1: make object! [x: 1 y: 90]
o2: make object! [x: 2 y: 1]
o3: make object! [x: 2 y: 2]

data: reduce [o1 o2 o3]

sort/compare data func [a b /local ra rb][
   ra: a/x + a/y
   print [ra a/x a/y]
   rb: b/x + b/y
   print [rb b/x b/y]
   return ra < rb
   ]
 probe data
```

Different results happen if either of the PRINTs are commented out. The results are not right, with or without either of the PRINTs. Repeated running can crash the console with no message.

**Expected behavior**
The data to be sorted.
It works just fine if the local words are global, or local to a function containing the SORT....
This works just fine:

```
my-func: func [/local o1 o2 o3 data ra rb][

o1: make object! [x: 1 y: 90]
o2: make object! [x: 2 y: 1]
o3: make object! [x: 2 y: 2]

data: reduce [o1 o2 o3]

sort/compare data func [a b][
   ra: a/x + a/y
   print [ra a/x a/y]
   rb: b/x + b/y
   print [rb b/x b/y]
   return ra < rb
   ]
 probe data
 
 ]
 
 my-func
```

**Platform version**

-----------RED & PLATFORM VERSION----------- 

RED: [ branch: "master" tag: #v0.6.4 ahead: 4610 date: 7-Dec-2022/11:23:15 commit: #864cd973f53298228f242f5653843cbd61b7cbe0 ]
PLATFORM: [ name: "Windows 11" OS: 'Windows arch: 'x86-64 version: 10.0.0 build: 22621 ]
--------------------------------------------

hiiamboris:
It's clear some stack misuse is going on here. Related: https://github.com/red/red/issues/4854 https://github.com/red/red/issues/4543
I vaguely recall having similar problem with sort, one I couldn't reproduce reliably. Thanks for finding this @dsunanda :)

hiiamboris:
Example outputs (commenting out `print`, removing `return`, etc):
```
ra a/x a/y 2 1
... 1 90
*** Script Error: cannot compare make vector! [] with [value make typeset! [datatype! unset! none! logic! block! paren! string! file!
*** Where: <
*** Near : return ra < rb
*** Stack:
```
```
ra a/x a/y 2 1
... 1 90
[...]

PR #5270, opened 2023-01-19
FEAT: improve sort

qtxie:
For fixing issue #5265

### DIFF
--- a/runtime/datatypes/block.reds
+++ b/runtime/datatypes/block.reds
@@ -1162,18 +1162,21 @@ block: context [
 	compare-value: func [								;-- Compare function return integer!
 		value1   [red-value!]
 		value2   [red-value!]
-		op		 [integer!]
-		flags	 [integer!]
+		args	 [sort-args!]
 		return:  [integer!]
 		/local
 			offset	[integer!]
 			count	[integer!]
 			res		[integer!]
 			temp	[red-value!]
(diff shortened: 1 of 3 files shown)


runtime/datatypes/block.reds
```diff
@@ -1375,6 +1380,11 @@ block: context [
 					]
 					cmp: as-integer :compare-call
 					op: as-integer comparator
+					f: as red-function! comparator
+					args/locals-cnt: _function/count-locals f/spec 0 no
+					more: as series! f/more/value
+					int: as red-integer! more/offset + 4
+					args/ctx: either TYPE_OF(int) = TYPE_INTEGER [as node! int/value][global-ctx]
```
dockimbel, line 1387: Please write a unit test for this line, as I'm not 100% sure that the generated code for ARM backend would be correct there.

dockimbel:
Great!

You didn't implement any spec validation in this commit, right?

runtime/datatypes/block.reds
qtxie: OK. The same code pattern occurs in several places.
https://github.com/red/red/blob/master/runtime/interpreter.reds#L215
https://github.com/red/red/blob/master/runtime/lexer.reds#L509
https://github.com/red/red/blob/master/runtime/parse.reds#L686

PR #5293, opened 2023-03-06
FEAT: by operator

hiiamboris:
Alias to `as-pair`, as in `2 by 4`, `w by h`, `cols by rows`. 
It's readable, concise and conveys 2D meaning, and hardly has a better use.
I can compare it to addition: we can use `add` and sometimes we prefer that, but most of the time stick with just `+`. Same here. I can find couple of dozens uses of `by` in Spaces vs only a couple of `as-pair`.

### DIFF
--- a/environment/operators.red
+++ b/environment/operators.red
@@ -10,7 +10,7 @@ Red [
 	}
 ]
 
-;-- #load temporary directive is used to workaround REBOL LOAD limitations on some words
+;-- #do keep directive is used to workaround REBOL LOAD limitations on some words
 
 #do keep [to-set-word "+"]		make op! :add
 #do keep [to-set-word "-"]		make op! :subtract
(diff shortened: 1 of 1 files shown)


greggirwin:
Where is `spaces` now (I was going to look at `by` in real use cases)? I still have a ref to the old repo it seems. I thought codeberg, but that one doesn't seem to have been touched in a while.

My gut says `by` makes sense in dialects, but not so much in general use. One of those cases where the goal of natural language style sounds tempting, but may not be a win overall. And when we implement point, as you note this is for 2D only, it hits a wall. It's a good prompt to think about wording for `pair!`, which today is described for use as coordinates, but has more dimensional possibilities.

hiiamboris:
@greggirwin still there but in WIP mode https://codeberg.org/hiiamboris/red-spaces/src/branch/document

If this hits a wall with `point!`, `as-pair` does that too :)

greggirwin:
Well, we would call it `as-point`. :^)

PR #5326, opened 2023-05-30
Move `time-it` to Alias section.

greggirwin:


greggirwin:
```
Compiling to native code...
*** Compilation Error: undefined symbol: f_time-it
*** in file: %/c/dev/greggirwin/red/environment/console/gui/gui-console.red
*** in function: exec/f_clock
*** at line: 1
*** near: [
```

dockimbel:
I've spent hours on that issue, trying to bend the Red compiler to support it, but it requires too many changes in both compilers and in libRedRT generation, so it is not worth it. Once we get rid of the Red compiler, that code pattern would pose not problem at all.

PR #5327, opened 2023-06-06
FEAT: Unboring `attempt`

hiiamboris:
Tried a few variants, but paths win over apply:
```
attempt1: func [
	"Tries to evaluate a block and returns result or NONE on error"
	code [block!]
	/safer "Capture all possible errors and exceptions"
][
	; apply 'try/:all [[return do code] safer]     ;-- same
	apply/all :try [[return do code] safer]
	none
]

attempt2: func [
	"Tries to evaluate a block and returns result or NONE on error"
	code [block!]
	/safer "Capture all possible errors and exceptions"
	/local all
][
	set 'all safer										;-- `all:` refuses to compile
	try/:all [return do code]
	none
]

clock/times [attempt []] 1e6
clock/times [attempt1 []] 1e6
clock/times [attempt2 []] 1e6
```

```
0.92 μs [attempt []]
0.89 μs [attempt1 []]
0.41 μs [attempt2 []]
```
Unfortunately on failure it's still slow and presses GC a lot by spawning errors and then discarding them.

### DIFF
--- a/environment/functions.red
+++ b/environment/functions.red
@@ -26,8 +26,11 @@ attempt: func [
 	"Tries to evaluate a block and returns result or NONE on error"
 	value [block!]
 	/safer "Capture all possible errors and exceptions"
+	/local all
 ][
-	unless error? set/any 'value (apply 'try/:all [:value safer]) [get/any 'value]
+	set 'all safer										;-- `all:` refuses to compile
+	try/:all [return do code]
+	none
 ]
 
 comment: func ["Consume but don't evaluate the next value" 'value][]


dockimbel:
@hiiamboris Did you see those errors? https://ci.appveyor.com/project/red/red/builds/47236355

hiiamboris:
> We could make a native `attempt` implementation or add a new refinement to `try` to avoid the creation of the error object. Though, feasibility would need to be confirmed.

https://github.com/red/REP/issues/113 addition will significantly lower the need for fast `attempt`. So let's not chase it yet.


> @hiiamboris Did you see those errors? https://ci.appveyor.com/project/red/red/builds/47236355

I'll take a look.

hiiamboris:
@dockimbel still trying to isolate it, but it's some compiler crap. When I compile `-c` the `run-all-interp.red`, it works, when I `-r` it it fails with `bitset has no value` or smth like that (and there's no word `bitset` in the file, only `bitset!`).

dockimbel:
@hiiamboris You've made a typo. In `try/:all [return do code]`, `code` is not defined. You should rename the `value` argument to `code`.

hiiamboris:
Oh of course copy-paste problems :D

hiiamboris:
I'll PR it.

PR #5343, opened 2023-06-29
FEAT: Improved high level tracer

hiiamboris:
All kludges of the early implementation removed now. Simplified code. Works with exceptions (throw/error). Preserves the look of original data from modifications, so beginner friendly. A test suite added to avoid regressions.


greggirwin:
Nice work!

PR #5350, opened 2023-07-22
TESTS: fixes for failing base tests

hiiamboris:


hiiamboris:
For some reason crc-12 and crc-13 need more than 2 seconds to be rendered. Gave them 10 seconds for now 🤷🏻‍♀️

hiiamboris:
@qtxie should be good to go

qtxie:
Great! 👍

PR #5361, opened 2023-08-06
TESTS: Draw tests with point2D for pairs

hiiamboris:
just replaced pairs with points, to test Draw support for point2D
some tests seems outdated though even in their original form, e.g. `matrix-test.red`


dockimbel:
Great! Thank you!

PR #5364, opened 2023-08-08
FIX: edit Red/System header, File: %wcwidth.reds

isoux:
In Red/System header, File: has wrong name: %win32.reds


greggirwin:
Good catch!

hiiamboris:
Illustrates why it's a bad idea to put filename into header ;)

isoux:
> Good catch!

Finally, luckily for me, I started seriously studying Red's source code ...

PR #5378, opened 2023-09-06
Text mode UI

qtxie:
A little text mode UI backend.

Features implemented:
* Styles: `base`, `panel`, `button`, `field`, `text`, `progress`, `rich-text`, `image` and `text-list`.
* Draw command: `text`. Both `string!` and `rich-text!` are supported.
* Keyboard handling. Only `key-down` and `key` events. They are the same event as the terminals only send a single key event.
* Mouse handling. Disable by default. Use `system/view/platform/mouse-event?: yes` to enable it.
* Image!
* Timer. Use `/rate` facets. The resolution of it is 30 ms.
* Facets supported: `/offset`, `/size`, `/text`, `/image`, `/color`, `/data`, `/enabled?`, `/visible?`, `/selected`, `/flags`, `/options`, `/pane`, `/rate`, `/para` and `/draw`.
* Limited ANSI escape codes support in `/text` facet. Only [Colors / Graphics Mode](https://gist.github.com/qtxie/a5f31ccba82320ac964c311d768149d4#colors--graphics-mode) codes.
* Use 256 colors for text. It should works fine on most of the terminals. Use [Truecolor](https://en.wikipedia.org/wiki/Color_depth#True_color_.2824-bit.29) (24-bit RGB) for image rendering if the terminal supports it. Otherwise use 256 colors.
* Works on the big-3 platforms (Windows, macOS and Linux).

屏幕截图 2023-09-06 152448

Same image shows in TUI and GUI backend respectively.

image

https://github.com/red/red/assets/1673525/4cb91681-d8c9-422f-89a6-c9a2dfa60e16


dockimbel (commented):
The changes in %VID.red are not good, they affect all platforms with any View backend...

qtxie:
> The changes in %VID.red are not good, they affect all platforms with any View backend...

Should the `/tight` flag be propagated to the child containers?

dockimbel:
I'm not sure what the best default would be: to propagate or not to child panels/groups. Though, `layout` does not pass the `/tight` flag to `fetch-options`... `/tight` is just a shortcut for `origin 0x0 space 0x0`, so I think it's fine to make it propagate to children by default and let user switch to manually setting origin/space if that behavior is not wanted.

Your change in VID forces the tight mode in all cases (`layout/parent/styles/tight`), which is a regression. You should be using a dynamic refinement for that (e.g. `layout/parent/styles/:tight?`) and passing that flag somewhere to `fetch-options`.

hiiamboris:
If it propagates, let user be able to manually override this propagated value with custom origin & space.

Issue #5442, opened 2023-12-10
[View, Regression] Grid panel is broken by point2D size


hiiamboris:
**Describe the bug**

This code fails now with an error when it receives floating point width from 1st row:
```
view [
	panel 2 [
		text "text must be longer than 80px" field
		text field
	]
]
```
```
*** Script Error: float! type is not allowed here
*** Where: either
*** Near : face/offset/:axis: list/:index/offset/:axis
*** Stack: dorc view layout fetch-options layout  
```

**Expected behavior**

No error.

**Platform version**
```
-----------RED & PLATFORM VERSION----------- 
RED: [ branch: "master" tag: #v0.6.4 ahead: 5060 date: 22-Nov-2023/14:24:25 commit: #34f26dff0dfc467b9df7862f59c5220c5f508110 ]
PLATFORM: [ name: "Windows 10" OS: 'Windows arch: 'x86-64 version: 10.0.0 build: 19044 ]
--------------------------------------------
```

hiiamboris:
@qtxie I don't think you've committed it to master.

PR #5443, opened 2023-12-11
FIX: issue #5442 ([View, Regression] Grid panel is broken by point2D …

qtxie:
…size)


dockimbel:
Changes from that PR are already applied, so closing it now.

Issue #4743, opened 2020-12-06
[GTK & Mac] Transparent surfaces are eating events


hiiamboris:
**Describe the bug**

`view [backdrop red box 100x100 on-down [probe "down"]]`
click anywhere on the box
`box` should not react to clicks (at least it's transparent parts when something is drawn on it), but it does

**Platform version**
```
Red 0.6.4 for Linux built 1-Nov-2020/23:51:29+03:00 commit #2d05900
```

9214:
https://github.com/red/red/issues/1931

qtxie:
That's another issue cannot be solved without changing the low-level GUI engine.

hiiamboris:
Can't we just check the pixel color and kick the event further from one face into another, in our own code?

qtxie:
It will be very costly if no low-level API to do the checking.

hiiamboris:
Don't we have cached images of each face?

greggirwin:
> Can't we just check the pixel color 

What it the pixel color matches the one underneath?

qtxie:
> Don't we have cached images of each face?

No, we don't. There should be an OS API to get the bitmap of a face. Then we'll need to get the face under the current face, pass the event to it. If no API to get the face, we'll have to enumerate the face tree to check the bounding box and the z-order. Not an easy fix but it's possible.

qtxie:
We already have some related code to handle it on macOS. The fix is much easier than I thought. 
For GTK backend, @bitbegin had been spent quite some time on this issue, seems no easy fix.

qtxie:
> Can't we just check the pixel color and kick the event further from one face into another, in our own code?

PR #5449, opened 2023-12-20
FIX: access violation on mouse events in GTK backend

erathke:
While using base component with transparency, the GTK backend checks for transparent pixel and triggers events when necessary. This fix is related to the calculation of the position of these pixels, avoiding negative values when mouse Y is 0.

This fix is related to a previous one: [4743](https://github.com/red/red/issues/4743)
@qtxie maybe you had some specific use case in mind. Could you check this one, please?
Thanks in advance.


qtxie:
@erathke  Thank you! You're right. The minimum value of mouse Y is 0.

Issue #5559, opened 2024-11-20
Molding of unset still doesn't round-trip


hiiamboris:
**Describe the bug**

This was reported twice but still remains unsolved:
- https://github.com/red/red/issues/3409#issuecomment-557267592
- https://github.com/red/red/issues/4126#issuecomment-557804476

```
>> type? load "#(unset)"
== unset!
>> mold/all ()
== "unset"
>> type? load mold/all ()
== word!
```

**To reproduce**

`type? load mold/all ()`

**Expected behavior**

Returned type should be `unset!`

**Platform version**
`Red 0.6.5 for Windows built 9-Nov-2024/4:54:12+07:00  commit #fc80690`

greggirwin:
Copying from `none` should be easy enough. 
```
		either all? [
			string/concatenate-literal buffer "#(unset)"
			part - 8
		] [
			form value buffer arg part
		]
```
OK with you @dockimbel ?

dockimbel:
@greggirwin Yes, it's fine.

greggirwin:
Change made. Next need to update tests.

dockimbel:
Not good, can't call `form` there (see my comments in the commit page).

PR #5564, opened 2024-11-26
FIX: #5559 mold/all for unset values.

greggirwin:
https://github.com/red/red/issues/5559

### DIFF
--- a/runtime/datatypes/block.reds
+++ b/runtime/datatypes/block.reds
@@ -497,9 +497,10 @@ block: context [
 				return part
 			]
 			depth: depth + 1
-			if all [
+			
+			either all [
 				not flat?
-				value/header and flag-new-line <> 0	;-- new-line marker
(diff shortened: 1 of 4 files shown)


dockimbel:
Many tests are failing after that change. `form <unset>` should return an empty string, but `mold` should return `unset` otherwise that would cause some troubles (and break compatibility with Rebol):

```
>> mold reduce [1 () 2]
== "[1 unset 2]"
```

greggirwin:
Ah, because there's no negative offset, got it.

greggirwin:
Been way too long since I've been in R/S-land. Even if I don't use `form` I have to do something like this, right (or just return `part` directly)?
```
string/concatenate-literal buffer ""
part
```
Which still errors here.

greggirwin:
Doh! Think I have my head on straight now.

PR #5591, opened 2025-02-09
FIX: Point3D! always returns false from `zero?`

greggirwin:
Due to incomplete copy/paste change and dupe Point2D! case.

### DIFF
--- a/runtime/natives.reds
+++ b/runtime/natives.reds
@@ -1922,7 +1922,7 @@ natives: context [
 				all [p/x = 0 p/y = 0]
 			]
 			TYPE_POINT2D [
-				pt: as red-point3D! i
+				pt: as red-point2D! i
 				all [pt/x = as-float32 0 pt/y = as-float32 0]
 			]
 			TYPE_POINT3D [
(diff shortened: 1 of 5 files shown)


runtime/natives.reds
```diff
@@ -1925,7 +1925,7 @@ natives: context [
 				pt: as red-point3D! i
```
endo64, line 1925: red-point3D! should be red-point2D! in TYPE_POINT2D case.

greggirwin:
Now it look like the fix broke something. So much for quick and easy.

greggirwin:
Oh, because that first poitn3D was by design, to reuse the var, and only comparing x and y.

dockimbel:
@greggirwin You need to delete your merge commit and rebase your branch on red/master instead, as you are pushing commits that are from master already.

dockimbel:
> @dockimbel it shouldn't need the second part of this fix, unless you want to add another point var to keep them separate. Which is your preference, since you thumbed-up Endo's comment?

I've overlooked it. I prefer we just keep one variable `pt` defined as `red-point3D!` (as both structs are supposed to overlap properly).

greggirwin:
I had rebased for the first part, then the second part horked it up. :^\

PR #5605, opened 2025-03-30
FIX: missing 'planar?' function

hiiamboris:


greggirwin:
Thanks for adding that. :+1:

Issue #5661, opened 2025-10-30
`query` behavior is highly filesystem-specific


hiiamboris:
**Describe the bug**

I was trying to use `query` as a poor man's dir-or-file detector:
- `query %dir/` -> not-none
- `query %file/` -> none (if `file` is not a directory)

That works on NTFS on *clean paths*. As long as there's a reparse point in the path (junction, symlink, volume mount point) - `query %file/` succeeds as if it was a directory.

The faulty behavior is not observed on Linux, only on Windows (Mac not tested, assumed Linux-like).

**To reproduce**
1. In the shell:
   ```cmd
   $ mkdir test$ & mklink /j link$ test$ & echo. >test$\1
   ```
2. In Red:
   ```
   query %test$/1/        ;) => none
   query %link$/1/        ;) => date! value
   ```

**Expected behavior**

Returns `none` in both cases

**Platform version**
```
Red 0.6.6 for Windows built 9-Sep-2025/3:21:04+08:00  commit #c8b6ebc
```

dockimbel:
Not Red's fault:

> When you call GetFileAttributesExW(L"link$\\1\\", ...), Windows must resolve the junction before evaluating the rest of the path.
> 
> The junction link$ points to test$.
> So internally, the resolved path is the same as "test$\\1\\".
> 
> You’d expect the same failure — but:
> 
> When you access through a reparse point, the file system driver (NTFS) sometimes normalizes paths differently:
> 
> The object manager and NTFS reparse resolution strip extra trailing slashes after the reparse resolution.
> 
> As a result, "link$\\1\\" can be effectively resolved as "test$\\1" before attribute lookup.
> 
> Thus, GetFileAttributesExW() succeeds, because it ends up querying "test$\\1" — a valid file.
> 
> This behavior is not officially documented, but it has been observed in practice:
> junctions and symbolic links sometimes tolerate redundant trailing backslashes that normal file paths reject.

https://github.com/red/red/blob/master/runtime/simple-io.reds#L1095

dockimbel:
@hiiamboris If you want to make a PR for this ticket, I'll merge it, otherwise, I'll mark it "won't fix" as it's too much work for making a workaround to this odd Windows behavior.

hiiamboris:
Maybe we can just insert a check if it's physically a directory? 
I'll test if something as simple as that will work...

hiiamboris:
[...]

PR #5663, opened 2025-11-02
FIX: issue #5661 (query behavior is highly filesystem-specific)

hiiamboris:
```
>> query %/d/link$/1
== 2-Nov-2025/12:58:38.858
>> query %/d/link$/1/
== none
>> query %/d/test$/1
== 2-Nov-2025/12:58:38.858
>> query %/d/test$/1/
== none
```

### DIFF
--- a/runtime/simple-io.reds
+++ b/runtime/simple-io.reds
@@ -1080,19 +1080,29 @@ simple-io: context [
 		filename [red-file!]
 		return:  [red-value!]
 		/local
-			name [c-string!]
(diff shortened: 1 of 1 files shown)


hiiamboris:
Should I add a test to replicate all this behavior? I'm hesitant.. for this small case.. two directories..

dockimbel:
No need. I'm not sure we'll keep that API as it is now.

dockimbel (approved):

Issue #3921, opened 2019-06-20
Two scripts in subdirectory included inside context result in a compilator error.


loziniak:
I have 3 files:
```
Red [File: %test.red]
print "test"
context [
  #include %sub/a.red
  #include %sub/b.red
]
```
```
Red [File: %sub/a.red]
print "a"
```
```
Red [File: %sub/b.red]
print "b"
```
When I try to compile it with `red -r test.red`, **I get an error**:
```
*** Compilation Error: include file not found: sub/b.red 
*** in file: include_test/test.red
*** near: [
    context [
        print "a" #pop-path 
        #include %sub/b.red
    ]
]
```
When running without compilation with `red test.red`, output is as expected:
```
test
a
b
```

**Platform version**
```
Linux: Red 0.6.4 for Linux built 8-Jun-2019/10:04:27+02:00 commit #1b68168
Windows (wine): red-20jun19-33a1fb7a.exe
```

endo64:
Confirmed on Win10 x64.
Additionally if the files are in the same folder `*** Syntax Error: Missing #"]" character` error happens.

loziniak:
I found a workaround. You have to use paths relative to the file included before. Following code works:

> ```
> Red [File: %test.red]
> print "test"
> context [
>   #include %sub/a.red
>   #include %b.red ;-- NOTE CHANGE IN PATH
> ]
> ```
> 
> ```
> Red [File: %sub/a.red]
> print "a"
> ```
> 
> ```
> Red [File: %sub/b.red]
> print "b"
> ```

greggirwin:
Thanks for the further analysis @loziniak. Probably not to hard to fix with that info. %compiler.r `process-include-paths` is where the path processing probably goes awry.

loziniak:
Could be perhaps also `| #include (comp-include/only pos) :pos` line in `comp-context` function of `%compiler.r`?

loziniak:
For the record. The error message is:
```
*** Compilation Error: include file not found: sub/b.red
```

toomasv:
Somewhat connected to different interpretation of file-paths when interpreted and compiling, is another difference -- if interpreted, #include is replaced by `do` and following file is loaded and "done" but is not bound to context from within which it was included. But when compiled, source text is included in context and later bound to context.

qtxie:
@Copilot Fix it

PR #5719, opened 2026-04-03
FIX: issue #3921 (Two scripts in subdirectory included inside context…

qtxie:
… result in a compilator error.)

Fixed by AI: Qwen 3.6 Plus


qtxie:
This fix works fine on local machine and WSL, but error on Docker images on CI.

Issue #5732, opened 2026-05-18
call without /shell on POSIX still executes shell command substitution via wordexp() missing WRDE_NOCMD


ghost:
On Linux and macOS, the call function executes shell command substitution even when the /shell refinement is not used. This breaks the documented security contract of call. The root cause is in runtime/call.reds lines 561-578: the non-shell branch calls wordexp() with WRDE_SHOWERR but without WRDE_NOCMD. Per POSIX spec, omitting WRDE_NOCMD means wordexp() silently executes $(command) and backtick substitutions before handing arguments to execvp. Any user-supplied input passed to call without /shell is still interpreted by the shell. Note: the WRDE_CMDSUB handler in the switch block is dead code — without WRDE_NOCMD, wordexp() never returns WRDE_CMDSUB, it just runs the substitution and returns 0.

To reproduce
1. Open Red console on Linux or macOS
2. Run: call "echo $(id)"
3. The output will contain the result of the id command, e.g. uid=1000(user), meaning it was executed despite the absence of /shell

Expected behavior
call without /shell should pass the command string to execvp as-is, with no shell expansion, no variable substitution and no command substitution. The /shell refinement is documented as the explicit opt-in for shell interpretation.

Suggested fix
Add WRDE_NOCMD to the wordexp() call. Change: wordexp cmd wexp WRDE_SHOWERR to: wordexp cmd wexp WRDE_SHOWERR or WRDE_NOCMD. With this fix wordexp() will return WRDE_CMDSUB on substitution attempts and the existing error handler will fire correctly.

Screenshots
Not applicable, reproducible via console commands above.

Platform version
Reproducible on Linux and macOS. Affected file: runtime/call.reds, lines 561-578.

PR #5743, opened 2026-05-20
FIX: issue #5732 (call without /shell on POSIX still executes shell c…

qtxie:
…ommand)


dockimbel:
Missed that PR, I pushed the fix too. Though, making that external testing script was not necessary, a simple `call` test is enough:

https://github.com/red/red/commit/bdeb7603594973a7d2c7296b6a5d8de46342691e

PR #5745, opened 2026-05-20
Fixed width integers

qtxie:


encapper/lexer.r
```diff
@@ -898,6 +902,30 @@ lexer: context [
 		new
 	]
 
+	greater-decimal?: func [a [string!] b [string!]][
+		any [
+			greater? length? a length? b
+			all [equal? length? a length? b greater? a b]
+		]
+	]
```
dockimbel, line 910: Not a reliable way to compare integers in strings:
```
>> greater? "4" "39"
== true
>> greater? "04" "39"
== false
```

PR #5749, opened 2026-05-29
FEAT: Implemented Linux ELF import symbol versioning

qtxie:
Changed:

- system/compiler.r:1972: parses optional @VERSION refs, keeps versioned imports distinct, and rejects unsupported targets/static imports.
- system/formats/ELF.r:71: emits .gnu.version, .gnu.version_r, DT_VERSYM, DT_VERNEED, and DT_VERNEEDNUM for versioned imports.
- system/linker.r:265: handles block import keys during import cleanup.
- docs/red-system/red-system-specs.txt:3534: documents [@<version>].
- system/tests/source/compiler/regression-test-rsc.r:1598: adds compile coverage for memcpy: "memcpy" @GLIBC_2.14 [...].

### DIFF
--- a/docs/red-system/red-system-specs.txt
+++ b/docs/red-system/red-system-specs.txt
@@ -3531,13 +3531,13 @@ Red/System is able to load external shared libraries at the time a Red/System ex
 
  #import [
  	"<library>" <convention> [
- 		<function name>: "<ID>" [
+		<function name>: "<ID>" [@<version>] [
  			<argument> [<datatype>]
  			...
  			return: [<datatype>]		;-- optional part
  		]
  		...								;-- more functions mappings
- 		<variable name>: "<VAR>" [
+		<variable name>: "<VAR>" [@<version>] [
  			<datatype>
  		]
  		...								;-- more variables mappings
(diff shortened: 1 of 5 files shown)


system/tests/source/compiler/regression-test-rsc.r
```diff
@@ -1595,6 +1595,34 @@ probe [this that]
...
+			#if OS = 'Linux [
+				#import [
+					LIBC-file cdecl [
+						memcpy-default: "memcpy" [
+							target [c-string!]
+							source [c-string!]
+							size   [integer!]
+							return: [c-string!]
+						]
+						memcpy-214: "memcpy" @GLIBC_2.14 [
```
dockimbel, line 1610: Does that work if GLIBC 2.14 is not installed?

system/linker.r
```diff
@@ -265,9 +265,16 @@ linker: context [
...
 		foreach [lib list] imports/3 [
-			remove-each [name refs] list [empty? refs]
+			pos: list
+			while [not tail? pos] [
+				either empty? pos/2 [
+					remove/part pos 2
+				][
+					pos: skip pos 2
+				]
+			]
```
dockimbel: Why that change?

docs/red-system/red-system-specs.txt
```diff
@@ -3531,13 +3531,13 @@ Red/System is able to load external shared libraries at the time a Red/System ex
 
  #import [
  	"<library>" <convention> [
- 		<function name>: "<ID>" [
+		<function name>: "<ID>" [@<version>] [
```
dockimbel, line 3534: There's a problem with this syntax. `[...]` can't be used to denote an optional feature when it's already part of the overall syntax.

docs/red-system/red-system-specs.txt
```diff
@@ -3560,6 +3561,10 @@ Red/System is able to load external shared libraries at the time a Red/System ex
 
 *An absolute path to the library can be provided in OS-specific format.
 
+*The optional symbol version ref is only supported by dynamic Linux ELF
+imports. If it is omitted, the platform dynamic linker resolves the default
+symbol version exported by the shared library.
+
```
dockimbel, line 3567: The description should start by explaning clearly what that option does.

PR #5750, opened 2026-05-31
FEAT: output IDAPython sidecar that names Red/System functions in IDA

qtxie:
Added --emit-ida-script option

Usage: `rebol.exe -s red.r -r --emit-ida-script -o tmp/console.exe .\environment\console\CLI\console.red`

It will generate `console.exe` and `console.ida.py`. We can load script file `console.ida.py` in IDA after open the `console.exe`.

{FB02B10E-5414-467C-A546-4EAA4F3A7DE6}


system/rsc.r
```diff
@@ -86,6 +86,7 @@ rsc: context [
 				  "--no-runtime"		    (opts/runtime?: no)		;@@ overridable by config!
 				| ["-g" | "--debug-stabs"]  (opts/debug?: yes)		;@@ overridable by config!
 				| ["-l" | "--literal-pool"] (opts/literal-pool?: yes)
+				| "--emit-ida-script"		(opts/emit-ida-script?: yes)
```
dockimbel, line 89: Move that line above, just after "--no-runtime" option, so that long-only options stay grouped together.

dockimbel:
Please put all the Python and IDA related code in a separate file in `system/utils/`, not in the `linker.r` file.

Issue #5644, opened 2025-08-21
[CSV] Abruptly stops decoding at an arbitrary point


hiiamboris:
**Describe the bug**

I have 4088 rows in my file, but CSV only reads 244 of them. There's no error of any kind.

**To reproduce**

1. Download this file: [red-issues-list.csv](https://github.com/user-attachments/files/21912049/red-issues-list.csv)
2. Decode and inspect the data:
   ```
   >> data: load-csv/with text: read %red-issues-list.csv tab
   == [["5641" "OPEN" "LOAD-CSV returns error instead of throwing it" "status.duplicate" "2025-08-19T13:00:07Z"] ["5640" "CLOSED" "TEXT-LIST isn't updated on...
   >> length? text
   == 463019
   >> length? data
   == 244
   >> last data
   == ["5365" "CLOSED" {Transcode on file! scans different than previous Red version} {status.built, status.tested, type.bug, test.written} "2025-04-03T21:21...
   ```

**Expected behavior**

`length? data` should be 4088

**Platform version**
```
Red 0.6.6 for Windows built 13-Jul-2025/14:44:33
```

greggirwin:
Seems to be when a quoted string appears immediately after the delimiter. Based on this and some other testing.
```
>> text: read-clip
== {5363^-CLOSED^-"invalid function definition" on `return:` docstring^-status.built, status.tested, type.bug, test.written^...
>> data: load-csv/with text tab
== [["5363" "CLOSED" "invalid function definition"]]
```

greggirwin:
Per https://datatracker.ietf.org/doc/html/rfc4180, which our CSV codec implements, fields containing quotes must be enclosed in quotes, and the embedded quotes must be escaped by doubling them. Right now the behavior is different, based on where the quotes fall, but either case should throw an error IMO.

hiiamboris:
I would like codecs to be forgiving on input and strict on output. With optional warnings issued when input data is not conformant. This way we won't have to fix manually all the issues that can be fixed automatically.

greggirwin:
I often support Postel's Robustness Principle, but it's just a different set of tradeoffs. In the case of CSV the RFC is strictly informational as well. There is no actual standard. @rebolek did include a `strict?` flag that, so far, only checks field alignment. It's not accessible, but brings us back to DocKimbel's https://gist.github.com/dockimbel/8db123c382d76014b2232f09acb82e73 draft for codec interfaces.

[...]

Issue #5641, opened 2025-08-19
LOAD-CSV returns error instead of throwing it


GiuseppeChillemi:
**Describe the bug**

`LOAD-CSV` returns error instead of throwing it when not alignened data is provided

**To reproduce**

```
attempt [load-csv/with {a,b,c^M1,2,3^M4,5,6,7^M} ","]

*** User Error: Data are not aligned
*** Where: ???
*** Near : return make error! non-aligned 
```

Discussion has happened [here](https://matrix.to/#/!mjbZGzLqlsqlDLaQVP:gitter.im/$nGp1Xu2SqC3YNV4ab5AV-2Yd-uNPYn0QFnKfy8QxYVY?via=gitter.im&via=matrix.org&via=tchncs.de)

**Expected behavior**
Using `attempt` it should return `none`

**Platform version**
```
-----------RED & PLATFORM VERSION----------- 
RED: [ branch: "master" tag: #v0.6.6 ahead: 75 date: 13-Jul-2025/13:27:17 commit: #31205e5c6189648cfa852e73c71be1afc18445d0 ]
PLATFORM: [ name: "Windows 11" OS: 'Windows arch: 'x86-64 version: 10.0.0 build: 26100 ]
--------------------------------------------
```

hiiamboris:
See https://github.com/red/red/issues/5521

Issue #5521, opened 2024-07-08
CSV decoder should raise its errors


hiiamboris:
**Describe the bug**

I've tripped on this like 5 times already. When CSV decoder can't load a file, it just returns an `error!` object. Which propagates the error further, and requires a debugging effort.

**To reproduce**
```
>> type? load/as {# CSV with comments...^/a,b,c^/1,2,3} 'csv
== error!
```

**Expected behavior**

An error should be raised: behavior aligned with the rest of the language's runtime.

**Platform version**
`Red 0.6.5 for Windows built 2-Jul-2024/1:21:21+08:00  commit #bbecb7e`

greggirwin:
Image codecs just return `none` if the data isn't valid. 

Loading serialized data that has error values in it seems to lean in favor of errors as data. But if you mean that a genuine error, because of corrupt redbin data, being indistinguishable from an error value it loads successfully, let's come up with a how-to for dealing with that. If we generalize the redbin error of `Script Error: data not in correct format:` so other codecs are consistent with that, now we have a real problem to solve.

greggirwin:
I'm not disagreeing with the ticket, just adding things to consider.

hiiamboris:
> Image codecs just return `none` if the data isn't valid.

That probably deserves another ticket.

PR #5751, opened 2026-06-02
FIX: issue #5644 ([CSV] Abruptly stops decoding at an arbitrary point)

qtxie:
Changed environment/codecs/CSV.red:242:

- Tracks line-no while parsing.
- Checks parsed? after the top-level parse.
- Throws Invalid CSV data near line N instead of returning partial rows.
- Converts the strict non-aligned row path to throw with do make error!.

Added regressions in tests/source/units/csv-test.red:65:

- malformed quoted field must throw
- non-aligned rows must throw, not return an error! value

### DIFF
--- a/environment/codecs/CSV.red
+++ b/environment/codecs/CSV.red
@@ -223,7 +223,7 @@ context [
 		]
 		foreach refs disallowed [
 			if all refs [
-				return make error! rejoin [
+				do make error! rejoin [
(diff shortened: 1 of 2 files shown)


hiiamboris:
https://github.com/red/red/issues/5521 https://github.com/red/red/issues/5641

greggirwin (commented):
I prefer `line-num` to `line-no` as a convention, because the `no` part reads like logic yes/no. Otherwise this looks good.

PR #5766, opened 2026-08-31
Word remembering logic values

greggirwin:
Add round trip support for logic values and the words they came from.


hiiamboris:
Looks like a kludge :)

Start with logic words magically turning back into themselves and then one may ask how to do the same for other named constants one may define. Suddenly we'll have word ids embedded into every value, which is a road to disaster. 

Plus it introduces values that are both *same* and *strictly equal* but mold differently, which AFAICT is the first such case.

dockimbel:
Why is 90% of the PR unrelated to the feature?

I'm also not in favor of that change because:
* A logic datatype has by definition only 2 values: `true` and `false`. With your PR, it would have 6. That would confuse everyone (the word/value difference for `true/false` is already confusing enough for newcomers).
* The list of alternative names is subjective. Why not add `high/low`, `right/wrong`, `good/bad`,...?
* How do I get a `false` from an `off` or a `no`? Adding conversion is necessary but will end up just extending the confusion.
* I can forsee cases where some UI will display "true/no" or "yes/off" because the sources for those values were different. That could also happen in any serialized content like files or network payload. So it's a whole new class of bugs that would be made possible by such change.
* Users will keep asking for new "builtin" word-as-value constructs once that Pandora box is opened.

greggirwin:
Sorry I didn't catch your other updates before PRing. 

> Looks like a kludge :)

How else should we round-trip to the original words, if we want to? Mezzanines and ad hoc logic everywhere, which may easily get confused? That feels like the kludge to me, and I admit that I gave up doing that after a while, because maintaining it all manually was worse than losing my chosen words.

> Suddenly we'll have word ids embedded into every value
> Users will keep asking for new "builtin" word-as-value constructs once that Pandora box is opened.

The main issue for both of you seems to be the slippery slope.

So the worst case scenario is that people like the idea? Of course I understand the concern. What are some examples you think people might ask for? Since nobody but me has asked for this in all these years, how much of a risk do you think that really is? And if they do, that's a decision to make at that time, just like all the other things people ask for.

> It introduces values that are both same and strictly equal but mold differently

Indeed. So do blocks with differing new-line markers.

> A logic datatype has by definition only 2 values: true and false.

It still does. They just have a small number domain specific words.

> the word/value difference for true/false is already confusing enough for newcomers

I hope this addresses that in some cases, because the current system is lossy and inconsistent. Why do we support yes/no and on/off now, if we only ever throw them away?

> The list of alternative names is subjective.

They match what we already support. Your example names aren't good examples, as you intend. I don't feel like we should add `checked/unchecked` words on the first pass, but there's an argument for that. Think of checklists outside VID as well. Are those useful domain specific words?

`view [chk-yes?: check "Should I?" on [print mold face/data]]` 

> How do I get a false from an off or a no?

Why do you need to, if you used off or no initially? I didn't design something specifically for that (`make` and `to` both retain them but don't have to), so conversion is currently necessary: `to-true-false: func [val [logic!]] [val and val]` works right now.

> I can forsee cases where some UI will display "true/no" or "yes/off" because the sources for those values were different. That could also happen in any serialized content like files or network payload. So it's a whole new class of bugs that would be made possible by such change.

I didn't think to make `not` word-aware. Will fix that now. :^)

As far as bugs, can you give examples, since these are all still logic values? If it's what a UI displays, then that could happen. But we have that issue today as well, right? You can't tell if `true`  is a word or logic value without checking its type. For serialized data, that's the point of construction syntax support. We also have this case already:
```
>> mold/all construct [a: true b: #(off)]
== {make object! [^/    a: #(true)^/    b: #(off)^/]}
>> mold/all construct/only [a: true b: #(off)]
== {make object! [^/    a: 'true^/    b: #(off)^/]}
```
Why does that feature exist?

Coming back to a primary use case, I can't use anything but #(true)/#(false) in maps, without evaluation of some kind, right? So even if on/off/yes/no are a better fit, I have to do this.
```
>>  m: make map! compose [a: (yes)]
== #[
    a: true
]
```
My chosen word is already gone. And, look, you can't tell if that's `true` or `#(true)` in the body. Seems like a word, since maps don't evaluate, right?

greggirwin:
Ah, `not` doesn't let the logic value do the complement itself, but it could. In the meantime, `complement` works after commenting out one line that reset the header bits.
```
>> complement #(on)
== off
>> complement #(yes)
== no
>> complement #(true)
== false
```

hiiamboris:
>> It introduces values that are both same and strictly equal but mold differently

> Indeed. So do blocks with differing new-line markers.

Not true :) Blocks that are *same* mold the same.

Anyway we both feel that the negative long term repercussions for the language far outweigh the benefit. This is the same pushback Giuseppe always gets from trying to make his personal use case part of the language foundation. So let's start with a REP with a clear use case description? Let us understand your pain :) The lack of `mold` precision (which Spaces has to fix locally btw) is not a valid argument in support of this PR.

> `val and val`

Deeply esoteric feature.

> Why do we support yes/no and on/off now, if we only ever throw them away?

I'd say it's a very cheap addition for more code clarity and dialects (all - just input). `true`/`false` being canonical and others are just handy aliases. 

I will mention that HF got so pissed off with logic values that he eradicated them completely, leaving only `~null~` as falsey.

greggirwin:
> Not true :) Blocks that are same mold the same.

Not if they have differing new-line markers, which is for humans, right? Same argument for this feature.

> Anyway we both feel that the negative long term repercussions for the language far outweigh the benefit.

Which, of course, we don't know. It could be that this reduces, or helps track bugs, because you see an `#(on)` somewhere you didn't expect, and you ask "Where did that come from?"

> So let's start with a REP with a clear use case description? Let us understand your pain :)

I thought I did that, aside from the REP part. The pain is simple. second-class logic words get lost, but don't have to. The words have value to humans and, quite likely, to LLMs.

> I'd say it's a very cheap addition for more code clarity and dialects (all - just input). true/false being canonical and others are just handy aliases.

So they are cheap, clear, and useful for dialects (because they have meaning, right?), but only once?

Can I say with certainty this will be an improvement and make things better? Of course not. Not anymore than you can say it will make things worse.

`yes` appears 1'500 times in %.reds files, most of which are `#if debug? = yes ...` with many also being test results like `failed: yes  ...` . Only ~115 times in %.red files, many also in tests.

As to the risk, the only people ever at risk of suffering a bug due to this enhancement are those who use it. If true/false is all you ever need and use, canonically, you may never even know this feature exists.

hiiamboris:
> Not if they have differing new-line markers, which is for humans, right? Same argument for this feature.

Maybe you forgot what sameness means :)

greggirwin:
Ah, mental glitch on my end carried from the initial strictly equal comment. So that's design question. I don't recall ever using `same?` or `strict-equal?` on logic values, which then leads to what use cases that has and how this would affect them.

hiiamboris:
> > So let's start with a REP with a clear use case description? Let us understand your pain :)
> 
> I thought I did that, aside from the REP part. The pain is simple. second-class logic words get lost, but don't have to. The words have value to humans and, quite likely, to LLMs.

No, obviously that's too vague. 
```
>> a: on
== true
>> mold a
== "true"
>> pick ["on" "off"] a
== "on"
```
Ok, so we got `"true"` but the workaround is very easy. Where's the pain? You didn't show any.

What is even the problem with those words being lost? Why do you even have to reduce them (and lose the original spelling)? How is this different from e.g. defining `one: 1 two: 2 three: 3` and demanding `1` molds back to `one` and so on?

greggirwin:
I thought I wrote up examples in the PR, but maybe I'm losing my mind.

A simple map! example. In real life, the data is in a file maintained by a human and/or agent.
```
; This doesn't work
config: #[human-user?: yes debug-mode?: off]
if config/debug-mode? [print "Debugging"]

; So you do this, which means the user has to write the paren.
; You have to know to compose it, and how do you know it's 
; safe to do so?

config: make map! compose [human-user?: (yes) debug-mode?: (off)]
if config/debug-mode? [print "Debugging"]

; Rats, I lost my words. Do I now write a mezz system to check
; the list of known words and go back to plain words, forgoing
; the value of logic! ?

mold config
mold/all config

; Saving now means either my words are lost, or I have to use more
; mezz stuff to rebuild them, and those wrappers have to know what
; fields should be touched. So the simple workaround has to happen
; everywhere.

; So you do this instead, which isn't great for humans because
; of construction syntax, but at least it's safe. Sadly, now you
; can't use the most meaningful words. On the bright side, agents
; no longer say "I will see why the config data didn't round-trip
; as I expected. Please stand by..."

config: #[human-user?: #(true) debug-mode?: #(false)]

;!! You haven't built the new version, so these won't work for you yet.

; In the new approach, we're back to construction mode. 
config: #[human-user?: #(yes) debug-mode?: #(off)]
; But you can still compose if you want
config: make map! compose [human-user?: (yes) debug-mode?: (off)]
; Or make a mezz of things
config: #[human-user?: yes debug-mode?: off]
foreach [k v] config [
    if find [true false on off yes no] v [
        config/:k: get v
    ]
]
>> mold config ; for all cases
== #[
    human-user?: yes
    debug-mode?: off
]
>> mold/all config ; for all cases
#[
    human-user?: #(yes)
    debug-mode?: #(off)
]

; This works, and the words round-trip.
if config/debug-mode? [print "Debugging"]

; And if you have tools to help manage the files, they can deal with
; the construction syntax, without any special mezz system in place.

; End of config example.

; I haven't changed `not`, but did update `complement` locally to
; keep the header bits. Meant commenting out one line.
>> config/human-user?: complement config/human-user?
== no
>> config/debug-mode?: complement config/debug-mode?
== on
>> print mold/all config
#[
    human-user?: #(no)
    debug-mode?: #(on)
]

; The above points out that `no` and `on` are very close, so more risk
; of human typos.

>> system/options/float/full?: off
== off
>> system/options/float/pretty?: on
== on
>> print mold system/options
make object! [
    ...
    float: make object! [
        pretty?: on
        full?: off
    ]
]
```
When you talk to a human (or agent), do you say "Do you have float/pretty? true?" ;^)

hiiamboris:
It is as I anticipated then. 

The problem here is that your use case is a chain of misunderstandings and bad practices:

1. Config files must *never* be written by the program. In fact you wouldn't even know *where* to write because normally configuration is assembled from multiple levels: org-wide, system-wide and user-scoped - all different files, and you can only really write to the user-scope one but it's going to override the upper levels once they change under you. So no one does that in production. You should only ever read the config files, and the content gets preserved automatically.
2. Asking the user to write or alter config files is still a bad practice (yet unfortunately very common). Why should user bother learning your config format? Or any config format? Show a GUI instead.
3. Asking agent to modify config is fair, but to the agent true/false shouldn't be any worse than on/off.
4. You shouldn't expect every Red user to reinvent the wheel of reading the config. I have a [standard-compliant cross-platform config/state script](https://codeberg.org/hiiamboris/red-common/src/branch/master/data-store.red), which would benefit a lot from others testing it. At some point such functionality must be in the language (as a module). I want to simplify it, remove config saving, remove evaluation, make the usage more smooth, as it still feels bulkier than it should be to me.
5. What you're saving is program's *state*. Which is not for the user's eyes at all: it's what you exposed in the UI and want to preserve between restarts. Redbin is the best fit.
6. Don't put your config where your binary is, unless it's a portable setup by design. Normally, binary should be on a read-only FS, user config - writeable and synced between devices, state - volatile, discardable and purely local.
7. Ideally do not evaluate/compose state. Risk of code injection from some unvetted data fields.

dockimbel:
> How else should we round-trip to the original words, if we want to?

Use... words. `light?: 'on` round-trips.


> The main issue for both of you seems to be the slippery slope.

For me, it's the semantic issues I listed above coming from expanding the line blurring between words and logic values.


>> A logic datatype has by definition only 2 values: true and false.
> It still does. They just have a small number domain specific words.

No, it would factually have 6 values, all sharing the exact same properties, nothing distinguishing them other than subjective categorization.


> Why do we support yes/no and on/off now, if we only ever throw them away?

We don't have any specific support for yes/no and on/off other than defining words and giving them values. The same way as for `pi` and many other predefined words. A word is not its referred value. Unfortunately, logic values inherit the Rebol design and end up with the same literal form on molding. Though the radius of this possible source confusion is currently limited, your proposition would increase it.


>> The list of alternative names is subjective.
> They match what we already support. Your example names aren't good examples, as you intend.

Low/High are used interchangeably with 0/1 in electronics, so that example looks to me like a more relevant candidate than e.g. on/off.


>> How do I get a false from an off or a no?
> Why do you need to, if you used off or no initially?

For example, your library provides me with on/off and I want true/false. As we are in the subjectivity territory since the beginning, it's my preference. ;-)


> Why does that feature exist?

It does not in master Red branch. I think you tested on your own modified branch.

```
>>  mold/all construct [a: true b: #(off)]
*** Syntax Error: (line 1) invalid construction spec at off)]
*** Where: case
*** Near : transcode/trace buffer :delimiter-lex
*** Stack: error? load

>> mold/all construct/only [a: true b: #(off)]
*** Syntax Error: (line 1) invalid construction spec at off)]
*** Where: case
*** Near : transcode/trace buffer :delimiter-lex
*** Stack: error? load
```

> And, look, you can't tell if that's true or #(true) in the body. Seems like a word, since maps don't evaluate, right?

You can, a word value would be molded as a lit-word:
```
>> m: make map! compose [a: (yes)]
== #[
    a: true
]
>> m: make map! compose [a: 'true]
== #[
    a: 'true
]
```

> second-class logic words get lost

There no such thing, as there are no second-class float words (example: `pi`).



Your whole proposition is based on exploiting the current confusion between logic values literal value and the words commonly used to refer to those values, and making that confusion broader. I don't think that's an improvement. I believe Carl let that slip to avoid any esoteric representation for values that are so commonly used in programming, but it is an exception that needs to be contained, not expanded.

greggirwin:
> Why does that feature exist?

My fault in the example. Should have been the current way words are used, and `construct/only` is the feature in question.

Nenad, the use of words means having logic to check those exact words everywhere for "truthiness" or a mezz solution that is used at all load/save points, which I think is the much more fragile solution.

> For example, your library provides me with on/off and I want true/false.

Then *you* (and Boris) only ever use true/false. You would never even know there was another option.

To Boris' point, this isn't only about config files. They are a handy example. It's about persistent Red data in the grand scheme of things as a messaging language.

dockimbel:
> Nenad, the use of words means having logic to check those exact words everywhere for "truthiness" or a mezz solution that is used at all load/save points, which I think is the much more fragile solution.

It can be as simple as this:
```
>> light?: 'on
== on
>> get light?
== true
>> light?: 'off
== off
>> get light?
== false
```

greggirwin:
It is as simple as that, in an isolated example. Now do that everywhere, and ask everyone else to do it too. Forget the `get` and `light?` is always truthy (`either light? ...`), which means you have to *know* every instance that requires it. Assign a different word by mistake and you get...whatever that word refers to, or an error.

greggirwin:
And in case you think I haven't thought about wrappers for this... https://github.com/greggirwin/red-formatting/blob/master/format.red#L578  Somewhere I have their `load`ing counterparts.

hiiamboris:
Simply pre-process when you load/save, or don't use `mold` format, which was meant for output readability, not round-tripping.

Formatting example you mention is definitely not for round-tripping, but even further into readability territory.

greggirwin:
It's funny how both of you seem to think I'm new to this.

As to the wrappers, they weren't originally built for round-tripping, but for interop, then included as an idea in the `format` experiment.

`mold` is precisely for round-tripping. BTW. Except when it isn't, or is at least lossy, like today:
```
>> mold on
== "true"
```
But that's what you want, instead of this:
```
>> mold on
== "on"
```
Or maybe I've been using `mold` wrong all these years.

But we should probably stop beating this poor, dead horse.

hiiamboris:
No, I believe mold was meant for us - developers - to visually inspect the data. Comes back to https://github.com/red/REP/issues/134

